PR #12, opened 2025-01-23
Add `zcashd`-compatible JSON-RPC endpoint

str4d:
`jsonrpsee` logic adapted from `zebra-rpc` and `zebrad` crates.

### DIFF
--- a/zallet/src/commands/start.rs
+++ b/zallet/src/commands/start.rs
@@ -1,18 +1,60 @@
 //! `start` subcommand - example of how to write a subcommand
 
-use abscissa_core::{config, FrameworkError, Runnable, Shutdown};
+use abscissa_core::{config, tracing::Instrument, FrameworkError, Runnable, Shutdown};
+use tokio::{pin, select};
 
-use crate::{
-    cli::StartCmd,
-    config::ZalletConfig,
-    error::{Error, ErrorKind},
-    prelude::*,
-};
+use crate::{cli::StartCmd, components::json_rpc, config::ZalletConfig, error::Error, prelude::*};
 
 impl StartCmd {
     async fn start(&self) -> Result<(), Error> {
-        println!("run");
-        Err(ErrorKind::Discombobulated.into())
+        let config = APP.config();
+
+        // Launch RPC server.
+        let rpc_task_handle = if let Some(listen_addr) = config.rpc.listen_addr {
+            info!("Spawning RPC server");
+            info!("Trying to open RPC endpoint at {}...", listen_addr);
+            json_rpc::server::spawn(config.rpc.clone()).await?
+        } else {
+            warn!("Configure a listen_addr to start the RPC server");
+            // Emulate a normally-operating ongoing task to simplify subsequent logic.
+            tokio::spawn(std::future::pending().in_current_span())
+        };
+
+        info!("Spawned Zallet tasks");
+
+        // ongoing tasks.
+        pin!(rpc_task_handle);
+
+        // Wait for tasks to finish.
+        let res = loop {
+            let mut exit_when_task_finishes = true;
+
+            let result = select! {
+                rpc_join_result = &mut rpc_task_handle => {
+                    let rpc_server_result = rpc_join_result
+                        .expect("unexpected panic in the RPC task");
+                    info!(?rpc_server_result, "RPC task exited");
+                    Ok(())
+                }
+            };
+
+            // Stop Zallet if a task finished and returned an error, or if an ongoing task
+            // exited.
+            match result {
+                Err(_) => break result,
+                Ok(()) if exit_when_task_finishes => break result,
+                Ok(()) => (),
+            }
+        };
+
+        info!("Exiting Zallet because an ongoing task exited; asking other tasks to stop");
+
+        // ongoing tasks
+        rpc_task_handle.abort();
+
+        info!("All tasks have been asked to stop, waiting for remaining tasks to finish");
+
+        res
     }
 }
 
--- a/zallet/src/components/json_rpc.rs
+++ b/zallet/src/components/json_rpc.rs
@@ -1,0 +1,10 @@
+//! JSON-RPC endpoint.
+//!
+//! This provides JSON-RPC methods that are (mostly) compatible with the `zcashd` wallet
+//! RPCs:
+//! - Some methods are exactly compatible.
+//! - Some methods have the same name but slightly different semantics.
+//! - Some methods from the `zcashd` wallet are unsupported.
+
+pub(crate) mod methods;
+pub(crate) mod server;
--- a/zallet/src/components/json_rpc/methods/get_wallet_info.rs
+++ b/zallet/src/components/json_rpc/methods/get_wallet_info.rs
@@ -1,0 +1,65 @@
+use jsonrpsee::{core::RpcResult, tracing::warn};
+use serde::{Deserialize, Serialize};
+
+/// Response to a `getwalletinfo` RPC request.
+pub(crate) type Response = RpcResult<GetWalletInfo>;
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) struct GetWalletInfo {
+    /// The wallet version, in its "Bitcoin client version" form.
+    walletversion: u64,
+
+    /// The total confirmed transparent balance of the wallet in ZEC.
+    balance: f64,
+
+    /// The total unconfirmed transparent balance of the wallet in ZEC.
+    ///
+    /// Not included if `asOfHeight` is specified.
+    unconfirmed_balance: Option<f64>,
+
+    /// The total immature transparent balance of the wallet in ZEC.
+    immature_balance: f64,
+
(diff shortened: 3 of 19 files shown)


zallet/src/components/json_rpc.rs
```diff
@@ -0,0 +1,7 @@
+//! JSON-RPC endpoint.
+//!
+//! This provides JSON-RPC methods that are (mostly) compatible with the `zcashd` wallet
```
daira, line 3:
```suggestion
//! This provides JSON-RPC methods that are mostly compatible with the `zcashd` wallet
```
We are not trying for full compatibility, I think.

zallet/src/components/json_rpc/methods/list_unified_receivers.rs
```diff
@@ -0,0 +1,34 @@
...
+pub(crate) struct ListUnifiedReceivers {
+    /// The legacy P2PKH transparent address.
+    ///
+    /// Omitted if `p2sh` is present.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    p2pkh: Option<String>,
+
+    /// The legacy P2SH transparent address.
+    ///
+    /// Omitteed if `p2pkh` is present.
```
daira, line 17:
```suggestion
    /// Omitted if `p2pkh` is present.
```

zallet/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    /// The timestamp in seconds since epoch (midnight Jan 1 1970 GMT) that the wallet is
+    /// unlocked for transfers, or 0 if the wallet is locked.
+    unlocked_until: u32,
+
+    /// The BLAKE2b-256 hash of the HD seed derived from the wallet's emergency recovery phrase.
+    mnemonic_seedfp: String,
+}
+
+pub(crate) fn call() -> Response {
```
nuttycom, line 49: Couple of questions about this:

In the future, is the plan for this to take a `WalletRead` argument? 

What happens if this needs to be an async function, is that supported?

zallet/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -0,0 +1,292 @@
...
+impl<S> HttpRequestMiddleware<S> {
+    /// Create a new `HttpRequestMiddleware` with the given service.
+    pub fn new(service: S) -> Self {
+        Self { service }
+    }
+
+    /// Insert or replace client supplied `content-type` HTTP header to `application/json` in the following cases:
+    ///
+    /// - no `content-type` supplied.
+    /// - supplied `content-type` start with `text/plain`, for example:
```
daira, line 54:
```suggestion
    /// - supplied `content-type` starts with `text/plain`, for example:
```

zallet/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -0,0 +1,292 @@
...
+    ///
+    /// `application/json` is the only `content-type` accepted by the Zallet rpc endpoint:
+    ///
+    /// <https://github.com/paritytech/jsonrpc/blob/38af3c9439aa75481805edf6c05c6622a5ab1e70/http/src/handler.rs#L582-L584>
+    ///
+    /// # Security
+    ///
+    /// - `content-type` headers exist so that applications know they are speaking the correct protocol with the correct format.
+    ///   We can be a bit flexible, but there are some types (such as binary) we shouldn't allow.
+    ///   In particular, the "application/x-www-form-urlencoded" header should be rejected, so browser forms can't be used to attack
```
daira, line 67: I don't see where this is checked.

zallet/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -0,0 +1,292 @@
...
+    ///   a local RPC port. See "The Role of Routers in the CSRF Attack" in
+    ///   <https://www.invicti.com/blog/web-security/importance-content-type-header-http-requests/>
+    /// - Checking all the headers is secure, but only because hyper has custom code that just reads the first content-type header.
```
daira, line 70: I don't understand this comment.

zallet/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -0,0 +1,292 @@
...
+        let (parts, body) = response.into_parts();
+        let bytes = body
+            .collect()
+            .await
+            .expect("Failed to collect body data")
+            .to_bytes();
+
+        let bytes =
+            if let Ok(request) = serde_json::from_slice::<'_, JsonRpcResponse>(bytes.as_ref()) {
+                serde_json::to_vec(&request.to_version(version))
```
nuttycom, line 138: I'm puzzled by this variable being named `request`. If this is intended, document what's happening here; if it's a copy-paste error, then please rename it to something less confusing.

zallet/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -0,0 +1,292 @@
...
+    error: Option<serde_json::Value>,
+    id: serde_json::Value,
+}
+
+impl JsonRpcResponse {
+    fn to_version(mut self, version: JsonRpcVersion) -> Self {
+        match version {
+            JsonRpcVersion::Bitcoind => {
+                self.jsonrpc = None;
+                if self.result.is_none() {
```
daira, line 268:
```rust
self.result = self.result.or(serde_json::Value::Null).
```

nuttycom (approved):
tested ACK with questions/requests for additional documentation

daira (approved):
utACK

zallet/src/components/json_rpc.rs
str4d: What I was meaning here is that some JSON-RPC methods with the same name will be exactly compatible with their equivalent `zcashd` methods, while others may have deviations.

zallet/src/components/json_rpc/server/http_request_compatibility.rs
str4d: This comment was written by the Zebra devs, and I think it might be out of date after their recent migration to the `jsonrpsee` crate. As noted in an earlier comment, the `jsonrpc` crate they previously used only accepts `application/json`, and it's probably the same for `jsonrpsee`.

str4d: Indeed, `jsonrpsee` does the same thing: https://github.com/paritytech/jsonrpsee/blob/656f8bb0793c8e992d20b47c3d17e7a6c396fb8b/server/src/transport/http.rs#L14-L29

zallet/src/components/json_rpc/server/http_request_compatibility.rs
str4d: It comes from this review by teor: https://github.com/ZcashFoundation/zebra/pull/6885#pullrequestreview-1473817729

zallet/src/commands/start.rs
```diff
@@ -1,18 +1,59 @@
...
+        let config = APP.config();
+
+        // Launch RPC server.
+        let rpc_task_handle = if let Some(listen_addr) = config.rpc.listen_addr {
+            info!("Spawning RPC server");
+            info!("Trying to open RPC endpoint at {}...", listen_addr);
+            json_rpc::server::spawn(config.rpc.clone()).await?
+        } else {
+            warn!("Configure a listen_addr to start the RPC server");
+            tokio::spawn(std::future::pending().in_current_span())
```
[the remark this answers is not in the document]
str4d: This is the pattern used in `zebrad`. It means we don't need conditional handling of every optional task further down in the code because we're using a dummy value when the optional task is disabled. Recall that this does not spawn a thread, but an async task (one that in this case always reports "pending", never completes, and can only be dropped or aborted, which is the behaviour we want for normally-operating ongoing tasks).

zallet/src/components/json_rpc/server/http_request_compatibility.rs
str4d: copy-pasta 🍝

zallet/src/components/json_rpc/methods/get_wallet_info.rs
str4d: - Yes, these will take some `Arc<RwLock<WalletDb>>`-style argument once that is introduced in a subsequent PR.
- Yes, `async` functions are supported. Example in `zebra-rpc`: https://github.com/ZcashFoundation/zebra/blob/79fbc03ece01c1efab1a243219b4cc49b7fb149f/zebra-rpc/src/methods.rs#L97

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/6bd0dbe6eb2b1d4a83e43e928700874f825b7b00..481460a387b94f67d5b49cde2c43e2cec401f398) to address review comments.

nuttycom (approved):
re-utACK 481460a387b94f67d5b49cde2c43e2cec401f398

PR #13, opened 2025-01-27
Add `zallet migrate-zcash-conf` command

str4d:


zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,558 @@
...
+
+        if !self.this_is_alpha_code_and_you_will_need_to_redo_the_migration_later {
+            return Err(ErrorKind::Generic.context(fl!("migrate-alpha-code")).into());
+        }
+
+        // Serialize the config.
+        let config_toml =
+            toml::Value::try_from(config).map_err(|e| ErrorKind::Generic.context(e))?;
+        let output =
+            toml::to_string_pretty(&config_toml).map_err(|e| ErrorKind::Generic.context(e))?;
```
str4d, line 108: Due to how `serde + toml` serialization works, any config option that is `Option<T>` is omitted when `None`, but any config option that is just `T` is not omitted when set to its default value. This is how `zebrad` works as well, but I'm not very happy with it. I'm half inclined to make _every_ config option `Option<T>` so that it can correctly represent "user-set / not-user-set" in memory, and then have getters that map `None` to the intended default value.

nuttycom: I guess that seems reasonable. In some respects though I kind of like the idea that the defaults are explicitly set in the config file, because then the user doesn't have to go hunting in various places to find out what options are available and what their defaults are; they just get to see all the options up front when looking at the config file.

I know that I value when default config files are "complete" and self-documenting, such that I don't have to go looking for separate documentation.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,558 @@
...
+                None => Err(ErrorKind::Generic.context(fl!(
+                    "err-migrate-invalid-line",
+                    line = line,
+                    conf = conf.display().to_string(),
+                ))),
+            }?;
+            let value = rest
+                .split_once('#')
+                .map(|(value, _)| value.trim_end())
+                .unwrap_or(rest.trim_end());
```
nuttycom, line 66: nit: I'd use `map_or` here.

nuttycom (changes requested):
Generally looks good, though I think that some of the options that you've set up migrations for should not be migrated.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,558 @@
...
+        .chain(Action::map_bool("walletbroadcast", |config| {
+            &mut config.broadcast
+        }))
+        // TODO: Decide if we want to change how this is configured, or do anything to
+        // improve security.
+        .chain(Action::map(
+            "walletnotify",
+            |config| &mut config.notify,
+            |value| Ok(Some(value.into())),
+        ))
```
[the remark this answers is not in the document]
str4d: This is the feature that interacts with `walletbroadcast`; it's how I set up my "send transactions over private remailer" demo back in the day.

zallet/src/config.rs
```diff
@@ -8,18 +8,101 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct ZalletConfig {
+    /// Make the wallet broadcast transactions.
+    pub broadcast: bool,
```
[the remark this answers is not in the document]
str4d: This flag when set to `false` in `zcashd` means that created transactions do not get sent to the mempool automatically, instead allowing the user to control when and how the transaction is broadcast. This can be used for e.g. more private relaying than what `zcashd` / `zebrad` does, or org-level final transaction checks.

zallet/src/config.rs
```diff
@@ -8,18 +8,101 @@ use serde::{Deserialize, Serialize};
...
+
+    /// Run in the background as a daemon and accept commands.
+    #[cfg(unix)]
+    pub daemon: bool,
+
+    /// Directory to be used when exporting data.
+    pub export_dir: Option<String>,
+
+    /// Run a thread to flush wallet periodically.
+    pub flush: bool,
```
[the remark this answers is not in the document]
str4d: This task would periodically call `rusqlite::Connection::cache_flush` (and do anything else necessary to ensure changes in memory, for example anything stored outside the SQLite backend, are written to disk). The config flag in `zcashd` was for disabling it (as it defaulted to on).

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,558 @@
...
+    }
+
+    #[cfg(not(any(windows, target_os = "macos")))]
+    {
+        home::home_dir().map(|base| base.join(".zcash"))
+    }
+}
+
+/// The action to take when a specific `zcashd` option is encountered.
+enum Action {
```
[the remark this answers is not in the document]
str4d: I already have a TODO for adding support for options allowed to occur multiple times. Once that's done, I agree that options which aren't allowed to appear multiple times should cause an error if they do appear multiple times.

zallet/src/config.rs
```diff
@@ -8,18 +8,101 @@ use serde::{Deserialize, Serialize};
...
+    pub notify: Option<String>,
+
+    /// By default, the wallet will not allow generation of new spending keys & addresses
+    /// from the mnemonic seed until the backup of that seed has been confirmed with the
+    /// `zcashd-wallet-tool` utility. A user may start zallet with `--walletrequirebackup=false`
+    /// to allow generation of spending keys even if the backup has not yet been confirmed.
+    pub require_backup: bool,
+
+    /// Rescan the block chain for missing wallet transactions on startup.
+    pub rescan: bool,
```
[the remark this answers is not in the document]
str4d: I agree that whichever config options we decide should not ever be in config files, should in this migration be implemented as warnings to the user.

zallet/src/config.rs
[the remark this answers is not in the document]
nuttycom: Hmmm... my inclination is that we should narrow the scope of Zallet  as much as possible and not support this for the initial version.

zallet/src/config.rs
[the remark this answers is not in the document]
nuttycom: Since the semantics of this flag will almost certainly be different from they are in zcashd, I don't think we should migrate it.

zallet/src/commands/migrate_zcash_conf.rs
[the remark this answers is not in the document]
nuttycom: Even so, I don't think it needs to be in the initial version.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/fa126fd4384ca2589a968d8acb6648f1e65cf937..e494b89c0660e035ab2fc62db20b167c0ec14b13) to address review comments and implement some TODOs.

zallet/src/commands/migrate_zcash_conf.rs
str4d: I think what I'd prefer is to have a separate command (similar to `zebrad generate`) that generates a config file containing _all_ of the defaults (unlike `zebrad` which doesn't fill in defaults for some of its values like `Option<T>`s). Then users can run it to get a readable form of whatever the current defaults are, and adapt / use them at will. (The downside of offering this command is that most users might end up configuring the defaults explicitly, preventing us from ever updating them. So maybe the command should actually generate an empty config file with every default commented out.)

For this command, I prefer that it only generate a file containing settings that were explicitly configured by the user in their `zcash.conf`.

nuttycom: That makes sense. Having the defaults commented out is a pretty standard way of approaching the problem, and in the associated comments we can also describe *why* those particular defaults were chosen, and what the valid ranges etc. are.

nuttycom (approved):
utACK e494b89c0660e035ab2fc62db20b167c0ec14b13

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+                Ok(0..=3) => Err(()),
+                Ok(n @ 4..) => Ok(n),
+                Err(_) => Err(()),
+            },
+        ))
+        // TODO: Decide if we want to distinguish between database migrations (which we
+        // currently require) and wallet format upgrades.
+        .chain(Action::ignore("upgradewallet"))
+        // TODO: Decide whether we want to allow renaming the `WalletDb` backing file.
+        .chain(Action::ignore("wallet"))
```
daira, line 349: This should probably be a warning if it is not the default.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+
+    // Node options used directly by wallet code.
+    let node_options_direct_wallet = iter::empty()
+        // Experimental feature we aren't migrating.
+        .chain(Action::ignore("developerencryptwallet"))
+        // Used to re-enable CPU mining after disabling it during proving.
+        // Irrelevant for Zallet which doesn't include mining.
+        .chain(Action::ignore("genproclimit"))
+        // TODO: Figure out where this was used, and if we want to keep it.
+        .chain(Action::ignore("maxtxfee"))
```
daira, line 417: It's used to prevent accidentally specifying an excessively large fee.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+        .chain(Action::ignore("paymentdisclosure"))
+        .chain(Some((
+            "preferredtxversion",
+            Action::warn(|value| {
+                Some(fl!(
+                    "migrate-warn-unsupported",
+                    option = "preferredtxversion",
+                    value = value,
+                ))
+            }),
```
daira, line 431: There is no need to warn if the preferred version is 5, since that's the default for zallet (and will continue to be the default immediately after NU7 I think?)

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+            "daemon",
+            Action::warn(|value| {
+                (value != "0").then(|| fl!("migrate-warn-daemon", option = "daemon"))
+            }),
+        )))
+        // We don't want to store Zallet data in the same folder as `zcashd` data.
+        .chain(Action::ignore("datadir"))
+        // The logging systems of `zcashd` and Zallet differ sufficiently that we don't
+        // try to map existing log targets across.
+        .chain(Action::ignore("debug"))
```
daira, line 449: This could be a warning with a link to documentation for Zallet's logging.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+        // We aren't going to migrate over the `zcashd` wallet's experimental features
+        // compatibly. If we add a similar framework to Zallet it will be for from-scratch
+        // features.
+        .chain(Action::ignore("experimentalfeatures"))
```
daira, line 453: This could be a warning saying that all experimental feature flags are ignored.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+        .chain(Action::map(
+            "exportdir",
+            |config| &mut config.export_dir,
+            |value| Ok(value.into()),
+        ))
+        // TODO
+        .chain(Action::ignore("regtest"))
+        // TODO: Support mapping multi-arg options.
```
daira, line 461: I thought there is support for that in this PR?

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+        .chain(Action::ignore("rpcallowip"))
+        // Unsupported in `zcashd` since 1.0.0-beta1.
+        .chain(Action::ignore("rpcasyncthreads"))
```
daira, line 464: Oh, *that's* part of the cause of the excessive complexity in the zcashd RPC code. (It was a good decision to disable this, mind.)

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+        ))
+        // Unsupported in `zcashd` since 1.0.8.
+        .chain(Action::ignore("rpcssl"))
+        // `jsonrpsee` is an async JSON-RPC implementation, so it doesn't have dedicated
+        // threads; instead it shares the Tokio worker thread pool.
+        .chain(Action::ignore("rpcthreads"))
+        // TODO: Not quite the same thing as `ServerBuilder::set_message_buffer_capacity` I think?
+        .chain(Action::ignore("rpcworkqueue"))
+        // TODO
+        .chain(Action::ignore("testnet"));
```
daira, line 501: This should be a warning for now I think.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -0,0 +1,644 @@
...
+        "equihashsolver",
+        "externalip",
+        "forcednsseed",
+        "fundingstream",
+        "fuzzmessagestest",
+        "gen",
+        "help",
+        "help-debug",
+        "ibdskiptxverification",
+        "insightexplorer",
```
daira, line 551: This probably deserves a warning.

daira (commented):
Post-hoc ACK with comments.

PR #24, opened 2025-02-07
[#23] MacOS `cargo test` build fails

### LINKED ISSUE #23: MacOS `cargo test` build fails
`cargo tests` fails with:

```
   Compiling zallet v0.0.0 (/Users/pacu/Repos/Zcash/wallet/zallet)
error[E0433]: failed to resolve: use of undeclared crate or module `xdg`
   --> zallet/src/commands/migrate_zcash_conf.rs:200:9
    |
200 |         xdg::BaseDirectories::new()
    |         ^^^ use of undeclared crate or module `xdg`

```

pacu:
Adds xdg crate reference to Cargo workspace and toml 
closes #23

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ tokio = "1"
 
 # Filesystem
 home = "0.5"
+xdg = "2.5"
 
 # Localization
 i18n-embed = { version = "0.15", features = ["fluent-system"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3530,6 +3530,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "xdg"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -23,6 +23,7 @@ tokio = "1"
 
 # Filesystem
 home = "0.5"
+xdg = "2.5"
```
str4d, line 26: I also see now that the `known_folders` dependency is missing which breaks compilation on Windows. But as this PR is macOS-specific, I'll fix that in a separate PR.

str4d (approved):
utACK e7d99e15b36f1889015e1ef9de431e9204b7766a

PR #26, opened 2025-02-07
Add basic ci

pacu:
builds atop #24

### DIFF
--- a/.github/workflows/ci-basic-build.yml
+++ b/.github/workflows/ci-basic-build.yml
@@ -1,0 +1,51 @@
+# This is a boring build workflow that just builds the project on
+# mac, ubuntu and windows. Just to check that the optional compiler
+# flags that don't work on a developer's computer still build on 
+# other OS. 
+name: Boring Build
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**.rs'
+      - '**.toml'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '**.rs'
+      - '**.toml'
+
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-basic-build.yml
```diff
@@ -0,0 +1,51 @@
...
+# other OS. 
+name: Boring Build
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**.rs'
+      - '**.toml'
```
str4d, line 13: These path filters are insufficient, and in general I prefer to exclude specific files from affecting general builds (e.g. for doc-only changes).

.github/workflows/ci-basic-build.yml
```diff
@@ -0,0 +1,51 @@
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Use rust-toolchain.toml
+        run: rustup show
```
str4d, line 36: This does not work. We have a `rust-toolchain.toml` that sets an MSRV, and it is used automatically; `rust show` only checks what is being used. The `dtolnay/rust-toolchain` action only ensures the specified Rust version is available; you need additional steps to actually use it.

str4d (commented):
NACK, but thanks for the reminder; I will set up comprehensive CI today to ensure that things are caught going forward.

PR #30, opened 2025-02-24
deps: avoid duplicate crates

rex4539:


str4d:
These aren't duplicates, and I'm fine with where they are currently defined.

PR #33, opened 2025-03-07
Account RPCs

str4d:


zallet/src/components/json_rpc/methods.rs
```diff
@@ -17,6 +22,47 @@ pub(crate) trait Rpc {
...
+        account: JsonValue,
+        receiver_types: Option<Vec<String>>,
+        diversifier_index: Option<u128>,
+    ) -> get_address_for_account::Response;
+
+    /// Lists the addresses managed by this wallet by source, including those generated
+    /// from randomness by this wallet, Sapling addresses generated from the legacy HD
+    /// seed, imported watchonly transparent addresses, shielded addresses tracked using
+    /// imported viewing keys, and addresses derived from the wallet's mnemonic seed for
+    /// releases version 4.7.0 and above.
```
str4d, line 54: This text is copied from `zcashd`, but will need revising once we've determined what we're actually supporting.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+            // zcash_client_sqlite only returns `Ok(None)` if the diversifier index is
+            // invalid for one of the requested receiver types.
+            None => Err(LegacyCode::Wallet.with_message(format!(
+                "Error: diversifier index {} cannot generate an address with the requested receivers.",
+                u128::from(diversifier_index),
+            ))),
+        }
+    } else {
+        wallet
+            .get_next_available_address(account_id, request)
```
nuttycom, line 127: After the changes in https://github.com/zcash/librustzcash/pull/1673, this will not behave as described in the documentation (it will not necessarily produce the next diversifier index in sequence unless the request specifies a transparent receiver.). If the sequentiality of shielded-only diversifier indices is considered essential, then we'll have to add a `force_sequential_diversifier_index_allocation` flag or something of the sort to this API.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+            // This would be a race condition between this and account deletion.
+            .ok_or(RpcErrorCode::InternalError.into())
+    }?;
+
+    // Only include `account` in the response if it was provided in the request. We rely
+    // on the checks performed in `parse_account_parameter`.
+    let account = match account {
+        JsonValue::Number(n) => n.as_u64(),
+        _ => None,
+    };
```
nuttycom, line 138: This is weird behavior; if the caller provided it, why would they need it back? On the other hand, if the caller didn't provide it, then they might need it! My inclination here would be to always return it; that's a strictly additive change, and so it should be safe for clients of the API.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+                unified::Typecode::P2sh =>  LegacyCode::Wallet.with_static(
+                    "Error: P2SH addresses can not be created using this RPC method.",
+                ),
+                _ => LegacyCode::Wallet.with_message(format!(
+                    "Error: receiver type {:?} is not supported.",
+                    typecode,
+                ))
+            },
+            AddressGenerationError::KeyNotAvailable(typecode) => LegacyCode::Wallet.with_message(format!(
+                "Error: account {} does not have a receiver component with type {:?}.",
```
nuttycom, line 181:
```suggestion
                "Error: account {} does not have a key item with type {:?}.",
```

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+                "Error: cannot generate an address containing no shielded receivers."
+            ),
+        },
+        SqliteClientError::AccountUnknown => LegacyCode::Wallet.with_message(format!(
+            "Error: account {} has not been generated by z_getnewaccount.",
+            account
+        )),
+        SqliteClientError::ChainHeightUnknown => LegacyCode::Wallet.with_static(
+            "Error: chain height is unknown. Wait for syncing to progress and then try again."
+        ),
```
nuttycom, line 195: Is this error actually applicable in this context?

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+        SqliteClientError::ReachedGapLimit(_, index) => LegacyCode::Wallet.with_message(format!(
+            "Error: reached the transparent gap limit while attempting to generate a new address at index {}.",
+            index,
+        )),
+        SqliteClientError::DiversifierIndexReuse(diversifier_index, _) => {
+            LegacyCode::Wallet.with_message(format!(
+                "Error: address at diversifier index {} was already generated with different receiver types.",
+                u128::from(diversifier_index),
+            ))
+        }
```
nuttycom, line 205: This behavior also contradicts the documentation from zcashd, which says that it'll give back whatever was previously generated. However, I'm okay with this constricting change as I think that the zcashd behavior is likely a foot-gun.

zallet/src/components/json_rpc/methods/list_addresses.rs
```diff
@@ -0,0 +1,210 @@
...
+                    // Change addresses are never used in a UA receiver, so we need to
+                    // list them separately.
+                    change_addresses.push(taddr.encode(wallet.params()));
+                }
+            } else {
+                // TODO: Provide some way of determining whether this address is a UA receiver.
+            }
+        }
+
+        // TODO: Expose all addresses, not just the last generated.
```
nuttycom, line 124: Add a `librustzcash` issue for the required `zcash_client_backend` functionality.

str4d: https://github.com/zcash/librustzcash/issues/1740

zallet/src/components/json_rpc/utils.rs
```diff
@@ -0,0 +1,83 @@
...
+                })
+            } else {
+                Err(LegacyCode::Wallet.with_static("Account numbers are not supported in wallets with multiple seeds. Use the account UUID instead."))
+            }
+        }
+        // This might be an account UUID.
+        JsonValue::String(s) => s
+            .parse()
+            .map(AccountUuid::from_uuid)
+            .map_err(|_| RpcErrorCode::InvalidParams.into()),
```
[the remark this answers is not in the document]
str4d: I'm confused about what you think is not supported here.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
str4d: One could argue that "valid for the list of receiver types" is sufficient flexibility to let us switch to time-based for shielded-only. But we probably should avoid changing these particular semantics (unless we want to do outreach to the ecosystem participants who have said they are using this API, to determing what aspect of this guarantee they depend on).

Maybe the way to handle this is at time of constructing the `WalletDb`? If we don't have a clock, we can't do time-based address derivation. Then it could be a Zallet config option for whether we allow time-based derivation or not. (Is there anything else we might want the clock for in future, where we couldn't just treat clock presence as equivalent to time-based derivation?)

zallet/src/components/json_rpc/methods/get_address_for_account.rs
str4d: > if the caller provided it, why would they need it back?

It's what the current API does. There's probably rationale on the original PR.

> if the caller didn't provide it, then they might need it!

The caller always provided something. If the caller didn't provide `account` _as a number_, it means they provided it as a UUID, and that we _do_ provide in the response (unconditionally, as we want people to migrate to UUIDs).

My rationale here is that if the caller has migrate to using UUIDs, we don't want to provide them with the old `account` field that previously meant "the number you provided to the call", because it won't necessarily work on a repeat call (if there are multiple seeds).

An alternative approach would be to re-define the type of `account` in the return type similarly to in the request, i.e. it can now be either a number or a string. That would have symmetry, but also be more likely to break things (and we'd still want to provide the UUID next to an account number for forward migration, so then we'd post-migration be duplicating information).

zallet/src/components/json_rpc/methods/get_address_for_account.rs
str4d: Currently no (I think), but in future it could due to how the method is documented:
> If no list of receiver types is given (or the empty list `[]`), the best and second-best shielded receiver types, along with the \"p2pkh\" (i.e. transparent) receiver type, will be used.

This could easily depend on height.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
str4d: I don't believe this contradicts the documentation from zcashd:
> Once a Unified Address has been derived at a specific diversifier index, re-deriving it (via a subsequent call to `z_getaddressforaccount` with the same account and index) will produce the same address with the same list of receiver types. An error will be returned if a different list of receiver types is requested.

The error's documentation:
```rust
    /// The backend encountered an attempt to reuse a diversifier index to generate an address
    /// having different receivers from an address that had previously been exposed for that
    /// diversifier index. Returns the previously exposed address.
```

This means that as long as the same set of receivers is requested, the same address gets generated, otherwise this error occurs.

Where I think the behaviour _might_ differ is that if the user doesn't specify an explicit set (i.e. via `[]`), what zcashd does (IIRC) is return whatever is in the database if it exists, and _otherwise_ generates a new address using the default receiver set. But `WalletWrite::get_address_for_index` (or `WalletRead`) doesn't currently provide a way to just get the previously-derived address for a given index. After https://github.com/zcash/librustzcash/pull/1739 we can fetch all addresses and then filter by diversifier index, but that's a bit heavy. But maybe it's fine?

zallet/src/components/json_rpc/methods/list_addresses.rs
```diff
@@ -0,0 +1,222 @@
...
+            // This would be a race condition between this and account deletion.
+            .ok_or(RpcErrorCode::InternalError)?;
+
+        let addresses = wallet
+            .list_addresses(account.id())
+            .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?;
+
+        let mut transparent_addresses = vec![];
+        let mut transparent_change_addresses = vec![];
+        let mut transparent_ephemeral_addresses = vec![];
```
str4d, line 114: We didn't have ephemeral addresses in `zcashd`, so we need to decide where in the output to place them. Note that for historic reasons, the external and change addresses in the `transparent` block of a source are not subgrouped by account. We could bucket all ephemeral addresses together next to them, or we could create a new part of the structure for them.

zallet/src/components/json_rpc/methods/list_addresses.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// The receiver types that the UA contains (valid values are "p2pkh", "sapling", "orchard").
+    receiver_types: Vec<String>,
+
+    /// The unified address corresponding to the diversifier.
+    address: String,
+}
+
+pub(crate) fn call(wallet: &WalletConnection) -> Response {
+    let mut imported_watchonly = AddressSource::empty("imported_watchonly");
+    let mut mnemonic_seed = AddressSource::empty("mnemonic_seed");
```
str4d, line 96: `zcashd` has 5 defined sources:
- `imported`: addresses that correspond to imported standalone spending keys for transparent, Sprout, or Sapling.
- `imported_watchonly`: imported standalone transparent addresses, or addresses derived from imported Sprout or Sapling viewing keys.
- `legacy_random`: transparent addresses generated by the wallet via the keypool and Sprout addresses for which we have the spending key.
- `legacy_hdseed`: addresses derived from a raw BIP 32 seed. Sapling-only.
- `mnemonic_seed`: addresses derived from a BIP 39 mnemonic phrase.

In mapping these to Zallet:
- I've extended `imported_watchonly` to include imported Unified Viewing Keys (which we never supported in `zcashd`).
- We currently cannot represent `imported` or `legacy_random` in `zcash_client_sqlite`, because all addresses must have an associated account, and all accounts must have a UIVK.
- We currently cannot correctly represent `imported_watchonly` standalone transparent addresses in `zcash_client_sqlite`. What the current Zallet impl adds as transparent addresses, are derived addresses with a known `diversifier_index` (which must be non-null in the database).
- We cannot distinguish `legacy_hdseed` from `mnemonic_seed` using `zcash_client_sqlite` alone (but potentially can record this distinguisher in the Zallet keystore DB, once that is implemented).

nuttycom: A related note: https://github.com/zcash/librustzcash/pull/1739#discussion_r2003951283

zallet/src/components/json_rpc/methods/get_address_for_account.rs
daira: Any form of decorrelation to reduce timing metadata leakage might need a clock. I'd prefer making that an explicit flag (or however we handle other config options).

zallet/src/components/json_rpc/methods/get_address_for_account.rs
daira: Another one for `zcashd-compatibility-quirks`?

zallet/src/components/json_rpc/methods/get_address_for_account.rs
nuttycom: Part of the issue I had hoped to avoid with the time-based address generation is the question of what to do if a caller requests > GAP_LIMIT shielded-only addresses in a row. If we "use up" transparent child indices with shielded address derivation, then we'll end up in a situation where someone's desire to generate a new shielded-only address conflicts with the transparent gap limit. 

Suffice it to say that I think we're better off with the new `librustzcash` behavior, and should just document it.

zallet/src/components/json_rpc/utils.rs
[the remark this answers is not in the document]
nuttycom: Oh, sorry, I misread this. This is just the parser; I think that I had been interpreting this as missing functionality.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
daira: Just document the known difference from zcashd.

zallet/src/components/json_rpc/methods/list_addresses.rs
nuttycom: Ephemeral addresses are explicitly a per-account concern, so we should not group them with the historic addresses.

daira (approved):
utACK

zallet/src/components/json_rpc/methods/list_addresses.rs
nuttycom: > Maybe add a parameter to the RPC to split out {external, change, ephemeral}, but bucket them together by default?

We already put per-account transparent addresses under the correct per-account section. There will never be any ephemeral addresses associated with the legacy account; we don't need to group them together.

> Note that for historic reasons, the external and change addresses in the transparent block of a source are not subgrouped by account.

I thought that that was _only_ true for the legacy random source? If we failed to group per-account transparent addresses with the rest of the account's addresses, that was a major miss.

nuttycom (approved):
utACK 82cf348696db24c4e4989e3c3f49e52e2ba9d4ac with comments

zallet/src/components/json_rpc/methods/list_addresses.rs
str4d: > > There's a question to answer of whether or not we support migrating purely-random addresses or whether we require wallets to migrate away from them.
>
> [..] These addresses will continue to receive funds so long as the outputs to the transparent protocol are allowed on the network, and Zallet must be able to spend those funds.

I don't necessarily agree; exchanges can and do rotate user-facing addresses. And we know of at least one exchange that does not honor funds sent to unsupported public-facing addresses that were previously supported.

What I'd prefer to do here is open an issue tracking this specific capability (support for importing arbitrary standalone transparent spending keys), and then ask ecosystem entities whether that capability is blocking for them. Lack of this capability should not block other progress on Zallet (in particular because this capability requires significant changes to `zcash_client_sqlite`).

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+    },
+};
+
+/// Response to a `z_getaddressforaccount` RPC request.
+pub(crate) type Response = RpcResult<Address>;
+
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct Address {
+    /// The account's UUID within this Zallet instance.
+    uuid: String,
```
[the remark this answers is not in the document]
str4d: Yes, the names of these fields are publicly exposed as the JSON keys in the JSON-RPC output. You'll notice I explicitly rename some of these in `serde` (where the existing `zcashd`-inherited name uses camelCase).

Note however that the `Address` struct name is _not_ exposed publicly. The name `Address` is purely an internal helper for preparing the JSON-RPC response type.

As for the choice of `uuid` here, we do not currently have any such field in the `zcashd` wallet output, so we can pick whatever name we want here. Generally in `zcashd` we picked something that would be consistently used across all occurrences of it. So I'd be fine naming this `account_uuid`, or probably `accountUuid` for better consistency in the output.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
```diff
@@ -0,0 +1,208 @@
...
+                        "are invalid receiver types"
+                    }
+                )))
+            }
+        }
+        _ => {
+            // zcashd default is the best and second-best shielded receiver types, and the
+            // transparent (P2PKH) receiver type. That currently corresponds to all possible
+            // receiver types.
+            let request = UnifiedAddressRequest::unsafe_custom(
```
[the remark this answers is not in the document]
str4d: `REQUIRE_ALL` is not the same thing as the documented `z_getaddressforaccount` default. They currently happen to be the same, but that will not always be the case.

zallet/src/components/json_rpc/methods/list_addresses.rs
str4d: AFAICT from the `zcashd` impl, we did indeed fail to group per-account transparent addresses with the rest of the account's addresses. PR that added this logic is here: https://github.com/zcash/zcash/pull/5681

zallet/src/components/json_rpc/methods/list_addresses.rs
nuttycom: > And we know of at least one exchange that does not honor funds sent to unsupported public-facing addresses that were previously supported.

If you're referring to Binance, this I believe that this isn't currently the case, and that they still honor funds sent to pre-ZIP-320 addresses. In any case, I do not think that we can reasonably expect all exchanges to have fully migrated all users off of all randomly-derived keys by the time that we need NU7 to activate.

zallet/src/components/json_rpc/methods/get_address_for_account.rs
str4d: > We might want to consider adding another optional argument to this API for the seed fingerprint; then, if someone provides both a ZIP 32 id and the seed fingerprint, we can unambiguously determine what account it's referring to.

Opened #40.

str4d: > I suggest adding an "account_uuid" field to the response and making both that and "account" always present.

I'm doing the former but not (for now) doing the latter.

> If there are clients that don't tolerate that, gate the old behaviour on `zcashd-compatibility-quirks`. We never guaranteed that new fields wouldn't be added to the response (or that there wouldn't be new situations in which a field is present).

I was not arguing that the presence of `account_uuid` would be a problem; I agree adding fields is always fine. Removing an existing field is what can break things, which is why I only remove `account` if it is clear the caller has already migrated to the new behaviour.

If we decide to do #40 then I'd be open to always having `account` present (though I'd prefer we instead rename it).

zallet/src/components/json_rpc/methods/get_address_for_account.rs
str4d: Going with "cannot generate a receiver component" instead as that is clearer to the user (the RPC docs talk about receiver types, not key items).

zallet/src/components/json_rpc/methods/list_addresses.rs
str4d: I'm fixing this by adding a new `derivedTransparent` array to each source. `transparent` will only contain transparent addresses for which we do not know derivation information.

str4d:
Addressed review comments.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/47234d234ca4b0dc26215a9b4bbb60aae6785afe..ed6e3fd18fc9744ec14e9824f1b24d73940c0893) to remove TODO (I confirmed that `zcashd` uses Require semantics).

nuttycom (approved):
utACK ed6e3fd18fc9744ec14e9824f1b24d73940c0893

PR #34, opened 2025-03-07
Misc fixes

str4d:


zallet/src/components/json_rpc/server/error.rs
```diff
@@ -51,6 +51,40 @@ pub enum LegacyCode {
...
+    WalletPassphraseIncorrect = -14,
+    /// Command given in wrong wallet encryption state (encrypting an encrypted wallet etc.)
+    WalletWrongEncState = -15,
+    /// Failed to encrypt the wallet
+    WalletEncryptionFailed = -16,
+    /// Wallet is already unlocked
+    WalletAlreadyUnlocked = -17,
+    /// User must acknowledge backup of the mnemonic seed.
+    WalletBackupRequired = -18,
+}
```
daira, line 76: This doesn't have the standard JSON-RPC 2.0 errors:

https://github.com/zcash/zcash/blob/a3435336b0c561799ac6805a27993eca3f9656df/src/rpc/protocol.h#L35-L40

zallet/src/components/wallet/connection.rs
```diff
@@ -37,6 +38,10 @@ pub(super) type WalletPool = deadpool::managed::Pool<WalletManager>;
 
 pub(crate) struct WalletManager {
     inner: deadpool_sqlite::Manager,
+    /// Connection pools are thread-safe, but SQLite does not reliably follow the busy
+    /// handler (configured by `rusqlite` to a timeout after 5s), so we explicitly guard
+    /// against SQLite `DatabaseBusy` errors.
```
daira, line 43: What does "not reliably follow the busy handler" mean? (I understand that using this guard will avoid `DatabaseBusy` errors; I just don't understand the wording about what will happen if we don't.)

daira (approved):
utACK with comments.

nuttycom (approved):
utACK 91ff1fe7d74e320f8045387cfaf45e280eb956db

zallet/src/components/json_rpc/server/error.rs
str4d: Those are provided by `jsonrpsee` IIRC

zallet/src/components/wallet/connection.rs
str4d: https://www.sqlite.org/c3ref/busy_handler.html

> The presence of a busy handler does not guarantee that it will be invoked when there is lock contention. If SQLite determines that invoking the busy handler could result in a deadlock, it will go ahead and return SQLITE_BUSY to the application instead of invoking the busy handler.

PR #36, opened 2025-03-15
Add `gettransaction`

### LINKED ISSUE #38: Add `gettransaction` and `z_viewtransaction`
These are the APIs used to expose detailed information about wallet transactions (the first shows public information, the second shows information that is encrypted on chain).

str4d:
Closes zcash/wallet#38.

### DIFF
--- a/zallet/src/components/json_rpc/balance.rs
+++ b/zallet/src/components/json_rpc/balance.rs
@@ -99,9 +99,9 @@ pub(super) fn wtx_get_value_out(tx: &Transaction) -> Option<Zatoshis> {
                 .into_iter()
                 .flat_map(|bundle| bundle.vout.iter().map(|txout| txout.value)),
         )
-        // NB: negative valueBalanceSapling "takes" money from the transparent value pool just as outputs do
+        // Note: negative valueBalanceSapling "takes" money from the transparent value pool just as outputs do
         .chain((-tx.sapling_value_balance()).try_into().ok())
-        // NB: negative valueBalanceOrchard "takes" money from the transparent value pool just as outputs do
+        // Note: negative valueBalanceOrchard "takes" money from the transparent value pool just as outputs do
         .chain(
             tx.orchard_bundle()
                 .and_then(|b| (-*b.value_balance()).try_into().ok()),
--- a/zallet/src/components/json_rpc/balance.rs
+++ b/zallet/src/components/json_rpc/balance.rs
@@ -1,0 +1,293 @@
+use rusqlite::named_params;
+use transparent::bundle::TxOut;
+use zaino_state::{FetchServiceSubscriber, MempoolKey};
+use zcash_client_backend::data_api::WalletRead;
+use zcash_client_sqlite::error::SqliteClientError;
+use zcash_keys::encoding::AddressCodec;
+use zcash_primitives::transaction::Transaction;
+use zcash_protocol::{consensus::BlockHeight, value::Zatoshis};
+
+use crate::components::database::DbConnection;
+
+/// Coinbase transaction outputs can only be spent after this number of new blocks
+/// (network rule).
+const COINBASE_MATURITY: u32 = 100;
+
+enum IsMine {
+    Spendable,
+    WatchOnly,
+    Either,
+}
+
+/// Returns `true` if this output is owned by some account in the wallet and can be spent.
+pub(super) fn is_mine_spendable(
+    wallet: &DbConnection,
+    tx_out: &TxOut,
+) -> Result<bool, SqliteClientError> {
+    is_mine(wallet, tx_out, IsMine::Spendable)
+}
+
+/// Returns `true` if this output is owned by some account in the wallet, but cannot be
+/// spent (e.g. because we don't have the spending key, or do not know how to spend it).
+#[allow(dead_code)]
+pub(super) fn is_mine_watchonly(
+    wallet: &DbConnection,
+    tx_out: &TxOut,
+) -> Result<bool, SqliteClientError> {
+    is_mine(wallet, tx_out, IsMine::WatchOnly)
+}
+
+/// Returns `true` if this output is owned by some account in the wallet.
+pub(super) fn is_mine_spendable_or_watchonly(
+    wallet: &DbConnection,
+    tx_out: &TxOut,
+) -> Result<bool, SqliteClientError> {
+    is_mine(wallet, tx_out, IsMine::Either)
+}
+
+/// Logically equivalent to [`IsMine(CTxDestination)`] in `zcashd`.
+///
+/// [`IsMine(CTxDestination)`]: https://github.com/zcash/zcash/blob/2352fbc1ed650ac4369006bea11f7f20ee046b84/src/script/ismine.cpp#L121
+fn is_mine(
+    wallet: &DbConnection,
+    tx_out: &TxOut,
+    include: IsMine,
+) -> Result<bool, SqliteClientError> {
+    match tx_out.recipient_address() {
+        Some(address) => wallet.with_raw(|conn| {
+            let mut stmt_addr_mine = conn.prepare(
+                "SELECT EXISTS(
+                    SELECT 1
+                    FROM addresses
+                    JOIN accounts ON account_id = accounts.id
+                    WHERE address = :address
+                    AND (
+                        :allow_either = 1
+                        OR has_spend_key = :has_spend_key
+                    )
+                )",
+            )?;
+
+            Ok(stmt_addr_mine.query_row(
+                named_params! {
+                    ":address": address.encode(wallet.params()),
+                    ":allow_either": matches!(include, IsMine::Either),
+                    ":has_spend_key": matches!(include, IsMine::Spendable),
+                },
+                |row| row.get(0),
+            )?)
+        }),
+        // TODO: Use `zcash_script` to discover other ways the output might belong to
+        // the wallet (like `IsMine(CScript)` does in `zcashd`).
+        None => Ok(false),
+    }
+}
+
+/// Equivalent to [`CTransaction::GetValueOut`] in `zcashd`.
+///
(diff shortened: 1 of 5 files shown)


zallet/src/components/json_rpc/balance.rs
```diff
@@ -0,0 +1,293 @@
...
+            for spend in bundle.shielded_spends() {
+                if stmt_note_exists
+                    .query_row(named_params! {":nf": spend.nullifier().0}, |row| row.get(0))?
+                {
+                    return Ok(true);
+                }
+            }
+        }
+
+        // TODO: Fix bug in `zcashd` where we forgot to add Orchard here.
```
str4d, line 213: Fixed in https://github.com/zcash/zcash/pull/7035.

zallet/src/components/json_rpc/balance.rs
```diff
@@ -0,0 +1,293 @@
...
+        // TODO: Use `zcash_script` to discover other ways the output might belong to
+        // the wallet (like `IsMine(CScript)` does in `zcashd`).
+        None => Ok(false),
+    }
+}
+
+/// Equivalent to [`CTransaction::GetValueOut`] in `zcashd`.
+///
+/// [`CTransaction::GetValueOut`]: https://github.com/zcash/zcash/blob/2352fbc1ed650ac4369006bea11f7f20ee046b84/src/primitives/transaction.cpp#L214
+pub(super) fn wtx_get_value_out(tx: &Transaction) -> Option<Zatoshis> {
```
nuttycom, line 89: The semantics of this whole method are very weird; how would someone reasonably use this value?

zallet/src/components/json_rpc/balance.rs
```diff
@@ -0,0 +1,293 @@
...
+        // Equivalent to `CWallet::GetDebit(CTransaction)` in `zcashd`.
+        Some(bundle) => {
+            let mut acc = Some(Zatoshis::ZERO);
+            for txin in &bundle.vin {
+                // Equivalent to `CWallet::GetDebit(CTxIn)` in `zcashd`.
+                if let Some(txout) = wallet
+                    .get_transaction(*txin.prevout.txid())?
+                    .as_ref()
+                    .and_then(|prev_tx| prev_tx.transparent_bundle())
+                    .and_then(|bundle| bundle.vout.get(txin.prevout.n() as usize))
```
nuttycom, line 134: bug: This seems unusable in the context of `zallet`; in zcashd, this behavior was okay because you could be confident that you had access to the transaction inputs; here, this will just silently ignore any value for inputs for which you don't have the full transaction.

nuttycom (changes requested):
It seems to me that some of the balance computation performed here *cannot* be made reliable in `zallet`. I question whether it's even a good idea to implement this method.

zallet/src/components/json_rpc/balance.rs
```diff
@@ -0,0 +1,293 @@
...
+        Some(address) => wallet.with_raw(|conn| {
+            let mut stmt_addr_mine = conn.prepare(
+                "SELECT EXISTS(
+                    SELECT 1
+                    FROM addresses
+                    JOIN accounts ON account_id = accounts.id
+                    WHERE address = :address
+                    AND (
+                        :allow_either = 1
+                        OR has_spend_key = :has_spend_key
```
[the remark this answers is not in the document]
str4d: > Also, what are the semantics of IsMine for P2SH outputs where you hold one, but not all, of the keys?

`false`: https://github.com/zcash/zcash/blob/18238d90cd0b810f5b07d5aaa1338126aa128c06/src/script/ismine.cpp#L94

```cpp
        // Only consider transactions "mine" if we own ALL the
        // keys involved. Multi-signature transactions that are
        // partially owned (somebody else has a key that can spend
        // them) enable spend-out-from-under-you attacks, especially
        // in shared-wallet situations.
```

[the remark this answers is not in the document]
daira: Please reproduce that comment in the Rust code.

zallet/src/components/json_rpc/balance.rs
```diff
@@ -0,0 +1,293 @@
...
+/// Returns `true` if this output is owned by some account in the wallet.
+pub(super) fn is_mine_spendable_or_watchonly(
+    wallet: &DbConnection,
+    tx_out: &TxOut,
+) -> Result<bool, SqliteClientError> {
+    is_mine(wallet, tx_out, IsMine::Either)
+}
+
+/// Logically equivalent to [`IsMine(CTxDestination)`] in `zcashd`.
+///
```
daira, line 49:
```suggestion
///
/// A transaction is only considered "mine" by virtue of having a P2SH multisig
/// output if we own *all* of the keys involved. Multi-signature transactions that
/// are partially owned (somebody else has a key that can spend them) enable
/// spend-out-from-under-you attacks, especially in shared-wallet situations.
/// Non-P2SH ("bare") multisig outputs never make a transaction "mine".
///
```

zallet/src/components/json_rpc/balance.rs
oxarbitrage: Removed TODO at https://github.com/zcash/wallet/pull/36/commits/ac1011a286ce986c173e4f2b933e27b4bd038410

zallet/src/components/json_rpc/balance.rs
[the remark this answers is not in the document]
oxarbitrage: https://github.com/zcash/wallet/pull/36/commits/8aca24f6b4ba59df13e339f158538885bc1d282a and https://github.com/zcash/wallet/pull/36/commits/790f4a25ed2b3cacbd89265d74afffc4e748fb58

zallet/src/components/json_rpc/balance.rs
oxarbitrage: Can you provide a suggestion on what to do here?

zallet/src/components/json_rpc/methods/get_transaction.rs
```diff
@@ -0,0 +1,360 @@
...
+                .with_message(format!("Failed to fetch transaction: {}", e.to_string()))
+        })?
+        .ok_or(LegacyCode::InvalidParameter.with_static("Invalid or non-wallet transaction id"))?;
+
+    // TODO: In zcashd `filter` is for the entire transparent wallet. Here we have multiple
+    // mnemonics; do we have multiple transparent buckets of funds?
+
+    // `gettransaction` was never altered to take wallet shielded notes into account.
+    // As such, its `amount` and `fee` fields are calculated as if the wallet only has
+    // transparent addresses.
```
nuttycom, line 164: blocking: This has been a potential source of loss of funds for years, and we should not reproduce these bugs. Instead of modifying the semantics of this method, we should not include it in the `Zallet` alpha, and instead provide a replacement in `z_viewtransaction` that we can encourage users to migrate to instead.

zallet/src/components/json_rpc/balance.rs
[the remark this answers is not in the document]
nuttycom: Yes; we won't close this PR but we'll probably move it back to draft and keep it in our pocket in case the push to get users to migrate to `z_viewtransaction` is unsuccessful. I need to re-review what is reported by `gettransaction` and make a concrete plan for incorporating this into `z_viewtransaction`.

zallet/src/components/json_rpc/methods/get_transaction.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    expiryheight: Option<u64>,
+
+    walletconflicts: Vec<String>,
+
+    /// The transaction time in seconds since epoch (1 Jan 1970 GMT).
+    time: u64,
+
+    /// The time received in seconds since epoch (1 Jan 1970 GMT).
```
daira, line 87:
```suggestion
    /// The time received in seconds since epoch (1 Jan 1970 UTC).
```

zallet/src/components/json_rpc/balance.rs
daira: Notice the TODO at https://github.com/zcash/wallet/pull/36/files#diff-aa4c5eb784ec34fa821ac213851b030ecd0d609b8d5b9f94252db39e476b68f6R191 , just before the only call to this method. I think the best way to proceed is to remove this method and fix that TODO in the way it suggests.

daira (commented):
Reviewed with comments.

daira:
> It seems NB is short for "nota bene", Latin for "note well". This was too obscure for me, i changed it but can go back or change to something else.

NB used to be quite common at least in British English; I remember using it quite a bit but it has fallen out of favour. Google [Ngrams confirms this impression](https://books.google.com/ngrams/graph?content=NB&year_start=1800&year_end=2022&corpus=en&smoothing=3).

oxarbitrage:
Closing this as superseded.

Per the design decision reached in this thread (see https://github.com/zcash/wallet/pull/36#discussion threads from @nuttycom, 2025-08-20), `gettransaction` is **not** being included in Zallet: it inherits zcashd's balance/`IsMine` semantics that can't be made reliable in Zallet (we don't always hold the full input transactions), which is a known loss-of-funds footgun we don't want to reproduce.

The agreed replacement was to extend `z_viewtransaction` to cover this method's use cases in a non-buggy way. That work has since landed:
- #147 — add transparent-equivalent details to `z_viewtransaction` (closed)
- #228 — top-level map describing the effect on each involved account's balance (closed)
- #233 — report account UUID with each spend/receive entry (closed)

Issue #38, which this PR originally closed, was also edited to drop `gettransaction` and now only covers `z_viewtransaction` (closed).

Given the above, plus ~360 commits of drift and conflicts against `main`, there's no remaining value in reviving this branch. Anyone needing per-transaction detail should use `z_viewtransaction` (tracked under #156 / #440 / #450). Thanks @str4d for the original work — the useful pieces live on in `z_viewtransaction`.

PR #39, opened 2025-03-15
Sync wallet using a full node via Zaino instead of a lightwalletd

### LINKED ISSUE #37: Add a `Chain` component that wraps Zaino
Various parts of Zallet need access to chain information. Currently the only place is the sync engine, which calls out to lightwalletd temporarily, but we're also going to need it for various RPC methods like `z_getnewaccount` (to fetch the starting treestate). Per #7 the plan is for Zaino to serve this data.

This issue can be closed once we have a `Chain` component (or similar) that wraps Zaino as a library and can communicate with a full node.

str4d:
Closes #37.


str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/9f1ab4c7f3d4bc9196d13bbdb88d80734b43af59..35df94689f9946a3ace5136706eb0cefb229d154) to fix several bugs introduced by #44 (as well as make some cleanups). I could not fix one of the bugs in Zallet; this PR is now blocked on https://github.com/zingolabs/zaino/issues/254 unless I undo some of the changes from #44 and switch back to my previous approach.

str4d:
[Rebased](https://github.com/zcash/wallet/compare/35df94689f9946a3ace5136706eb0cefb229d154..50e90c99d7b394565924ef7a39c5ab3b359b3bad) on #45.

str4d:
[Rebased](https://github.com/zcash/wallet/compare/50e90c99d7b394565924ef7a39c5ab3b359b3bad..0326fe8f1558cc77cb3832695e1307dc42f6a47c) on #46.

Cargo.toml
```diff
@@ -66,6 +66,16 @@ transparent = { package = "zcash_transparent", version = "0.2" }
...
 
+# Zcash chain state
+zaino-fetch = { git = "https://github.com/zingolabs/zaino.git", rev = "2b2ebb4725d771724c724587e2dfe44af7b46377" }
+zaino-proto = { git = "https://github.com/zingolabs/zaino.git", rev = "2b2ebb4725d771724c724587e2dfe44af7b46377" }
+zaino-state = { git = "https://github.com/zingolabs/zaino.git", rev = "2b2ebb4725d771724c724587e2dfe44af7b46377" }
+# - We do not use branch dependencies in general because they are brittle, but are forced
+#   to do so here because Zaino does.
+zebra-chain = { git = "https://github.com/ZcashFoundation/zebra.git", branch = "main" }
+zebra-rpc = { git = "https://github.com/ZcashFoundation/zebra.git", branch = "main" }
+zebra-state = { git = "https://github.com/ZcashFoundation/zebra.git", branch = "main" }
```
str4d, line 77: We gain an awful lot of dependencies as a result of these, including a bunch of duplicate Zcash crates. It would be preferable to wait until at least either https://github.com/ZcashFoundation/zebra/pull/9300 or https://github.com/ZcashFoundation/zebra/pull/9369 is merged.

str4d:
[Rebased](https://github.com/zcash/wallet/compare/df1ed20fd12b90583f1d1ab98d1c77a44d0f1580..113a6152fb5db6fb0c1506988a17e2620ea0e0bf) to include the latest changes to #46.

str4d:
[Rebased](https://github.com/zcash/wallet/compare/8b7789ff5df5f031e034866900cfbb5299a1574b..ed6449c134c6e8fde99cc314fc367b4258341ebb) on top of #100 (which includes latest `main`).

deny.toml
```diff
@@ -20,8 +20,25 @@ allow = [
...
+    { name = "icu_normalizer_data", allow = ["Unicode-3.0"] },
+    { name = "icu_properties", allow = ["Unicode-3.0"] },
+    { name = "icu_properties_data", allow = ["Unicode-3.0"] },
+    { name = "icu_provider", allow = ["Unicode-3.0"] },
+    { name = "icu_provider_macros", allow = ["Unicode-3.0"] },
+    { name = "libloading", allow = ["ISC"] },
+    { name = "litemap", allow = ["Unicode-3.0"] },
     { name = "matchit", allow = ["BSD-3-Clause"] },
+    # Copyleft license. Temporary exception until Zebra stops depending on `dirs`.
+    { name = "option-ext", allow = ["MPL-2.0"] },
```
str4d, line 40: Turns out `zebra-chain`, `zebra-network`, and `zebra-state` depend on `dirs`, which has depended on a copyleft license since 2023. Zebra already uses `cargo-deny` and opened https://github.com/ZcashFoundation/zebra/pull/3397 in 2022 to add license denies (which would have caught this), but closed the PR without merging it and never resumed the work.

https://github.com/zcash/librustzcash/pull/864 is how I fixed this in the Zcash Rust crates; I will propose a similar change to Zebra before we cut the first alpha of Zallet.

daira: This should be fixed before merging this PR. It's technically a license violation to publish this crate under MIT/Apache while it depends on an MPL crate.

zallet/src/components/chain_view.rs
```diff
@@ -0,0 +1,127 @@
...
+            config.indexer.map_shard_amount,
+            db_path,
+            config.indexer.db_size,
+            config.network().to_zebra(),
+            false,
+            // Setting this to `true` causes start-up to block on completely filling the
+            // cache. Zaino's DB currently only contains a cache of CompactBlocks, so we
+            // make do for now with uncached queries.
+            // TODO: https://github.com/zingolabs/zaino/issues/249
+            true,
```
daira, line 69: "We make do for now with uncached queries" would imply to me that this is being set to false to avoid the performance problem on startup, but it is set to true, so the comment doesn't seem to match the code.

zallet/src/components/sync.rs
```diff
@@ -1,45 +1,323 @@
...
+    //     // The wallet either hasn't been synced before, or hasn't scanned any blocks. In
+    //     // either case, the wallet has no state that depends on a particular block being
+    //     // present, so we can just notify it of the current chain tip.
+    //     _ => {
+    //         let current_tip = steps::get_chain_tip(&chain).await?;
+    //         info!("Latest block height is {}", current_tip.height);
+    //         db_data.update_chain_tip(current_tip.height)?;
+    //         current_tip
+    //     }
+    // };
```
daira, line 105: Why is this commented out?

zallet/src/components/sync.rs
```diff
@@ -1,45 +1,323 @@
...
+        let consensus_branch_id = consensus::BranchId::for_height(params, mempool_height);
+        let mut mempool_stream = chain.get_mempool_stream().await?;
+        while let Some(result) = mempool_stream.next().await {
+            match result {
+                Ok(raw_tx) => {
+                    let tx = Transaction::read(
+                        SerializedTransaction::from(raw_tx.data).as_ref(),
+                        consensus_branch_id,
+                    )
+                    .expect("Zaino should only provide valid transactions");
```
daira, line 225: This should not be a panic. It might be okay to exit but the exit shouldn't be here, it should propagate as a SyncError.

zallet/src/components/sync.rs
str4d: I was just in the middle of fixing this TODO. It's done now.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/d1cbaf7e53e3cee2c8365c7a096dde2577efc511..6b4cca155cbd11df161ade77ce607ec032b23d28) to fix a TODO I'd planned to do before review (vs the other TODOs which will mostly be addressed in future PRs).

zallet/src/components/chain_view.rs
str4d: The comment is incorrect here, not the code.

Cargo.toml
str4d: I've addressed all the dependency reduction that I can achieve in the scope of this PR.

zallet/src/components/sync/steps.rs
```diff
@@ -0,0 +1,345 @@
...
+    while prev_tip != current_tip {
+        // We are fetching blocks from the same data source, and we assume here that the
+        // data source has one single block with no parent (the genesis block). Therefore
+        // if the blocks are not currently equal, they cannot be the genesis block, and so
+        // their parent hashes exist (per the `ChainBlock` invariant).
+        prev_tip =
+            ChainBlock::resolve(chain, prev_tip.prev_hash.expect("present by invariant")).await?;
+        current_tip =
+            ChainBlock::resolve(chain, current_tip.prev_hash.expect("present by invariant"))
+                .await?;
```
daira, line 188: These also should not be panics.

zallet/src/config.rs
```diff
@@ -142,6 +146,45 @@ impl BuilderSection {
...
+
+    /// Full node / validator Username.
+    pub validator_user: Option<String>,
+
+    /// Full node / validator Password.
+    pub validator_password: Option<String>,
+
+    /// Capacity of the Dashmaps used for the Mempool.
+    /// Also use by the BlockCache::NonFinalisedState when using the FetchService.
+    pub map_capacity: Option<usize>,
```
daira, line 170: Capacity in what unit? Consider adding the unit to the name.

zallet/src/config.rs
```diff
@@ -142,6 +146,45 @@ impl BuilderSection {
...
+    /// If a shard_amount which is not a power of two is provided, the function will panic.
+    pub map_shard_amount: Option<usize>,
+
+    /// Block Cache database file path.
+    ///
+    /// This is Zaino's Compact Block Cache db if using the FetchService or Zebra's RocksDB if using the StateService.
+    pub db_path: Option<PathBuf>,
+
+    /// Block Cache database maximum size in GB.
+    pub db_size: Option<usize>,
```
daira, line 185:
```suggestion
    pub db_size_gb: Option<usize>,
```
(or `db_size_gib` if measured in GiB).

daira (changes requested):
Some blocking comments.

deny.toml
```diff
@@ -30,5 +47,12 @@ exceptions = [
     { name = "unicode-ident", allow = ["Unicode-3.0"] },
     { name = "untrusted", allow = ["ISC"] },
     { name = "webpki-roots", allow = ["MPL-2.0"] },
```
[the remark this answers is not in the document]
str4d: Oh wait, it was actually mis-licensed in the crate metadata. The commit immediately after the latest release fixes it: https://github.com/rustls/webpki-roots/commit/90c48f38672f4300eea5f1b712d4ccca8ad82123

So we can continue to depend on `webpki-roots` as-is for now, and this exception will be updated to `CDLA-Permissive-2.0` in the next crate update.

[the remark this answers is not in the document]
daira: https://github.com/rustls/webpki-roots/blob/main/webpki-roots/src/lib.rs, which is the only source file in that crate, is just raw data, and so to me it doesn't seem copyrightable. I doubt that Mozilla could enforce its license on that crate, and if it tried there would be uproar. There isn't a good alternative source for that data as far as I know. So I'm ok with a license exemption in this specific case.

deny.toml
str4d: Hmm, reading Q11 on https://www.mozilla.org/en-US/MPL/2.0/FAQ/ suggests that MPL-2.0 is not quite as viral as other copyleft licenses, and that as long as we aren't trying to modify `option-ext` itself then we aren't required to relicense Zallet.

zallet/src/components/sync/steps.rs
str4d: These panics are fine, because if they fail it is a programming errors (namely a `ChainBlock` being constructed in violation of its documented invariant that `prev_hash` is `None` if and only if `height` is `consensus::H0`).

zallet/src/components/sync.rs
str4d: It is a programming error for Zaino to return an invalid transaction from the mempool stream. Really this is a problem with Zaino not having a typesafe Rust API.

zallet/src/config.rs
```diff
@@ -155,6 +159,45 @@ impl LimitsSection {
...
+
+    /// Full node / validator Password.
+    pub validator_password: Option<String>,
+
+    /// Capacity of the Dashmaps used for the Mempool.
+    /// Also use by the BlockCache::NonFinalisedState when using the FetchService.
+    pub map_capacity: Option<usize>,
+
+    /// Number of shard used in the DashMap used for the Mempool.
+    /// Also use by the BlockCache::NonFinalisedState when using the FetchService.
```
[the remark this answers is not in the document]
str4d: Yes, this was all copy-pasted from Zaino code.

zallet/src/components/sync.rs
```diff
@@ -94,3 +121,275 @@ impl WalletSync {
...
+    //     // The wallet either hasn't been synced before, or hasn't scanned any blocks. In
+    //     // either case, the wallet has no state that depends on a particular block being
+    //     // present, so we can just notify it of the current chain tip.
+    //     _ => {
+    //         let current_tip = steps::get_chain_tip(&chain).await?;
+    //         info!("Latest block height is {}", current_tip.height);
+    //         db_data.update_chain_tip(current_tip.height)?;
+    //         current_tip
+    //     }
+    // };
```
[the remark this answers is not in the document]
str4d: This was left over from a previous approach, and commented out but left in while I had an outstanding TODO. I've now implemented the TODO and deleted this.

zallet/src/components/chain_view.rs
```diff
@@ -0,0 +1,141 @@
...
+                    return Err(ErrorKind::Generic.into());
+                }
+
+                // Check for shutdown signals.
+                if status == StatusType::Closing {
+                    let service = indexer.write().await.take().expect("only happens once");
+                    service.inner().close();
+                    return Ok(());
+                }
+            }
```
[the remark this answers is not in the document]
str4d: This logic comes from Zaino (specifically from how the indexer is managed in `zainod`, because that is what the Zaino docs at the time told me to copy). It does appear to be doing error-condition handling, yes.

deny.toml
nuttycom: Yeah, that was my reading as well, and the MPL FAQ is very clear on this point:
> Q11: How 'viral' is the MPL? If I use MPL-licensed code in my proprietary application, will I have to give all the source code away?
> No. The license requires that Modifications (as defined in Section 1.10 of the license) must be licensed under the MPL and made available to anyone to whom you distribute the Source Code. However, new files containing no MPL-licensed code are not Modifications, and therefore do not need to be distributed under the terms of the MPL, even if you create a Larger Work (as defined in Section 1.7) by using, compiling, or distributing the non-MPL files together with MPL-licensed files. This allows, for example, programs using MPL-licensed code to be statically linked to and distributed as part of a larger proprietary piece of software, which would not generally be possible under the terms of stronger copyleft licenses.

zallet/src/config.rs
```diff
@@ -142,6 +146,31 @@ impl BuilderSection {
     }
 }
 
+/// Indexer configuration section.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct IndexerSection {
+    /// Full node / validator listen port.
+    pub validator_listen_address: Option<SocketAddr>,
```
nuttycom, line 154: This field was identified as being badly named in the Zaino Community signal chat; we should do a better job here (it's not an address that Zaino listens on.)
```suggestion
    pub validator_address: Option<SocketAddr>,
```

nuttycom (approved):
utACK with nonblocking nits & minor documentation requests.

zallet/src/components/sync/steps.rs
```diff
@@ -0,0 +1,345 @@
...
+) -> Result<CompactBlock, SyncError> {
+    let mut block_id = zaino_proto::proto::service::BlockId::default();
+    match hash_or_height {
+        HashOrHeight::Hash(hash) => block_id.hash = hash.0.to_vec(),
+        HashOrHeight::Height(height) => block_id.height = height.0 as u64,
+    };
+
+    let compact_block = chain.get_block(block_id).await?;
+
+    Ok(CompactBlock {
```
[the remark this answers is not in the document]
str4d: Correct; `zaino-proto` and `zcash_client_sqlite::proto`.

Once Zaino provides a full-block index we can use, this will go away.

zallet/src/config.rs
str4d: It's not; it's the address that the validator (Zebra) listens on. Unless you mean that there's potential for confusion about this being an address that Zallet listens on?

zallet/src/components/sync.rs
```diff
@@ -279,18 +279,18 @@ async fn recover_history(
...
+                if let Some((cur, next)) =
+                    remaining.split_at(remaining.block_range().start + batch_size)
+                {
+                    *acc = Some(next);
+                    cur
+                } else {
+                    *acc = None;
+                    remaining
+                }
+            })
```
nuttycom, line 294: Performing mutation to local variables inside of a `map` squicks me:
```suggestion
            if let Some((next, sub_range)) = acc.clone().map(|remaining| {
                if let Some((cur, next)) =
                    remaining.split_at(remaining.block_range().start + batch_size)
                {
                    (Some(next), cur)
                } else {
                    (None, remaining)
                }
            }) {
                *acc = next;
                Some(sub_range)
            } else {
                None
            }
```

nuttycom (approved):
re-utACK 3b98af19fb16488cc3dd63d641cc27077f5bbe49

deny.toml
nuttycom: That's not correct, AFAICT: you have to redistribute the modified version of the MPL-licensed files, but you don't have to relicense:

> Q10: I want to distribute (outside my organization) an executable program based on MPL-licensed source code that I have modified. What do I have to do?
> You must make available the MPL-licensed portions of the source code as described in the previous question, and inform the recipients how they can obtain such source code (Section 3.2).

PR #42, opened 2025-03-21
rpc: Permit JSON-RPC IDs to be non-strings

str4d:
This is ZcashFoundation/zebra#9314 (the code in question was reused in Zallet).


nuttycom (approved):
utACK

PR #44, opened 2025-03-28
use zaino traits

idky137:
Updates code to use the `ZcashIndexer` and `LightwalletIndexer` traits from Zaino-State rather than the internal functionality of the `FetchService`.

### DIFF
--- a/zallet/src/components/sync/steps.rs
+++ b/zallet/src/components/sync/steps.rs
@@ -1,7 +1,11 @@
 use jsonrpsee::tracing::info;
 use orchard::tree::MerkleHashOrchard;
-use zaino_fetch::jsonrpc::{error::JsonRpcConnectorError, response::GetBlockResponse};
-use zaino_state::fetch::FetchServiceSubscriber;
+use zaino_fetch::jsonrpc::error::JsonRpcConnectorError;
+use zaino_state::{
+    error::FetchServiceError,
+    fetch::FetchServiceSubscriber,
+    indexer::{LightWalletIndexer as _, ZcashIndexer as _},
+};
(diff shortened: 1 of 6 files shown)


str4d (approved):
ACK. I don't want to be using lightwalletd-specific methods or have any eventual dependency on the lightwalletd gRPC protocol, but this will do for now if it's how Zaino is currently intended to be used.

str4d:
Something in this PR has broken compilation, but I'll merge it and fix in my branch.

zallet/src/components/sync/steps.rs
```diff
@@ -48,29 +58,53 @@ impl ChainBlock {
...
+                height: 0,
+                hash: hash.0.to_vec(),
+            },
+            HashOrHeight::Height(height) => zaino_proto::proto::service::BlockId {
+                height: height.0 as u64,
+                hash: Vec::new(),
+            },
+        };
+
+        let compact_block = chain.get_block(block_id).await?;
```
str4d, line 96: @idky137 you switched to using this API in multiple places in this PR. After several hours of debugging issues post-merge, I've discovered that this API is incomplete. It assumes that `BlockId.height` is always set, and never checks for the presence of `BlockId.hash`. This is a problem because in gRPC, not setting a field is equivalent to its default value, which means that every by-hash block query is now instead attempting to fetch the genesis block, which then fails somewhere in the depths of either Zaino or the backing full node because something is expecting `fOverwintered` to be set (which is not the case for any genesis block).

str4d: Opened https://github.com/zingolabs/zaino/issues/254.

PR #45, opened 2025-03-31
Set up key material storage

### LINKED ISSUE #18: Key material storage
In `zcashd`, spending key material (and the HD seed it was derived from where relevant) were stored in the same BDB database as the rest of the wallet data.

`zcash_client_sqlite` (which Zallet uses as its backend) is designed to never store spending key material, due to its history as the backing library for the mobile SDKs. On mobile devices, platform-level keystores exist which are ideal for storing seed phrases or key material. However, Zallet is a desktop wallet, and therefore cannot assume the presence of these kinds of affordances.

We therefore need to decide where and how to store key material in Zallet.

str4d:
Closes #18.


Cargo.toml
```diff
@@ -80,3 +86,11 @@ tonic = "0.12"
 [patch.crates-io]
 abscissa_core = { git = "https://github.com/str4d/abscissa.git", rev = "8da8e10cad02aa4a93099cb4942b48fbdc54eb89" }
 abscissa_tokio = { git = "https://github.com/str4d/abscissa.git", rev = "8da8e10cad02aa4a93099cb4942b48fbdc54eb89" }
+age = { git = "https://github.com/str4d/rage.git", rev = "84dc1e9f641994388f107ceebdbaa00126e30e16" }
```
str4d, line 90: This is https://github.com/str4d/rage/pull/565.

zallet/src/components/keystore/db/migrations/initial_setup.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub(super) struct Migration;
+
+impl schemerz::Migration<Uuid> for Migration {
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        V_0_15_0.iter().copied().collect()
```
str4d, line 17: This should be `V_0_16_0`, but we keep forgetting to create these constants during the `zcash_client_sqlite` release process.

Cargo.toml
daira: Reviewed and checked the hash.

deny.toml
```diff
@@ -20,6 +20,7 @@ allow = [
 ]
 exceptions = [
     { name = "arrayref", allow = ["BSD-2-Clause"] },
+    { name = "curve25519-dalek", allow = ["BSD-3-Clause"] },
```
daira, line 23: Is there a reason why we don't just put {BSD-2-Clause, BSD-3-Clause, ISC, CC0-1.0} in `allow`? Those are permissive licenses that I think are universally accepted to be compatible with Apache-2.0 and MIT.

supply-chain/config.toml
```diff
@@ -255,14 +271,18 @@ criteria = "safe-to-deploy"
...
-version = "0.2.3"
-criteria = "safe-to-deploy"
-
 [[exemptions.crypto-common]]
 version = "0.2.0-rc.1"
 criteria = "safe-to-deploy"
 
+[[exemptions.curve25519-dalek]]
+version = "4.1.0"
+criteria = "safe-to-deploy"
```
daira, line 280: There are a lot of exemptions for cryptographic dependencies being added here. Should we try to review some of them?

zallet/src/components/keystore.rs
```diff
@@ -0,0 +1,568 @@
...
+        }
+
+        *self.identities.write().await = decrypted_identities;
+
+        // Start a task to relock the keystore after the given timeout.
+        let duration = Duration::from_secs(timeout);
+        let identities = self.identities.clone();
+        *relock_task = Some((
+            SystemTime::now() + duration,
+            tokio::spawn(async move {
```
daira, line 318: This doesn't lock the mutex when it starts writing. So how do we know a write isn't in progress when we lock `self.relock_task` in order to change the timeout (in a concurrent call to `unlock`, say)?

daira (dismissed):
I am not sure about the correctness of the locking on writing identities.

zallet/src/components/keystore.rs
str4d: I don't believe there is a problem here; @daira does. I'm going to avoid the problem by awaiting `existing_timeout` after aborting it, which ensures that the write below can never be in play.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/d21449a1a7d55298616c9c356dcf44b92de909ac..c41b4aeff911ab540d73c0ac9efb83c0eec0d629) to address review comment.

deny.toml
str4d: I'm taking the same approach here that we do in `librustzcash` and (effectively) `zcashd`. If we want to take a different approach, we should do so in separate PRs everywhere.

supply-chain/config.toml
str4d: Maybe at some point, but currently we don't have time. My general approach for these is to exempt the first time a dependency lands, and then attempt to review diffs as they occur if they are small enough.

zallet/src/config.rs
```diff
@@ -138,6 +142,15 @@ impl BuilderSection {
     }
 }
 
+/// Key store configuration section.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct KeyStoreSection {
+    /// Path to the age identity file that encrypts key material.
+    // TODO: Change this to `PathBuf` once `age::IdentityFile::from_file` supports it.
+    pub identity: String,
```
nuttycom, line 151: Why not simply convert from `PathBuf` to `String` at the use site, and have this correct from the outset?

Also do config fields have to be public?

nuttycom (changes requested):
utACK apart from one blocking question/possible bug

zallet/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -40,15 +44,28 @@ pub(crate) struct GetWalletInfo {
 
     /// The timestamp in seconds since epoch (midnight Jan 1 1970 GMT) that the wallet is
     /// unlocked for transfers, or 0 if the wallet is locked.
```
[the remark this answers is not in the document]
str4d: `None` (i.e. field omitted) does not mean the wallet is locked, it means the wallet is not using passphrase encryption (i.e. the identities for decrypting key material are always available in memory). The existing documentation is correct.

zallet/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -40,15 +44,28 @@ pub(crate) struct GetWalletInfo {
...
 
     /// The BLAKE2b-256 hash of the HD seed derived from the wallet's emergency recovery phrase.
     mnemonic_seedfp: String,
 }
 
-pub(crate) fn call() -> Response {
+pub(crate) async fn call(keystore: &KeyStore) -> Response {
     warn!("TODO: Implement getwalletinfo");
 
+    let unlocked_until = if keystore.is_crypted() {
```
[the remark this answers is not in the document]
str4d: No. Compare this to https://github.com/zcash/zcash/blob/a3435336b0c561799ac6805a27993eca3f9656df/src/wallet/rpcwallet.cpp#L2573-L2574.

zallet/src/config.rs
str4d: Conversion from `PathBuf` to `String` is potentially lossy, as paths may contain invalid UTF-8 on some systems (case in point: Windows).

Abscissa structures its apps as "library + binary that uses it", and AFAICT it usually makes config fields public in the library part so they can be accessed programmatically in tests, but I've not been fully consistent here. We will probably want to do a pass over the entire codebase at some point to normalise / decide what we want to expose in the library view.

zallet/src/components/json_rpc/methods/get_wallet_info.rs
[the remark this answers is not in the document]
nuttycom: So, I also don't understand that code, or else I don't understand what `is_crypted` means. I would think that `is_crypted` would return `false` if the secret material is available in memory, i.e. if the wallet is unlocked. If the wallet is locked (if the secret material is encrypted) then it doesn't make sense to return the `unlocked_until` field. 

If my understanding of `is_crypted` is incorrect, please rename it because it's exceedingly confusing and we shouldn't reproduce `zcashd`'s weirdness.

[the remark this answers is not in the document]
str4d: > I would think that is_crypted would return false if the secret material is available in memory, i.e. if the wallet is unlocked.

No, that's `is_locked` which is also present. `is_crypted` in `zcashd` meant "is the wallet encrypted at all", which was relevant because when it wasn't encrypted (which was the default in `zcashd`, as encryption was experimental) then all spending key material was available at any time. When the wallet is encrypted, the spending key material is only available if the wallet is also unlocked.

In Zallet, I'm requiring that all spending key material is always encrypted, but the `is_crypted` distinction is still relevant because if a non-encrypted identity is used, it means the spending key material is functionally available at any time, whereas if an encrypted identity is used, the spending key material is only available if the wallet is also unlocked.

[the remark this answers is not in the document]
str4d: Would it be clearer if I rename `is_crypted` to `is_lockable`?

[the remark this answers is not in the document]
nuttycom: Maybe rename to `uses_encrypted_identities` then? I would prefer however that this method have its semantics changed such that it could use `!keystore.is_locked()`. Given that wallet encryption was experimental in zcashd, I think that this semantic change would be okay.

nuttycom (approved):
utACK e82f2cc4a3334d69d58f2311f424c2f7c15a9bd5

nuttycom:
Clippy failures are just due to methods that are currently unused, merging.

daira (commented):
Post-hoc ACK

Issue #95, opened 2025-04-03
rpc: Add support for async operations


str4d:
Some `zcashd` RPCs kick off a process internally and return an operation ID that can be queried later, to avoid running into JSON-RPC timeouts in client libraries for potentially-long operations like `z_sendmany`. To maintain the same API we need similar logic (which will be simpler to implement as we already have Tokio doing most of the work).

PR #102, opened 2025-04-09
rpc: Add support for async operations

str4d:
Closes zcash/wallet#77.
Closes zcash/wallet#78.
Closes zcash/wallet#83.
Closes zcash/wallet#95.

### DIFF
--- a/zallet/src/components/json_rpc/asyncop.rs
+++ b/zallet/src/components/json_rpc/asyncop.rs
@@ -1,0 +1,188 @@
+use std::future::Future;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use jsonrpsee::{core::RpcResult, types::ErrorObjectOwned};
+use serde::Serialize;
+use serde_json::Value;
+use tokio::sync::RwLock;
+use uuid::Uuid;
+
+/// The possible states that an async operation can be in.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize)]
+#[serde(into = "String")]
+pub(super) enum OperationState {
+    Ready,
+    Executing,
+    Cancelled,
+    Failed,
+    Success,
+}
+
+impl FromStr for OperationState {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "queued" => Ok(Self::Ready),
+            "executing" => Ok(Self::Executing),
+            "cancelled" => Ok(Self::Cancelled),
+            "failed" => Ok(Self::Failed),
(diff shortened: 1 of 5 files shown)


zallet/src/components/json_rpc/methods/get_operation.rs
```diff
@@ -0,0 +1,52 @@
...
+
+pub(crate) async fn status(async_ops: &[AsyncOperation], operationid: Vec<&str>) -> Response {
+    let filter = operationid.into_iter().collect::<HashSet<_>>();
+
+    let mut ret = vec![];
+
+    for op in async_ops {
+        if !filter.is_empty() && !filter.contains(op.operation_id()) {
+            continue;
+        }
```
nuttycom, line 18: Why not:
```suggestion
    for op in async_ops.iter().filter(|op| filter.is_empty() || filter.contains(op.operation_id())) {
```

zallet/src/components/json_rpc/methods/get_operation.rs
```diff
@@ -0,0 +1,52 @@
...
+) -> Response {
+    let filter = operationid.into_iter().collect::<HashSet<_>>();
+
+    let mut ret = vec![];
+    let mut remove = HashSet::new();
+
+    for op in async_ops.iter() {
+        if !filter.is_empty() && !filter.contains(op.operation_id()) {
+            continue;
+        }
```
nuttycom, line 38:
```suggestion
    for op in async_ops.iter().filter(|op| filter.is_empty() || filter.contains(op.operation_id()) {
```

nuttycom (approved):
utACK 4ebe830f8c4f8a3ae4f9615b53ffe27a602c2390 with nonblocking style suggestions

zallet/src/components/json_rpc/methods/get_operation.rs
nuttycom: (I find this style easier to reason about than thinking through the negations.)

zallet/src/components/json_rpc/asyncop.rs
```diff
@@ -0,0 +1,188 @@
...
+            "executing" => Ok(Self::Executing),
+            "cancelled" => Ok(Self::Cancelled),
+            "failed" => Ok(Self::Failed),
+            "success" => Ok(Self::Success),
+            _ => Err("invalid operation status"),
+        }
+    }
+}
+
+impl From<OperationState> for String {
```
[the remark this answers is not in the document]
str4d: I thought this was a `serde` constraint in that it needed an owned type for the `#[serde(into)]` annotation, but I just tried and indeed `#[serde(into = "&'static str")]` works fine, so I can use that as the intermediate instead.

PR #103, opened 2025-04-14
rpc: Add `z_sendmany`

str4d:


zallet/src/components/json_rpc/payments.rs
```diff
@@ -0,0 +1,446 @@
...
+        self.is_compatible_with(PrivacyPolicy::AllowRevealedSenders)
+    }
+
+    pub(super) fn allow_fully_transparent(&self) -> bool {
+        self.is_compatible_with(PrivacyPolicy::AllowFullyTransparent)
+    }
+
+    pub(super) fn allow_linking_account_addresses(&self) -> bool {
+        self.is_compatible_with(PrivacyPolicy::AllowLinkingAccountAddresses)
+    }
```
str4d, line 191: These helpers are present in `zcashd` but are currently unused here. We need to figure out what `zcashd` used them for and perform equivalent checks here, or remove them if truly unnecessary.

zallet/src/components/json_rpc/payments.rs
```diff
@@ -0,0 +1,446 @@
...
+    // No step revealed anything.
+    PrivacyPolicy::FullPrivacy
+}
+
+pub(super) enum AddressResolutionError {
+    /// Requested [`PrivacyPolicy`] doesn’t include `AllowRevealedRecipients`.
+    TransparentRecipientNotAllowed,
+
+    /// Requested [`PrivacyPolicy`] doesn’t include `AllowRevealedRecipients`.
+    TransparentChangeNotAllowed,
```
str4d, line 265: This is similarly a `zcashd` case we don't yet handle here.

zallet/src/components/json_rpc/payments.rs
```diff
@@ -0,0 +1,446 @@
...
+
+    /// Requested [`PrivacyPolicy`] doesn’t include `AllowRevealedAmounts`, but we don’t
+    /// have enough Sapling funds to avoid revealing amounts.
+    RevealingSaplingAmountNotAllowed,
+
+    /// Despite a lax [`PrivacyPolicy`], other factors made it impossible to find a
+    /// receiver for a recipient UA.
+    CouldNotResolveReceiver,
```
str4d, line 273: This is similarly a `zcashd` case we don't yet handle here.

str4d:
[Rebased on `main`](github.com/zcash/wallet/compare/2b72c31c1d0db07244ab73f36803a1f758bcba7b..d1e6a37caa2d967f7a8e57f7366fc5cd57174005).

zallet/src/components/json_rpc/payments.rs
str4d: Now handled.

zallet/src/components/json_rpc/payments.rs
str4d: Now handled.

zallet/src/components/json_rpc/payments.rs
str4d: In `zcashd` this case's error message indicated what it meant was "Orchard output with insufficient Orchard funds to avoid revealing amounts". Handled now.

nuttycom (approved):
utACK 4aa52b4fa3f4763c497af9fc6f979647a8b86782

zallet/src/components/json_rpc/utils.rs
```diff
@@ -92,3 +93,318 @@ pub(super) fn parse_diversifier_index(diversifier_index: u128) -> RpcResult<Dive
...
+    let mut point_ofs = 0;
+
+    let mantissa_sign = match val.split_at_checked(1) {
+        Some(("-", rest)) => {
+            val = rest;
+            true
+        }
+        _ => false,
+    };
+    match val.split_at_checked(1) {
```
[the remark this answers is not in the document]
str4d: I started with an exact copy with a hunch of panicking behaviour, and then modified it to what is here now. In hindsight splitting that into 2 commits could have helped.

zallet/src/components/json_rpc/methods.rs
```diff
@@ -226,6 +227,72 @@ pub(crate) trait Rpc {
...
+
+    /// Send a transaction with multiple recipients.
+    ///
+    /// This is an async operation; it returns an operation ID string that you can pass to
+    /// `z_getoperationstatus` or `z_getoperationresult`.
+    ///
+    /// Amounts are decimal numbers with at most 8 digits of precision.
+    ///
+    /// Change generated from one or more transparent addresses flows to a new transparent
+    /// address, while change generated from a legacy Sapling address returns to itself.
```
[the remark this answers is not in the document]
str4d: Correct that this documentation does not yet match the impl; it matches the `zcashd` documentation. If we want to change this (by way of not making changes to the backend) then we need to document this difference as it is very significant. Note also that this entire half of the paragraph is linked to the legacy account; UAs (every other account) behave sensibly.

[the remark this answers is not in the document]
nuttycom: I propose that legacy Sapling addresses be treated by `zallet` as "spend-only", such that funds from them may be spent but that change must go to a mnemonic-derived HD account. We could add a new argument for specifying the account to which change should be sent that is treated as required in the case that the "from" address is a legacy Sapling address.

str4d:
[Rebased on `main`](https://github.com/zcash/wallet/compare/4aa52b4fa3f4763c497af9fc6f979647a8b86782..7a6cb4db18bacda13f42a28e5accaf8323642a5b) to fix merge conflict.

nuttycom (approved):
re-utACK https://github.com/zcash/wallet/compare/4aa52b4..ab72580

Issue #104, opened 2025-04-15
Please add a few lines to get started in the README


oxarbitrage:
No command works, except for `migrate-zcashd-conf` that i did not tried. All commands fail with `Failed to initialize zallet: wallet_db must be set (for now)`

Please add a few lines in the readme to start a new wallet. 

```
alfredo@spaceship:~/zallet/wallet$ cargo run --release start
warning: associated function `new` is never used
  --> zallet/src/components/json_rpc/asyncop.rs:64:25
   |
62 | impl AsyncOperation {
   | ------------------- associated function in this implementation
63 |     /// Launches a new async operation.
64 |     pub(super) async fn new<T: Serialize + Send + 'static>(
   |                         ^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: method `start_async` is never used
   --> zallet/src/components/json_rpc/methods.rs:219:14
    |
193 | impl RpcImpl {
    | ------------ method in this implementation
...
219 |     async fn start_async<T: Serialize + Send + 'static>(
    |              ^^^^^^^^^^^

warning: `zallet` (lib) generated 2 warnings
    Finished `release` profile [optimized] target(s) in 0.35s
     Running `target/release/zallet start`
Failed to initialize zallet: wallet_db must be set (for now)

[ Did zallet not do what you expected? Could the error be more useful? ]
[ Tell us: https://github.com/zcash/wallet/issues                      ]
alfredo@spaceship:~/zallet/wallet$ cargo run --release init-wallet-encryption
warning: associated function `new` is never used
  --> zallet/src/components/json_rpc/asyncop.rs:64:25
   |
62 | impl AsyncOperation {
   | ------------------- associated function in this implementation
63 |     /// Launches a new async operation.
64 |     pub(super) async fn new<T: Serialize + Send + 'static>(
   |                         ^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: method `start_async` is never used
   --> zallet/src/components/json_rpc/methods.rs:219:14
    |
193 | impl RpcImpl {
    | ------------ method in this implementation
...
219 |     async fn start_async<T: Serialize + Send + 'static>(
    |              ^^^^^^^^^^^

warning: `zallet` (lib) generated 2 warnings
    Finished `release` profile [optimized] target(s) in 0.32s
     Running `target/release/zallet init-wallet-encryption`
Failed to initialize zallet: wallet_db must be set (for now)

[ Did zallet not do what you expected? Could the error be more useful? ]
[...]

PR #106, opened 2025-04-17
Document and test the current workflow for setting up a new wallet

str4d:
Closes zcash/wallet#104.


oxarbitrage:
I am getting an error in the last step, previous steps were all sucesful.

```
$ zallet -c zallet.toml start
2025-04-17T23:44:00.172000Z  INFO zallet::components::database: Applying latest database migrations
2025-04-17T23:44:00.175356Z  INFO schemerz: Migrating everything    
2025-04-17T23:44:00.175988Z  INFO zallet::components::chain_view: Starting Zaino indexer
2025-04-17T23:44:00.176014Z  INFO zaino_state::fetch: Launching Chain Fetch Service..
2025-04-17T23:44:09.183559Z ERROR zaino_fetch::jsonrpsee::connector: Error: Could not establish connection with node. Please check config and confirm node is listening at the correct address and the correct authorisation details have been entered. Exiting..
$ 
```

Do we need zebra or zaino running ?

str4d:
> Do we need zebra or zaino running ?

Yes you need Zebra running. That's the `validator_*` settings in the `[indexer]` section. If the Zebra node is not on localhost at the default port for that network, you'll need to set `validator_address` in that section.

No you do not need `zainod` running; Zallet currently bundles Zaino internally.

str4d:
... nope, the error is still there, and I just figured it out:
```
 error: zallet fatal error: parse error: TOML parse error at line 2, column 18
  |
2 | wallet_db = "C:\Users\RUNNER~1\AppData\Local\Temp\.tmpLFS8VU\data.sqlite"
  |                  ^
invalid unicode 8-digit hex code
```

The `toml` crate is parsing the `\U` inside the path string as a Unicode escape sequence, which is [part of the TOML spec](https://github.com/toml-lang/toml/blob/78fcf9dd7eab7acfbaf147c684b649477e7bdd9c/toml.abnf#L88-L99). What I don't know is whether this is a bug in `serde`'s `PathBuf` serializer, or in `toml`'s deserializer, or if this is an edge case in the `serde` deserialization pathway that we need to handle manually.

README.md
```diff
@@ -6,6 +6,90 @@ Work in progress replacement of the `zcashd` wallet.
...
+
+```
+$ zallet -c /path/to/zallet/datadir/zallet.toml generate-mnemonic
+```
+
+Each time you run this, a new [BIP 39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)
+mnemonic will be added to the wallet. Be careful to only run it multiple times
+if you want multiple independent roots of spend authority!
+
+### Start Zallet
```
oxarbitrage, line 87:
```suggestion
### Start Zebra

To match the `zallet.toml` example above, Zebra must be running on its default RPC endpoint with no authentication (alternatively, you can set the `validator_*` variables in your `zallet.toml`). In your Zebra config (e.g. `~/.zebra/config.toml`), add or update:

[rpc]
enable_cookie_auth = false
listen_addr       = "127.0.0.1:8232"

### Start Zallet
```

zallet/tests/acceptance.rs
```diff
@@ -40,3 +44,69 @@ fn version_no_args() {
     let mut cmd = runner.arg("--version").capture_stdout().run();
     cmd.stdout().expect_regex(r"\A\w+ [\d\.\-]+\z");
 }
+
+#[test]
```
oxarbitrage, line 48:
```suggestion
#[test]
#[cfg(not(target_os = "windows"))]
```

README.md
str4d: "with no authentication" is false, that's what the `validator_user` and `validator_password` lines are. If you need cookie auth, you have `validator_cookie_auth` and `validator_cookie_path` settings.

oxarbitrage: You need to turn that off anyway in Zebra if you want the `zallet.toml` example to work.

str4d: Added a paragraph to the config section where it is relevant.

str4d: I'm not including Zebra config details in the Zallet README, in part because currently Zallet does work with `zcashd`, and partly because it is up to the user to configure their node appropriately and I don't want to keep updating the README if external changes make it stale. When we get closer to production and have dedicated guides for setup that are more stable, we can consider providing guides for specific full node versions.

oxarbitrage (approved):
ACK https://github.com/zcash/wallet/pull/106/commits/c03901b0796b0dcd2cae2e88ed7c4a4f55a205b2

oxarbitrage:
@str4d Can you give me write access on this repo ?

PR #107, opened 2025-04-17
Add support for storing pre-BIP39 zcashd HD seed data.

nuttycom:


zallet/src/components/keystore/db/migrations/initial_setup.rs
```diff
@@ -32,7 +32,7 @@ impl RusqliteMigration for Migration {
                 added TEXT NOT NULL
             );
             CREATE TABLE ext_zallet_keystore_mnemonics (
-                hd_seed_fingerprint BLOB NOT NULL,
+                hd_seed_fingerprint BLOB NOT NULL UNIQUE,
```
str4d, line 35: Second commit needs to update `components/keystore/db.rs`

zallet/src/components/keystore/db/migrations/initial_setup.rs
```diff
@@ -34,6 +34,10 @@ impl RusqliteMigration for Migration {
             CREATE TABLE ext_zallet_keystore_mnemonics (
                 hd_seed_fingerprint BLOB NOT NULL UNIQUE,
                 encrypted_mnemonic BLOB NOT NULL
+            );
+            CREATE TABLE ext_zallet_keystore_legacy_seeds (
```
str4d, line 38: Third commit also needs to do so.

zallet/i18n/en-US/zallet.ftl
```diff
@@ -84,6 +86,12 @@ err-migrate-multiple-related-zcashd-options =
     Remove one of the conflicting options, then re-run this command.
 err-migrate-unknown-zcashd-option = Unknown {-zcashd} option '{$option}'
 
+# errors in migration of wallet data from the zcashd `wallet.dat` database format
+
+err-failed-seed-fingerprinting =
```
str4d, line 91: I generally try to pick IDs that are namespaced:
```suggestion
err-migrate-wallet-invalid-seed =
```

str4d (changes requested):
Reviewed 933ee3c23692378cc2997baf426fc6aa69b0ba5b

zallet/i18n/en-US/zallet.ftl
nuttycom: In this case, while the error will often be generated from migration, the code that it lives in isn't strictly speaking migration code, so I didn't add that prefix as I have for other migration-related error messages.

nuttycom:
[force-pushed](https://github.com/zcash/wallet/compare/933ee3c..51177fad367dc6e8a3a522a02fc2103ef026b943)

zallet/src/components/keystore/db.rs
```diff
@@ -52,3 +52,21 @@ CREATE TABLE ext_zallet_keystore_mnemonics (
...
 )
 "#;
+
+/// Stores encrypted raw HD seeds.
+///
+/// ### Columns
+///
+/// - `hd_seed_fingerprint` is the [ZIP 32 fingerprint] for the raw seed bytes.
+///   This is present to enable quick lookups of which seed needs to be decrypted
+///   at spend time (rather than trial-decrypting every mnemonic).
```
str4d, line 62:
```suggest
///   at spend time (rather than trial-decrypting every legacy seed).
```

str4d (approved):
utACK 51177fad367dc6e8a3a522a02fc2103ef026b943

PR #109, opened 2025-04-18
Fix clippy lints

oxarbitrage:
To prevent introducing more Clippy warnings, we can get the Clippy workflow passing. If that’s not the best approach, I can alternatively remove the unused functions or tackle the warnings in another way, let me know.

### DIFF
--- a/zallet/src/components/json_rpc/asyncop.rs
+++ b/zallet/src/components/json_rpc/asyncop.rs
@@ -61,6 +61,7 @@ pub(super) struct AsyncOperation {
 
 impl AsyncOperation {
     /// Launches a new async operation.
+    #[allow(unused)]
     pub(super) async fn new<T: Serialize + Send + 'static>(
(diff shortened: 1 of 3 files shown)


str4d:
NACK. The clippy workflow will pass once #103 is merged, which uses these methods.

oxarbitrage:
Generally, I prefer to add #[allow(unused)] to new unused functions as they’re introduced, rather than let Clippy failures block the entire repo on future PRs that use the functions.

oxarbitrage:
I noticed that in [PR #103](https://github.com/zcash/wallet/pull/103) some clippy warnings were fixed, but a few new ones appeared. Right now PRs can introduce new lints ignoring failing CI and get merged, leaving the cleanup to future PRs. Could we clear all existing warnings now and require any new ones to be accompanied by `#[allow(...)]` and `// TODO` comments in the PR that introduces them?

PR #112, opened 2025-04-18
Enable importing existing accounts

str4d:


zallet/src/components/json_rpc/methods/recover_accounts.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        let account_index =
+            zip32::AccountId::try_from(account.zip32_account_index).map_err(|e| {
+                LegacyCode::InvalidParameter
+                    .with_message(format!("Invalid ZIP 32 account index: {e}"))
+            })?;
+
+        let birthday_height = BlockHeight::from_u32(account.birthday_height);
+
+        let treestate = {
```
nuttycom, line 82: This code now exists in a few different places in the node, we should factor it out.

nuttycom (approved):
utACK 4d90d412a1830aed8e6be67216d6815486b38c6b modulo clippy failures

Issue #115, opened 2025-04-22
Can't initialize wallet in regtest mode


oxarbitrage:
Zallet.conf:

```
network = "regtest"
wallet_db = "/home/alfredo/zallet/regtest-test/2/datadir/data.sqlite"

[builder]

[indexer]
validator_user = ".."
validator_password = ".."
db_path = "/home/alfredo/zallet/regtest-test/2/datadir/zaino"

[keystore]
identity = "/home/alfredo/zallet/regtest-test/2/datadir/identity.txt"

[limits]

[rpc]
bind = ["127.0.0.1:9999"]
```

Init wallet:

```
     Running `target/debug/zallet -c ../datadir/zallet.toml init-wallet-encryption`
2025-04-22T13:58:14.178789Z  INFO zallet::components::database: Creating empty database
2025-04-22T13:58:14.701050Z  INFO schemerz: Migrating everything    
2025-04-22T13:58:14.702158Z  INFO schemerz: Applying migration bc4f5e57-d600-4b6c-990f-b3538f0bfce1    
2025-04-22T13:58:15.269536Z  INFO schemerz: Applying migration be57ef3b-388e-42ea-97e2-678dafcf9754    
2025-04-22T13:58:15.864488Z  INFO schemerz: Applying migration 0ddbe561-8259-4212-9ab7-66fdc4a74e1d    
2025-04-22T13:58:16.411318Z  INFO schemerz: Applying migration a2e0ed2e-8852-475e-b0a4-f154b15b9dbe    
2025-04-22T13:58:16.961725Z  INFO schemerz: Applying migration d956978c-9c87-4d6e-815d-fb8f088d094c    
2025-04-22T13:58:17.512325Z  INFO schemerz: Applying migration 761884d6-30d8-44ef-b204-0b82551c4ca1    
2025-04-22T13:58:17.877954Z  INFO schemerz: Applying migration 282fad2e-8372-4ca0-8bed-71821320909f    
2025-04-22T13:58:18.171703Z  INFO schemerz: Applying migration 2aa4d24f-51aa-4a4c-8d9b-e5b8a762865f    
2025-04-22T13:58:18.457059Z  INFO schemerz: Applying migration bdcdcedc-7b29-4f1c-8307-35f937f0d32a    
2025-04-22T13:58:18.765753Z  INFO schemerz: Applying migration 7029b904-6557-4aa1-9da5-6904b65d2ba5    
[...]

PR #118, opened 2025-04-22
Allow `nuparams` to be added to the config

oxarbitrage:
I am unsure if TOML support `&str` values directly, so this PR switches to using `String`. With that in place, you can now write, for example:

```
regtest_nuparams = [
  "5ba81b19:1",
  "76b809bb:10",
  "2bb40e60:15",
  "f5b9230b:20",
  "e9ff75a6:25",
  "c2d6d0b4:30",
  "c8e71055:35",
]
```

Closes https://github.com/zcash/wallet/issues/115

### DIFF
--- a/zallet/src/commands/migrate_zcash_conf.rs
+++ b/zallet/src/commands/migrate_zcash_conf.rs
@@ -507,7 +507,7 @@ fn build_actions() -> HashMap<&'static str, Action> {
         .chain(Action::map_multi(
             "nuparams",
             |config| &mut config.regtest_nuparams,
-            |value| RegTestNuParam::try_from(value).map_err(|_| ()),
+            |value| RegTestNuParam::try_from(value.to_string()).map_err(|_| ()),
         ))
         .chain(Action::map_related(
             "regtest",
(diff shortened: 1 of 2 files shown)


str4d:
> I am unsure if TOML support `&str` values directly,

Did you test it without this change? And if so, what error did you get?

The code was compiling fine, so the Serde side is supported. Meanwhile the `&str` is only an intermediary, we don't store it anywhere.

oxarbitrage:
This is not a compilation issue, is when you try to add `regtest_nuparams` to zallet.toml.

zallet/src/commands/migrate_zcash_conf.rs
```diff
@@ -507,7 +507,7 @@ fn build_actions() -> HashMap<&'static str, Action> {
         .chain(Action::map_multi(
             "nuparams",
             |config| &mut config.regtest_nuparams,
-            |value| RegTestNuParam::try_from(value).map_err(|_| ()),
+            |value| RegTestNuParam::try_from(value.to_string()).map_err(|_| ()),
```
str4d, line 510: The reason I was confused is that the code altered here was working perfectly. I've just written additional testing and reconfirmed that the existing code works fine for serializing a `ZalletConfig`. The fact that serialization works fine but deserialization apparently fails at runtime is an annoying serde aspect.

PR #122, opened 2025-04-24
Add a `zallet example-config` command

str4d:


zallet/src/config.rs
```diff
@@ -233,3 +236,69 @@ impl RpcSection {
...
+        }
+
+        fn write_field<T: DocumentedFields>(config: &mut String, field_name: &str) {
+            let field_doc = T::get_field_docs(field_name).expect("present");
+            for line in field_doc.lines() {
+                writeln!(config, "# {}", line).unwrap();
+            }
+
+            // TODO: Get the default value for the field.
+            let field_default = toml::Value::try_from(vec![0u64]).unwrap();
```
str4d, line 283: Currently we hard-code defaults inside getter methods, e.g. here:
https://github.com/zcash/wallet/blob/449eab1693b543bfcb9c1c4fac02eb866793cb38/zallet/src/config.rs#L130-L137

We instead need a way to map field names to default values, that we can share somehow between here and those getters.

str4d:
It occurs to me that once #120 is merged, we won't need to alter `build.rs` to generate the current example config for inclusion in Debian packages. We can instead have a CLI test that pins its output (and that we update whenever a change occurs), and then just copy that output file.

zallet/src/config.rs
str4d: For now I've gone with a minimally-duplicating approach; we'll need to remember to update a list at the bottom of the file whenever we alter the structure of the config. But it's pretty short and should be manageable for now, and can be turned into a proc macro later if we want better UX.

zallet/src/commands/example_config.rs
```diff
@@ -0,0 +1,62 @@
...
+        if !self.this_is_alpha_code_and_you_will_need_to_recreate_the_example_later {
+            return Err(ErrorKind::Generic.context(fl!("migrate-alpha-code")).into());
+        }
+
+        // Serialize the example config.
+        let output = ZalletConfig::generate_example();
+
+        // Write the Zallet config file.
+        let output_path = match self.output.as_deref() {
+            None => todo!("No default Zallet config path yet, use -o/--output"),
```
str4d, line 25: This will be fixed by #123.

zallet/src/commands/example_config.rs
nuttycom, line 1: Incorrect error message: 

```
λ cargo run -- example-config
     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/zallet example-config`
Error: This command is not stable, and parts of your zcashd config may not get
migrated correctly. You will need to rerun this command again once zallet
is stable to migrate your config correctly. To confirm you are aware of
this, use '--this-is-alpha-code-and-you-will-need-to-redo-the-migration-later'.

[ Tell us: https://github.com/zcash/wallet/issues                      ]
```

```
~/work/zallet/canon on ᚠ22-generate-config [$] via 🦀 v1.85.1
✗ cargo run -- example-config --this-is-alpha-code-and-you-will-need-to-redo-the-migration-later
     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/zallet example-config --this-is-alpha-code-and-you-will-need-to-redo-the-migration-later`
error: unexpected argument '--this-is-alpha-code-and-you-will-need-to-redo-the-migration-later' found

  tip: a similar argument exists: '--this-is-alpha-code-and-you-will-need-to-recreate-the-example-later'

Usage: zallet example-config --this-is-alpha-code-and-you-will-need-to-recreate-the-example-later

For more information, try '--help'.
```

nuttycom (changes requested):
Tested, not working. Looks like the help message for requiring the "you will have to redo this" flag is incorrect.

nuttycom:
Also, a question: the `keystore` `identity` entry is commented out, but I believe that that is a required field? At present, there is nothing uncommented other than the section heads.

str4d:
[Rebased on `main`](https://github.com/zcash/wallet/compare/6773c47f208add0df7d55a5df612282acecbc62f..3ca9106f26960c3bf02e4754a3d3b8ca65eb8d7a) and addressed comments.

zallet/src/config.rs
```diff
@@ -233,3 +237,179 @@ impl RpcSection {
...
+        };
+
+        let mut config = r"# Default configuration for Zallet.
+#
+# This file is generated as an example using Zallet's current defaults. It can
+# be used as a skeleton for custom configs.
+#
+# Fields that are required to be set are uncommented, and set to an example
+# value. Every other field is commented out, and set to the current default
+# value that Zallet will use for it (or `UNSET` if the field has no default).
```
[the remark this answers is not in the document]
str4d: We now set `network` and will eventually make the rest properly optional.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/3ca9106f26960c3bf02e4754a3d3b8ca65eb8d7a..e927a05d0b5146c8bca86a602db90f7102fde17e) to fix clippy lints.

nuttycom (approved):
re-ACK

Issue #116, opened 2025-04-22
rpc: Add a `help` method


str4d:
`zcashd` has a `help (METHOD_NAME)` method for introspecting API functionality. It would be nice to have the same in Zallet.

One downside to `zcashd`'s implementation is that it was done as an multiline string returned at the start of the method via an error, which was hard to maintain (and often led to the documentation getting outdated).

In Zallet we've been documenting the RPCs with Rust docstrings. It would be very convenient if we could access these via macros in order to create the `help` method responses.

str4d:
If https://github.com/paritytech/jsonrpsee/issues/737 were finished then we could instead have OpenRPC support for API introspection. That being said, maybe manually implementing that would be preferable to replicating `zcash-cli help`? Although the latter is useful for familiarity.

oxarbitrage:
Documenting every Zcash RPC endpoint with a standard OpenAPI/Swagger spec would be beneficial for users and integrators.

Back in a hackathon I put together a very basic, fragile prototype for Zebra:

https://zebra.zfnd.org/user/openapi.html

The core idea is similar what https://github.com/zcash/wallet/pull/125 does: extract each RPC method’s doc-comments and use them to build an OpenAPI spec. In practice though, we needed:

- Structured comments: Each RPC handler must use a prescribed comment format so the generator can recognize parameters, request bodies, responses, etc.
- A custom generator tool (openapi-generator) parses those comments and builds the YAML.
    - https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/README.md#openapi-generator
    - https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/src/bin/openapi-generator/main.rs
- Manual tweaks: A few things still require hand-editing ( https://github.com/ZcashFoundation/zebra/blob/main/zebra-utils/src/bin/openapi-generator/main.rs#L477), especially to fill in details that can’t be inferred purely from comments, i think this can be improved.

I’m not suggesting we do this now, https://github.com/zcash/wallet/pull/125 approach is already a great start. But if we can refine the generator so it’s robust and mostly hands-off, we’ll automatically get interactive, up-to-date API docs from the comments we’re already writing.

str4d:
OpenAPI is for REST-ful HTTP APIs; it is the wrong thing to use for JSON-RPC. Instead we'd use [OpenRPC](https://open-rpc.org/) which is designed for JSON-RPC 2.0. Opened #127 for this.

[...]

PR #125, opened 2025-04-29
rpc: Add `help` command

str4d:
This is compatible with `zcash-cli help` in that it can be called the same way and returns a string to be printed. However, instead of the ad-hoc format used in `zcashd`, it returns the documentation from the internal Rust method (which is easy for us to maintain).

Closes zcash/wallet#116.


oxarbitrage:
```
$ curl --silent -X POST -H "Content-Type: application/json" -d '{"jsonrpc": "2.0", "method": "help", "params": [], "id":123 }' 127.0.0.1:9999 | jq
{
  "jsonrpc": "2.0",
  "result": "getwalletinfo\nhelp\nlistaddresses\nwalletlock\nwalletpassphrase\nz_getaddressforaccount\nz_getnewaccount\nz_getnotescount\nz_getoperationresult\nz_getoperationstatus\nz_listaccounts\nz_listoperationids\nz_listunifiedreceivers\nz_listunspent\nz_recoveraccounts\n",
  "id": 123
}
$ 

$ curl --silent -X POST -H "Content-Type: application/json" -d '{"jsonrpc": "2.0", "method": "help", "params": ["listaddresses"], "id":123 }' 127.0.0.1:9999 | jq
{
  "jsonrpc": "2.0",
  "result": "listaddresses\n\nLists the addresses managed by this wallet by source.\n\nSources include:\n- Addresses generated from randomness by a legacy `zcashd` wallet.\n- Sapling addresses generated from the legacy `zcashd` HD seed.\n- Imported watchonly transparent addresses.\n- Shielded addresses tracked using imported viewing keys.\n- Addresses derived from mnemonic seed phrases.\n\nIn the case that a source does not have addresses for a value pool, the key\nassociated with that pool will be absent.\n\nREMINDER: It is recommended that you back up your wallet files regularly. If you\nhave not imported externally-produced keys, it only necessary to have backed up\nthe wallet's key storage file.\n",
  "id": 123
}
$ 
```

Although a future `cli_wallet` parser will handle embedded `\n`s, wouldn’t it be clearer to return a `Vec<String>` with one entry per line?

zallet/build.rs
```diff
@@ -137,3 +150,78 @@ impl Cli {
...
+    let methods_ast = syn::parse_file(&methods_rs)?;
+
+    let rpc_trait = methods_ast
+        .items
+        .iter()
+        .find_map(|item| match item {
+            syn::Item::Trait(item_trait) if item_trait.ident == "Rpc" => Some(item_trait),
+            _ => None,
+        })
+        .expect("present");
```
oxarbitrage, line 166: Optional:
```suggestion
        .expect("Rpc trait is present in methods file");
```

zallet/build.rs
```diff
@@ -137,3 +150,78 @@ impl Cli {
...
+                                    contents.push_str(&escaped);
+                                    contents.push_str("\\n");
+                                }
+                            }
+                        }
+                    }
+
+                    contents.push_str("\",\n");
+                }
+            }
```
oxarbitrage, line 216: It looks like this could be simplified but i am not sure if it will worth the effort. If you are interested i can try to send a suggestion.

oxarbitrage (approved):
Tested ACK with some comments.

Thanks!

str4d:
> Although a future `cli_wallet` parser will handle embedded `\n`s, wouldn’t it be clearer to return a `Vec<String>` with one entry per line?

The intent here was to ensure that it continues to work the same way that `zcash-cli help` worked (and indeed this method works with `zcash-cli help`). Sending the output as an array of strings would break all existing tooling that expects newline-separated lines.

For structured output I'm more interested in OpenRPC.

PR #129, opened 2025-05-01
rpc: Add OpenRPC support

str4d:
This adds minimal boilerplate to the JSON-RPC method implementations (which were already pretty regular), to enable the OpenRPC document to be generated. The generator itself enforces correct addition of future RPC methods via the type system.

The only duplication of content is in the method parameters. I have ideas for how this could be reduced, but it will do for now.


daira (approved):
utACK

PR #130, opened 2025-05-01
rpc: Reduce duplication of param details in OpenRPC support

str4d:


nuttycom (approved):
utACK. I don't really understand the relationship between the magic constants and the arguments to the `call` methods; the magic constants seem to refer to a subset of the arguments; is what's happening simply that parameter documentation is only rendered when one of these constants is present?

str4d:
> I don't really understand the relationship between the magic constants and the arguments to the `call` methods; the magic constants seem to refer to a subset of the arguments; is what's happening simply that parameter documentation is only rendered when one of these constants is present?

The arguments to the `call` methods include the internal state along with the method parameters. The trait to look at (which is what the buildscript is also "looking" at) is the `methods::Rpc` "trait" (it's actually a template from which `jsonrpsee` generates the `RpcServer` trait); the methods of this trait are what define the JSON-RPC method arguments.

Issue #119, opened 2025-04-23
Decide on text encoding for all new `seedfp` arguments and fields


str4d:
[ZIP 32](https://zips.z.cash/zip-0032#seed-fingerprints) specifies how to derive seed fingerprints. It does not specify how to encode them as text strings.

Currently in JSON-RPC methods where we've added seed fingerprints, we've parsed and serialized them in hex using the plain byte order:
https://github.com/zcash/wallet/blob/a3be12860c70185b93209d4c67a398d17bc0daea/zallet/src/components/json_rpc/methods/get_new_account.rs#L40-L49

However, I just noticed that in the `z_exportwallet` format's comments in `zcashd`, we byte-reverse before the hex encoding, because the fingerprint is internally represented as the oft-maligned `uint256` type:
https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/wallet/rpcdump.cpp#L607
https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/zcash/address/zip32.h#L48
https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/uint256.cpp#L21-L25

We need to examine where else in the Zcash ecosystem seed fingerprints have been / are being exposed as text strings, and confirm whether or not the byte-reversed hex encoding is what should be considered canonical. If that's what we decide, then we also need to update Zallet ASAP before it's too late.

nuttycom:
> does that get exposed anywhere as strings rather than [bytes in the wallet DB](https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/wallet/walletdb.cpp#L1350-L1354)?

The `zewif-zcashd`  wallet export tool reads these bytes directly from the wallet database dump format, so it needs to take this byte-reversal into account in its operation as well; the encoding in the `zewif` format itself is as bytes, which doesn't pose a problem but the byte reversal needs to occur for all types being extracted that were encoded as u256 values.

str4d:
`zip32` has a `Debug` impl that renders plain-byte hex:
https://github.com/zcash/zip32/blob/92aec1f9f47aef55ed2b0aa64373af4aa1615ff5/src/fingerprint.rs#L10-L26
https://github.com/zcash/zip32/pull/14#discussion_r1525481307

[...]

PR #131, opened 2025-05-01
Reverse bytes in seed fingerprint hex encoding to match `zcashd`

str4d:
Part of zcash/wallet#119.


nuttycom:
I thought we were going to define a new encoding, and use that?

str4d:
Yes, that's what the TODOs in the PR are for, and why this PR does not close the issue. I just wanted to get this fixed ASAP so it does not propagate.

nuttycom (approved):
utACK 85878cbdc5acc2516f82e33d6c8c5f494c54f6ca

PR #133, opened 2025-05-06
Create Dockerfile

y4ssi:
This PR adds a multi-stage Docker build for the `zallet` binary, resulting in a minimal and secure container image.

### Build Stage (`builder`)
- Based on `rust:1-slim` (amd64).
- Installs only the necessary packages for building (`clang`, `libclang-dev`, `pkg-config`, `git`).
- Compiles the project in `--release` mode.
- Uses `strip` to reduce binary size.

### Runtime Stage (`distroless`)
- Uses `gcr.io/distroless/cc` for a minimal and secure runtime environment.
- Copies only the final `zallet` binary.
- Runs as a non-root user by default.
- No shell and no package manager, which minimizes the attack surface.

### Additional Changes

- Adds a GitHub Action to build the image and verify that the binary runs with at least one argument (more tests can be added in the future).
- Adds a GitHub Action to release the Docker image to Docker Hub.

This setup ensures small image size, strong security practices, and a clear separation between build and runtime environments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 # --- Stage 1: Build with Rust --- (amd64) (RUST version: 1.86.0)
 FROM rust:1.86.0-slim@sha256:57d415bbd61ce11e2d5f73de068103c7bd9f3188dc132c97cef4a8f62989e944 AS builder
 
@@ -11,18 +12,28 @@ RUN apt-get update && \
         git && \
     rm -rf /var/lib/apt/lists/*
 
-COPY . .
+COPY --link . .
 
-RUN cargo build --release && strip target/release/zallet
+# Build the zallet binary
+# Leverage a cache mount to ${CARGO_HOME} for downloaded dependencies,
+# and a cache mount to ${CARGO_TARGET_DIR} for compiled dependencies.
+RUN --mount=type=bind,source=zallet,target=zallet \
+    --mount=type=bind,source=Cargo.toml,target=Cargo.toml \
+    --mount=type=bind,source=Cargo.lock,target=Cargo.lock \
+    --mount=type=cache,target=/app/.cargo \
+    --mount=type=cache,target=/app/target/ \
+    cargo build --locked --release --package zallet --bin zallet && \
+    cp /app/target/release/zallet /usr/local/bin/
+
 
 # --- Stage 2: Minimal runtime with distroless ---
-FROM gcr.io/distroless/cc
+FROM gcr.io/distroless/cc AS runtime
 
-COPY --from=builder /app/target/release/zallet /usr/local/bin/zallet
+COPY --link --from=builder /usr/local/bin/zallet /usr/local/bin/
 
 # USER nonroot (UID 65532) — for K8s, use runAsUser: 65532
 USER nonroot
 
 WORKDIR /var/lib/zallet
 
-ENTRYPOINT ["/usr/local/bin/zallet"]
+ENTRYPOINT ["zallet"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,20 @@
+# Before the docker CLI sends the context to the docker daemon, it looks for a file
+# named .dockerignore in the root directory of the context. If this file exists, the
+# CLI modifies the context to exclude files and directories that match patterns in it.
+#
+# You may want to specify which files to include in the context, rather than which
+# to exclude. To achieve this, specify * as the first pattern, followed by one or
+# more ! exception patterns.
+#
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+
+# Exclude everything:
+#
+*
+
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
+# --- Stage 1: Build with Rust --- (amd64)
+FROM rustlang/rust:nightly-slim@sha256:e1fb83983ca45a2f2afe386077cdf6873a7c0a496a5ae6e469b25dc1a90b1561 AS builder
```
str4d, line 2: - We should not be building with nightly Rust. Why did you select it?
- Where does `rustlang/rust` pull from?

https://hub.docker.com/_/rust suggests we should be using:
```suggestion
FROM rust:1-slim@sha256:TODO AS builder
```

.github/workflows/build-and-push-docker-hub.yaml
```diff
@@ -0,0 +1,35 @@
...
+      tags: ${{ steps.version_step.outputs.tags }}
+      zallet_version: ${{ steps.version_step.outputs.zallet_version }}
+    steps:
+    - id: version_step
+      run:  |
+        echo "tags=latest,${{ github.ref_name }},${{ github.sha }}" >> $GITHUB_OUTPUT
+        echo "zallet_version=$(echo ${{ github.ref_name }} | sed 's/v//g')" >> $GITHUB_OUTPUT
+
+  build_push:
+    uses: zcash/.github/.github/workflows/build-and-push-docker-hub.yaml@main
```
str4d, line 23: What is this doing? It looks recursive to me which makes no sense. Is this actually pulling from the `zcash/zcash` repo?

.github/workflows/test-docker-container.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build Docker image
+        run: |
+          docker build -t zallet .
+
+      - name: Run command inside Docker container
+        run: |
+          docker run --rm zallet -h
```
str4d, line 22: FYI the problem you last commit was fixing is that you tried to use `-help` as the flag, which is incorrect for Zallet. `--help` is what you should use (or `-h`), as the help text itself should show.

In general, `zcashd`'s support for single-hypen word flags to CLI is not something we are going to support in Zallet.

str4d (changes requested):
Reviewed b44296ab051c08a527d18977f91b88ae3e1c5024. In addition to my comments, the zizmor failures need to be fixed.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY . .
+
+RUN cargo build --release && strip target/release/zallet
+
+# --- Stage 2: Minimal runtime with distroless ---
+FROM gcr.io/distroless/cc
+
+COPY --from=builder /app/target/release/zallet /usr/local/bin/zallet
+
+USER nonroot
```
gustavovalverde, line 23: It would be ideal to have a user created with a predefined UID:GID (we recently had a discussion about this, and created an ADR for future reference: https://github.com/ZcashFoundation/zebra/blob/main/docs/decisions/devops/0001-docker-high-uid.md). Aside from the things explained there, it also helps for visibility in case a user needs to change the permissions on their docker-compose (or Kubernetes) to match the user running the container.

.github/workflows/build-and-push-docker-hub.yaml
gustavovalverde, line 1: **Optional:** This can be a reusable workflow. For example: 
- https://github.com/ZcashFoundation/z3/blob/dev/.github/workflows/sub-build-docker-image.yaml
- https://github.com/ZcashFoundation/zebra/blob/main/.github/workflows/sub-build-docker-image.yml

gustavovalverde (commented):
I left some comments for better UX and reproducibility.

gustavovalverde:
I opened https://github.com/zcash/wallet/pull/150 targeting this PR, with minimal design changes, but adding some of the optimizations mentioned in the comments

Dockerfile
```diff
@@ -0,0 +1,24 @@
+# --- Stage 1: Build with Rust --- (amd64)
+FROM rust:1-slim@sha256:57d415bbd61ce11e2d5f73de068103c7bd9f3188dc132c97cef4a8f62989e944 AS builder
+
+WORKDIR /app
```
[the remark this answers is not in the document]
gustavovalverde: That's correct. And that's something I completely missed as the actual place where this is more relevant is in the final image, where it would be preferable to have a `WORKDIR`. I'd have to evaluate the distroless image to confirm what's the actual directory.

P.S.: This also has an impact if `zallet` needs to write inside this directory. I don't recall if the default locations are relative to the HOME/WORDIR.

str4d (approved):
utACK 0347f862794c53a62c47bf865a7b3479e8d4f446. We can iterate on this in-repo as necessary in the lead-up to the first alpha release.

PR #142, opened 2025-05-17
sync: Poll non-ephemeral transparent addresses for mined UTXOs

str4d:


zallet/src/components/sync.rs
```diff
@@ -346,6 +366,75 @@ async fn recover_history(
...
+            .expect("the UTXO was detected via a supported address kind");
+
+            db_data.put_received_transparent_utxo(&output)?;
+        }
+
+        // Now wait on the chain tip to change.
+        // TODO: Once Zaino has an index over the mempool, monitor it for changes to the
+        // unmined UTXO set (which we can't get directly from the stream without building
+        // an index because existing mempool txs can be spent within the mempool).
+        while mempool_stream.next().await.is_some() {}
```
str4d, line 434: I spent a significant amount of time adding logic to turn mempool tx outputs into mempool UTXOs, only to remember at the end that later mempool transactions can spend outputs of earlier ones. Due to transactions not being guaranteed to leave the mempool across chain tip changes (e.g. during high load), that means we can't use the mempool stream to get UTXOs, and instead need to build and maintain an index (which is Zaino's job).

nuttycom (approved):
utACK bf79a55c6e74ff164578134835eba2e45bbeb9be

daira (commented):
Post-hoc ACK

PR #143, opened 2025-05-17
rpc: Fix `z_recoveraccounts` error message when treestate fetch fails

str4d:


oxarbitrage (approved):
Nice one, i saw the birthday error before and this one is more informative.

PR #144, opened 2025-05-19
Add `z_viewtransaction`

str4d:
Part of zcash/wallet#38.

### DIFF
--- a/zallet/src/components/json_rpc/methods/view_transaction.rs
+++ b/zallet/src/components/json_rpc/methods/view_transaction.rs
@@ -1,0 +1,480 @@
+use std::collections::BTreeMap;
+
+use documented::Documented;
+use jsonrpsee::core::RpcResult;
+use orchard::note_encryption::OrchardDomain;
+use rusqlite::{OptionalExtension, named_params};
+use schemars::JsonSchema;
+use serde::Serialize;
+use zcash_address::{
+    ToAddress, ZcashAddress,
+    unified::{self, Encoding},
+};
+use zcash_client_backend::data_api::WalletRead;
+use zcash_note_encryption::{try_note_decryption, try_output_recovery_with_ovk};
+use zcash_protocol::{ShieldedProtocol, TxId, consensus::Parameters, memo::Memo, value::Zatoshis};
+
+use crate::components::{
+    database::DbConnection,
+    json_rpc::{
+        server::LegacyCode,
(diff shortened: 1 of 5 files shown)


zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -0,0 +1,480 @@
...
+            .map_err(|e| {
+                LegacyCode::Database.with_message(format!("Failed to fetch sent-to address: {}", e))
+            })?
+            // If we don't have a cached recipient, fall back on an address that
+            // corresponds to the actual receiver.
+            .unwrap_or_else(fallback_addr))
+    }
+
+    if let Some(bundle) = tx.sapling_bundle() {
+        let incoming: BTreeMap<u16, (sapling::Note, Option<sapling::PaymentAddress>, [u8; 512])> =
```
nuttycom, line 261: Why do we need to re-do trial decryption? Don't we store sufficient information in the `*_received_notes` tables?

nuttycom (commented):
I'm very confused by this PR; why is it doing trial-decryption instead of querying the data stored in the wallet database?

This seems like an extremely awkward workaround; if we need to store more data, we should store more data.

zallet/src/components/json_rpc/methods/view_transaction.rs
str4d: I implemented this the same way as `zcashd` does to ensure identical semantics. Now it happens that `zcashd` decrypts _everything_ on-demand because it stores the shielded information once non-relationally (in `CWalletTx`). So we may be able to switch to the data stored in the various tables, as a performance optimisation. But we would need to determine whether the _semantics_ of the information in those tables is consistent with what `z_viewtransaction` provides, which would take longer.

nuttycom (approved):
conditional utACK:

I don't like the existing `z_viewtransaction` semantics, in that, given certain import timing circumstances, the results of `z_viewtransaction` can appear to conflict with other wallet information, such as wallet balance. Additionally, I think that performing trial decryption here is not ideal.

The change I would prefer is that, in the backend, viewing key import should immediately result in re-evaluation of each transaction already known to the wallet. Then, the trial decryptions should be removed from this method, and `z_viewtransaction` should provide exactly the *wallet's* view of the transaction data, including information about sent outputs that cannot otherwise be recovered from the chain. As part of that, it should also include information about the transparent inputs and outputs that involve the wallet.

PR #149, opened 2025-05-20
Create deb-release.yml

y4ssi:
### Summary

This PR adds a new GitHub Actions workflow (`deb.yml`) that automatically builds and publishes a `.deb` package when a new GitHub release is created.

### What it does

* Triggers only on GitHub release creation (`release: created`)
* Builds the Rust project and creates a `.deb` package using `cargo-deb`
* Publishes the package to a Debian repository hosted on Google Cloud Storage

### Why

This automates the release process, ensuring every versioned release includes a matching `.deb` package without manual steps.


str4d:
Hmm, there are also other things we should be doing, but I'll propose them in a follow-up PR:
- Build with latest stable Rust (there's no need for our own binaries to use MSRV; we only need to ensure that the build environment has a sufficiently-old glibc for the target platforms).
- Build for arm64, not just `x86_64`.
- Generate artifact attestations (because it's easy for us to do, even though they provide relatively little value with how GitHub chose to implement them).
- Upload the Debian packages to the releases (for manual installation) in addition to our APT repo.
- Test that we can install and use the resulting Debian package.

nuttycom (approved):
utACK

.github/workflows/deb-release.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Mirror apt.z.cash
+        shell: bash
+        run: |
+          aptly mirror create zcash https://apt.z.cash bookworm
+
+      - name: Create local snapshot of apt.z.cash with zallet
+        shell: bash
+        run: |
+          aptly repo create --distribution bookworm --component main zcash_local
+          aptly repo add zcash_local target/debian/zallet_*_amd64.deb
```
nuttycom, line 87: What is `zcash_local` here?

daira (approved):
utACK

PR #152, opened 2025-05-22
Add `migrate_zcashd_wallet` command.

nuttycom:
This adds support for importing the secret key material from a `zcashd` `wallet.dat` file into `zallet`.

It does not at present support import of wallet history; a wallet initialized using this command will perform recovery via scanning the chain.


zallet/src/components/json_rpc/methods/list_addresses.rs
```diff
@@ -181,21 +181,37 @@ pub(crate) fn call(wallet: &DbConnection) -> Response {
...
+                        diversifier_index: match address_info.source() {
+                            zcash_client_backend::data_api::AddressSource::Derived {
+                                diversifier_index,
+                            } => diversifier_index.into(),
+                            zcash_client_backend::data_api::AddressSource::Imported => {
+                                error!(
+                                    "Unified address {} lacks HD derivation information.",
+                                    address
+                                );
+                                return Err(RpcErrorCode::InternalError.into());
```
nuttycom, line 196: TODO

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -0,0 +1,664 @@
...
+/// A source string to identify an account as being derived from the mnemonic HD seed used for
+/// key derivation after the zcashd v4.7.0 upgrade.
+pub const ZCASHD_MNEMONIC_SOURCE: &str = "zcashd_mnemonic";
+
+impl AsyncRunnable for MigrateZcashdWalletCmd {
+    async fn run(&self) -> Result<(), Error> {
+        let config = APP.config();
+
+        // Start monitoring the chain.
+        let (chain_view, _chain_indexer_task_handle) = ChainView::new(&config).await?;
```
str4d, line 55: Non-blocking: if we start any component that fires off a background task, we need to shut it down as well. This particular callsite will be altered by #237 so I will fix this then.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -0,0 +1,664 @@
...
+                        ZCASHD_LEGACY_ACCOUNT,
+                        derivation.as_ref().and_then(|d| d.legacy_address_index()),
+                    ))?
+                    .is_some(),
+                None => db_data.get_account_for_ufvk(&ufvk)?.is_some(),
+            };
+
+            if !account_exists {
+                db_data.import_account_ufvk(
+                    &format!("zcashd legacy sapling {}", idx),
```
str4d, line 431: Non-blocking, but for the record this will be incorrect for imported Sapling keys (which are not associated with any legacy account) until #263 is done.

str4d (changes requested):
utACK cb94bf3a9ee8a1f69d784f578924aa4ca3fad2dc with nits, but blocked on knowing what dependency version we are depending on.

Cargo.toml
```diff
@@ -123,17 +130,20 @@ age = { git = "https://github.com/str4d/rage.git", rev = "84dc1e9f641994388f107c
...
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "10caf455e3f52744b5392af226a408b05721f70f" }
+
+zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
+zewif-zcashd = { git = "https://github.com/zcash/zewif-zcashd.git", rev = "02a98d6236e24819904e084180da9ba0f5c9b5d0" }
```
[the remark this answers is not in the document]
str4d: It is not on https://github.com/zcash/zewif-zcashd/pull/1 which is very confusing to me, because I _do_ now see it on that branch.

str4d (changes requested):
utACK f53923ba5b853454e605f43a15a3007381b2c4b3 however CI is still failing. The tests now need to have the new features enabled; something like this in `zallet/Cargo.toml` should work:
```toml
[[test]]
name = "cli_tests"
required-features = ["zcashd-import"]
```

Place this between `[package.metadata.deb]` and `[dependencies]` (I always try to follow the section order in [The Manifest Format](https://doc.rust-lang.org/cargo/reference/manifest.html) for consistency).

str4d (approved):
utACK e46fb92964cce7f388579eb9b1f0e0971f30e87c

Issue #15, opened 2025-01-30
Replace `f64` ZEC values in JSON-RPC output with a fixed-point decimal type


str4d:
There are a bunch of `zcashd` JSON-RPC methods that return decimal ZEC values. Currently to return those with `jsonrpsee` and its `serde` interface, we use an `f64`, but that is fragile.

In `zcashd` we instead used [an explicit fixed-point decimal string representation that was then encoded as a number](https://github.com/zcash/zcash/blob/a3435336b0c561799ac6805a27993eca3f9656df/src/rpc/server.cpp#L122-L130). I'm not quite sure how to replicate this via `serde`, but we need to figure it out.

PR #158, opened 2025-06-06
Add partial support for `z_gettotalbalance`

nuttycom:
The `allowwatchonly` jsonrpc parameter is not yet supported, as it requires changes to `WalletRead::get_wallet_summary`.

### DIFF
--- a/zallet/src/components/json_rpc/methods/z_get_total_balance.rs
+++ b/zallet/src/components/json_rpc/methods/z_get_total_balance.rs
@@ -9,13 +9,13 @@ use crate::components::{
     json_rpc::{server::LegacyCode, utils::value_from_zatoshis},
 };
 
-/// Response to a `z_listaccounts` RPC request.
+/// Response to a `z_gettotalbalance` RPC request.
 pub(crate) type Response = RpcResult<ResultType>;
-pub(crate) type ResultType = ZGetTotalBalance;
+pub(crate) type ResultType = TotalBalance;
 
-/// The number of notes in the wallet.
+/// The total value of funds stored in the wallet.
 #[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
-pub(crate) struct ZGetTotalBalance {
+pub(crate) struct TotalBalance {
     /// The total value of unspent transparent outputs, in ZEC
     transparent: f64,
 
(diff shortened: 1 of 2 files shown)


zallet/src/components/json_rpc/methods/z_get_total_balance.rs
```diff
@@ -0,0 +1,76 @@
...
+use schemars::JsonSchema;
+use serde::Serialize;
+use zcash_client_backend::data_api::WalletRead;
+
+use crate::components::{
+    database::DbConnection,
+    json_rpc::{server::LegacyCode, utils::value_from_zatoshis},
+};
+
+/// Response to a `z_listaccounts` RPC request.
```
str4d, line 12: Copy-pasta:
```suggestion
/// Response to a `z_gettotalbalance` RPC request.
```

zallet/src/components/json_rpc/methods/z_get_total_balance.rs
```diff
@@ -0,0 +1,76 @@
...
+pub(crate) type Response = RpcResult<ResultType>;
+pub(crate) type ResultType = ZGetTotalBalance;
+
+/// The number of notes in the wallet.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct ZGetTotalBalance {
```
str4d, line 18:
```suggestion
pub(crate) struct TotalBalance {
```
etc

str4d (approved):
ACK f5b08c8c2d18e2ad0e29ac5a7161b4c5c82b10d3. I tested this against my testnet wallet, and it shows balance. However, it also shows that we need to fix #15 for the alpha:
```json
{
  "transparent": 5.0,
  "private": 14.29997,
  "total": 19.299970000000002
}
```

Additionally, the API shows way less balance than my testnet wallet does in `zcashd`, but I'm not making that a blocker for this PR as I suspect the issue lies elsewhere (either in my testing setup, or Zallet not detecting funds in the legacy pool reliably yet).

zallet/src/components/json_rpc/methods/z_get_total_balance.rs
```diff
@@ -0,0 +1,76 @@
...
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct TotalBalance {
+    /// The total value of unspent transparent outputs, in ZEC
+    transparent: f64,
+
+    /// The total value of unspent Sapling and Orchard outputs, in ZEC
+    private: f64,
+
+    /// The total value of unspent shielded and transparent outputs, in ZEC
+    total: f64,
```
str4d, line 26: Bug: I just realised that these outputs are formatted explicitly as strings in `zcashd`, not JSON numbers as done here. I'll fix that at the same time as #15.

PR #160, opened 2025-06-09
Explicitly depend on `tracing` with required feature flags

str4d:
The `attributes` feature is ambiently enabled in our pinned MSRV dependencies, but something in a recent change to latest dependencies means that it is no longer ambiently enabled. We should always ensure that non-default feature flags we depend on are explicitly enabled.


nuttycom (approved):
utACK 199ed28bcb394c4090c58c7c4a318447524b10f9

PR #162, opened 2025-06-12
Detect TDR errors corresponding to `TransactionStatus::TxidNotRecognized`

str4d:
This prevents Zallet from crashing when it encounters a coinbase transaction. It will however indefinitely re-request the null txid until we bring in zcash/librustzcash#1821.

### DIFF
--- a/zallet/src/components/sync.rs
+++ b/zallet/src/components/sync.rs
@@ -7,8 +7,11 @@ use transparent::{
     address::Script,
     bundle::{OutPoint, TxOut},
 };
+use zaino_fetch::jsonrpsee::error::JsonRpSeeConnectorError;
 use zaino_proto::proto::service::GetAddressUtxosArg;
-use zaino_state::{FetchServiceSubscriber, LightWalletIndexer as _, ZcashIndexer};
+use zaino_state::{
+    FetchServiceError, FetchServiceSubscriber, LightWalletIndexer as _, ZcashIndexer,
+};
 use zcash_client_backend::{
     data_api::{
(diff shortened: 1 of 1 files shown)


zallet/src/components/sync.rs
```diff
@@ -489,8 +492,15 @@ async fn data_requests(
...
-                        // `TransactionStatus::TxidNotRecognized`.
+                        // TODO: Zaino is not correctly parsing the error response, so we
+                        // can't look for `LegacyCode::InvalidAddressOrKey`. Instead match
+                        // on these three possible error messages:
+                        // - "No such mempool or blockchain transaction" (zcashd -txindex)
+                        // - "No such mempool transaction." (zcashd)
+                        // - "No such mempool or main chain transaction" (zebrad)
+                        Err(FetchServiceError::JsonRpcConnectorError(
+                            JsonRpSeeConnectorError::JsonRpSeeClientError(e),
+                        )) if e.contains("No such mempool") => TransactionStatus::TxidNotRecognized,
```
str4d, line 503: For reference, the error we get here is (example from `zcashd -txindex`):
```rust
FetchServiceError::JsonRpcConnectorError(
    JsonRpSeeConnectorError::JsonRpSeeClientError(
        "Error: Error status from node's rpc server: 500 Internal Server Error, {\"result\":null,\"error\":{\"code\":-5,\"message\":\"No such mempool or blockchain transaction. Use gettransaction for wallet transactions.\"},\"id\":327969}\n"
    )
)
```

PR #163, opened 2025-06-12
Use signals for chain tip changes & notify backend of transaction data request service.

nuttycom:


zallet/src/components/database/connection.rs
```diff
@@ -20,7 +20,8 @@ use zcash_client_backend::{
 use zcash_client_sqlite::{WalletDb, util::SystemClock};
 use zcash_primitives::{block::BlockHash, transaction::Transaction};
 use zcash_protocol::{ShieldedProtocol, consensus::BlockHeight, value::Zatoshis};
-use zip32::{DiversifierIndex, fingerprint::SeedFingerprint};
+use zip32::DiversifierIndex;
+use zip32::fingerprint::SeedFingerprint;
```
str4d, line 24: Unclear what happened here. Remove this change from the second commit.

zallet/src/components/sync.rs
```diff
@@ -574,39 +574,44 @@ async fn data_requests(
...
+                    // information immediately without exposing correlations between addresses to
+                    // untrusted parties, so we can ignore the `request_at` field.
+
+                    // TODO: we're making the *large* assumption that the chain data doesn't update
+                    // between the multiple chain calls in this method. Ideally, Zaino will give us
+                    // a "transactional" API so that we can ensure internal consistency; for now,
+                    // we pick the chain height as of the start of evaluation as the "evaluated-at"
+                    // height for this data request, in order to not overstate the height for which
+                    // all observations are valid.
+                    let chain_tip = chain.chain_height().await?;
```
str4d, line 593: This helps to put https://github.com/zcash/librustzcash/pull/1824 into context. Maybe this kind of client consideration is what needs to go into the documentation of `WalletWrite::notify_address_checked`? (Or maybe it becomes unnecessary if that API is changed to enable the backend to detect divergences.)

zallet/src/components/sync.rs
```diff
@@ -574,39 +574,44 @@ async fn data_requests(
...
-                    tx_status_filter,
-                    output_status_filter,
-                } => {
-                    let address = address.encode(params);
+                TransactionDataRequest::TransactionsInvolvingAddress(req) => {
+                    info!(
+                        "Checking for new UTXOs at address {}",
+                        req.address().encode(params)
+                    );
+
```
str4d, line 582: Do we need this in addition to the log line below? I made that one `debug!` in 1606318fecf8d058961f6e90731fca471a6f04fd to reduce noise, and I think this `info!` line is still going to be somewhat noisy (even if only occurring on each new block), or at least not much less noisy than the `debug!` line (in which case we could just make that one `info!` again instead of logging twice in debug mode).

str4d (changes requested):
Reviewed a6842520b4a448b94fc8fa66fa19a162992099ac.

Cargo.toml
```diff
@@ -110,16 +110,16 @@ tonic = "0.13"
...
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "b3727870470e58d188207e2aa663d66163faa64c" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "76e1ce3054442311f4b642e4314d00b7d04c7097" }
```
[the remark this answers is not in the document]
nuttycom: Updated to zcash/librustzcash@caae2b9cc8d68498d862d2cff2e6e150902089fa, which will be stable and valid in the history once zcash/librustzcash#1824 merges.

zallet/src/components/sync.rs
nuttycom: Removed.

Issue #113, opened 2025-04-18
cli: Add a `zcash-cli` equivalent for interacting with the JSON-RPC API


str4d:
We need some easy way to interact with the JSON-RPC interface for at least debugging and testing. We can use `zcash-cli` for any unaltered APIs, but if there are differences in arguments (#41) or entirely new methods, `zcash-cli` can't be used unaltered.

I'm not sure whether we want to make this something we intend people to use in production; there people should instead rely on JSON-RPC client libraries (especially as we have JSON-RPC 2.0 support from the start). Maybe what would be most useful here is some kind of REPL?

PR #165, opened 2025-06-17
cli: Add `zallet rpc`

str4d:
Part of zcash/wallet#113.

### DIFF
--- a/zallet/src/cli.rs
+++ b/zallet/src/cli.rs
@@ -53,6 +53,10 @@ pub(crate) enum ZalletCmd {
 
     /// Import a BIP 39 mnemonic phrase into the wallet.
     ImportMnemonic(ImportMnemonicCmd),
+
+    /// Communicate with a Zallet wallet's JSON-RPC interface.
+    #[cfg(feature = "rpc-cli")]
+    Rpc(RpcCliCmd),
 }
 
 /// `start` subcommand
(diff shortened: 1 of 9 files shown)


str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/44d68d33f0b2ae6c16fafb36395999efa4572533..13b68f07c13dbceba0aaf486e36230757f096180) to fix parameter-handling code (I thought `jsonrpsee` was handling the JSON parsing, but it actually requires the caller to provide it with correctly-typed parameters).

str4d:
[Rebased on `main`](https://github.com/zcash/wallet/compare/13b68f07c13dbceba0aaf486e36230757f096180..709b68bf0cf2f5d137f8a3b8ed045960d38b41b6) to fix trivial merge conflict in `Cargo.toml`.

zallet/src/cli.rs
```diff
@@ -128,3 +132,15 @@ pub(crate) struct GenerateMnemonicCmd {}
...
+/// `rpc` subcommand
+#[cfg(feature = "rpc-cli")]
+#[derive(Debug, Parser)]
+#[cfg_attr(outside_buildscript, derive(Command))]
+pub(crate) struct RpcCliCmd {
+    /// The JSON-RPC command to send to Zallet.
+    pub(crate) command: String,
+
+    /// Any parameters for the command.
+    pub(crate) params: Vec<String>,
```
str4d, line 145: I'd like to support both array params and map params:
```
$ zallet rpc z_gettotalbalance null true
$ zallet rpc z_gettotalbalance --include_watch_only=true
```
but I don't believe Clap supports arbitrary unnamed non-positional args, and I don't have time to write our own parser to implement it.

PR #166, opened 2025-06-18
Overhaul config

str4d:


zallet/tests/cmd/migrate_zcash_conf_mainnet.toml
```diff
@@ -16,11 +24,17 @@ identity = ""
 
 [limits]
 
+[note_management]
+
+[outputs]
```
str4d, line 29: Currently every section header is required to be present in the config file (vs individual fields which can almost entirely be omitted if the defaults are fine). Do we want that to remain the case, or should I experiment with making entire sections optional?

I think I'd still keep the section headers uncommented in the example config (as otherwise uncommenting a commented-out field would cause it to apply to a different section, leading to confusing error messages).

zallet/src/config.rs
```diff
@@ -66,53 +43,36 @@ pub struct ZalletConfig {
...
             builder: Default::default(),
+            consensus: Default::default(),
+            features: Default::default(),
             indexer: Default::default(),
             keystore: Default::default(),
             limits: Default::default(),
+            note_management: Default::default(),
+            outputs: Default::default(),
             rpc: Default::default(),
+            storage: Default::default(),
```
str4d, line 71: Other sections I am toying with locally:
```toml
[limits.scanning]
[logging]
[metrics]
[metrics.prometheus]
[rpc.auth]
```

zallet/tests/cmd/example_config.out/zallet.toml
```diff
@@ -101,6 +119,65 @@ network = "main"
...
+# double-spends the funds.
+#
+# Values smaller than `trusted_confirmations` are ignored.
+#untrusted_confirmations = 10
+
+
+# Outputs configuration section.
+#
+# TODO: Rename this, and/or merge with `[storage]` in some way.
+[outputs]
```
str4d, line 167: I don't like this name but couldn't think of anything better, so it's a placeholder. I envision this section as being where you configure how Zallet reaches out to other processes (either directly or indirectly) with its own data (as opposed to reaching out to e.g. Zebra for Zebra's data).

zallet/src/config.rs
```diff
@@ -282,6 +288,115 @@ impl LimitsSection {
...
 }
 
+/// Note management configuration section.
+///
+/// TODO: Decide whether this should be part of `[builder]`.
+#[derive(Clone, Debug, Default, Deserialize, Serialize, Documented, DocumentedFields)]
+#[serde(deny_unknown_fields)]
+pub struct NoteManagementSection {
+    /// The minimum value that Zallet should target for each shielded note in the wallet.
+    pub min_note_value: Option<u32>,
```
nuttycom, line 298: I'd like to think of a better name for this field; it's a little bit ambiguous. Maybe `min_managed_note_value`?

zallet/src/config.rs
```diff
@@ -282,6 +288,115 @@ impl LimitsSection {
...
+
+    /// The target number of shielded notes with value at least `min_note_value` that
+    /// Zallet should aim to maintain within each account in the wallet.
+    ///
+    /// If an account contains fewer such notes, Zallet will split larger notes (in change
+    /// outputs of other transactions) to achieve the target.
+    pub target_output_count: Option<NonZeroU16>,
```
nuttycom, line 305: It's probably a little confusing to have both `note` and `output` used more or less interchangeably in this section. We should either go with `min_note_value` and `target_note_count` or `min_output_value` and `target_output_count`.

zallet/tests/cmd/example_config.out/zallet.toml
```diff
@@ -101,6 +91,23 @@ network = "main"
...
+
+# Whether the wallet should broadcast transactions.
+#broadcast = true
+
+# Directory to be used when exporting data.
+#export_dir = UNSET
+
+# Execute command when a wallet transaction changes.
```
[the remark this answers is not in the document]
str4d: In `zcashd` it means "whenever `CWallet::AddToWallet` is called with a transaction, run the command `-walletnotify` using its txid". This is triggered by various events:
- Whenever a newly-created transaction is finished (via `CWallet::CommitTransaction`):
  - Every Bitcoin-inherited money-sending JSON-RPC method.
  - `z_sendmany`
  - `z_shieldcoinbase`
  - `z_mergetoaddress`
  -  Sprout-to-Sapling migration txs.
- Whenever the status of the transaction in the chain changes (via `CWallet::AddToWalletIfInvolvingMe`):
  - Everywhere `SyncWithWallets` is called:
    - When a transaction is added to the mempool.
    - When a block containing the tx is disconnected (tx moves from "mined" to "in mempool")
    - When a block connection moves a tx from "in mempool" to "mined".
    - When a block connection moves a tx from "in mempool" to "conflicted".
  - `zcashd -rescan`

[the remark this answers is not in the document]
nuttycom: This should be added to the documentation.

zallet/tests/cmd/example_config.out/zallet.toml
```diff
@@ -101,6 +119,65 @@ network = "main"
 #orchard_actions = 50
```
str4d, line 119: Move this to a `[builder.limits]` section, due to it being irrelevant to non-builder contexts.

zallet/src/config.rs
str4d: Resolution: for now we are not going to support transparent output management in Zallet, so as long as there are RPCs that enable wallet operators to manually mangage their outputs (e.g. `z_mergetoaddress` to combine transparent inputs), then that is sufficient for Zallet's target audience.

zallet/tests/cmd/example_config.out/zallet.toml
```diff
@@ -101,6 +119,65 @@ network = "main"
...
+# An untrusted TXO is a TXO received by the wallet that is not trusted (in the sense
+# used by the `trusted_confirmations` setting).
+#
+# This setting is a trade-off between latency and security: a smaller value makes
+# trusted TXOs spendable more quickly, but the spending transaction has a higher
+# risk of failure if the sender of the receiving transaction is malicious and
+# double-spends the funds.
+#
+# Values smaller than `trusted_confirmations` are ignored.
+#untrusted_confirmations = 10
```
str4d, line 161: For now, these will be treated as builder-specific (because most of the viewing RPCs have `minconf` defaults that we need to preserve). In future I'd like these to become view-wide defaults, but I guess we can do that as either an experimental or deprecated feature.

str4d:
[Rebased on `main`](https://github.com/zcash/wallet/compare/de045584b361f91281e51cbf3a5b4308ea8faf7a..61b987aea96149a5fa913cdfef5a639269381e82).

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/61b987aea96149a5fa913cdfef5a639269381e82..d1badb73459f17c68cae1ef7d4d4146c1bdb2f06) to address some review comments.

zallet/src/config.rs
```diff
@@ -132,6 +137,78 @@ impl ConsensusSection {
...
+    /// along with any changes you need to make to your usage of Zallet.
+    pub as_of_version: String,
+
+    /// Enable support for the "legacy pools" of funds.
+    ///
+    /// This enables scanning and tracking of funds sent to addresses generated via the
+    /// `getnewaddress` and `z_getnewaddress` RPC methods (such as in wallets imported
+    /// from a `zcashd` node). It also enables any RPC methods that only work with the
+    /// legacy pool of funds, such as `getbalance`.
+    pub legacy_pools: Option<bool>,
```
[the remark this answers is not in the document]
str4d: > When we import a zcashd wallet, we'll always track the pools that we need to track as determined by what features/keys appear in that wallet data.

Tracking funds related to imported keys is _not_ the same as interpreting those imported keys in the legacy pool of funds approach. If you import a zcashd wallet into Zallet and don't have this option set to `true` (i.e. the default), the importer should warn you if the wallet contains funds that might previously have been interpreted with the legacy pool of funds view (i.e. either `getnewaddress` or `z_getnewaddress` was used with that wallet), and have the user confirm that either they are happy to proceed as-is, or enable this setting.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/19410bf7a86a5fe75af426713f16a39cf6126544..299e31377936dac8e1919748e23cfed6c6ce07a5) to update the README for the overhaul.

zallet/tests/cmd/example_config.out/zallet.toml
```diff
@@ -122,3 +199,15 @@ network = "main"
 # Timeout (in seconds) during HTTP requests.
 #timeout = 30
 
+
+# Storage configuration section.
+#
+# TODO: Rename this, and/or merge with `[outputs]` in some way.
+[storage]
```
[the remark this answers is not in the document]
str4d: I decided to make this a `[database]` section, as there are `deadpool` config settings we may want to expose here. And in general I've decided that storage paths will be settings within context-specific sections.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/299e31377936dac8e1919748e23cfed6c6ce07a5..53fe7682e3fe689a891bd5296b3f0718f4ae71df) to rename `[storage]` to `[database]`.

zallet/src/config.rs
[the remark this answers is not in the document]
str4d: In a pairing we remembered that what we'd actually planned to do here was have this be enabled for a specified seed fingerprint. I've implemented that, as well as significantly expanding the documentation.

nuttycom (approved):
utACK 9f3b94db3501ac54d4a811d496d38293348631be

Issue #153, opened 2025-05-23
Implement a `stop` RPC method


oxarbitrage:
Both `zebrad` and `zcashd` provide a stop method to gracefully shut down the server. This functionality is only available in regtest mode. I think we should implement something similar for the wallet, since we already know the mode it's running in. This would be especially useful for regtest integration.

I'm happy to give it a try, but I wanted to open an issue first in case there are already other alternatives being considered. Let me know what you think.

PR #168, opened 2025-06-18
Add a `stop` RPC method for regtest mode usage

oxarbitrage:
Add a `stop` RPC method, intended for use in regtest mode. The implementation is based on the approach used in Zebra:

https://github.com/ZcashFoundation/zebra/blob/v2.3.0/zebra-rpc/src/methods.rs#L1570-L1595

In Zebra, shutting down cleanly is non-trivial. The signal method was the most consistent solution we found. While there might be a better approach, this solution is simple and effective for regtest scenarios.

Tested manually and confirmed working.

Closes https://github.com/zcash/wallet/issues/153

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ rpassword = "7"
 http-body-util = "0.1"
 hyper = "1"
 jsonrpsee = "0.24"
+nix = "0.29" # `stop` RPC method
 rust_decimal = { version = "1.37", default-features = false, features = [
     "serde-arbitrary-precision",
     "serde-float",
--- a/zallet/src/components/json_rpc/methods/stop.rs
+++ b/zallet/src/components/json_rpc/methods/stop.rs
@@ -1,0 +1,31 @@
+use documented::Documented;
+use jsonrpsee::{core::RpcResult, types::ErrorCode as RpcErrorCode};
+use schemars::JsonSchema;
+use serde::Serialize;
+use zcash_protocol::consensus::{NetworkType, Parameters};
+
+use crate::components::database::DbHandle;
+
+/// Response to a `stop` RPC request.
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// The stop response.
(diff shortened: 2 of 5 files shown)


zallet/src/components/json_rpc/methods/stop.rs
```diff
@@ -0,0 +1,31 @@
...
+#[serde(transparent)]
+pub(crate) struct ResultType(());
+
+pub(crate) fn call(wallet: DbHandle) -> Response {
+    #[cfg(not(target_os = "windows"))]
+    match wallet.params().network_type() {
+        NetworkType::Regtest => {
+            match signal::raise(signal::SIGINT) {
+                Ok(_) => Ok(ResultType(())),
+                Err(_) => Err(RpcErrorCode::InternalError.into()),
```
oxarbitrage, line 24: Will be nice to send a message with the error here and in general, the `InternalError` alone can be confusing for the users.

Cargo.toml
```diff
@@ -61,6 +61,7 @@ hyper = "1"
 jsonrpsee = "0.24"
 schemars = "0.9"
 tower = "0.4"
+nix = "0.30" # `stop` RPC method
```
str4d, line 64: Please add dependencies alphabetically within their group.

Also, due to us depending on Zebra which also has this dependency, this should use the same version (0.29).
```suggestion
nix = "0.29" # `stop` RPC method
schemars = "0.9"
tower = "0.4"
```

str4d (changes requested):
Reviewed b0c6bdb305220bf2ea5cec973a25af0e169afe78. Change LGTM, just needs dependency fix.

str4d:
[Rebased on `main`](https://github.com/zcash/wallet/compare/d71342c5fe90958663f6f918d31567890fb6fe91..47466049b0d66c87d60d276847e9135766521e8e) to fix merge conflict in `Cargo.toml`, and fixed `Cargo.lock` bug.

str4d (approved):
utACK 47466049b0d66c87d60d276847e9135766521e8e

PR #172, opened 2025-06-20
rpc: Replace `f64`s with a type that wraps `Decimal`

str4d:
Also fixes a bug in `z_gettotalbalance` where the outputs were being returned as JSON numbers instead of strings.

Closes zcash/wallet#15.


zallet/src/components/json_rpc/methods/z_get_total_balance.rs
```diff
@@ -17,23 +18,13 @@ pub(crate) type ResultType = TotalBalance;
 #[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
 pub(crate) struct TotalBalance {
     /// The total value of unspent transparent outputs, in ZEC
-    transparent: f64,
+    transparent: String,
```
nuttycom, line 21: Why are these not `JsonZec`?

str4d: They were implemented as JSON strings containing numeric values, which is not the same thing as JSON numbers. (Specifically, they used `FormatMoney()` instead of `ValueFromAmount()` in `zcashd`.)

PR #173, opened 2025-06-20
rpc: Fix `z_listunspent` memo output to not truncate null bytes

str4d:
Null byte truncation only applies to text memos, per ZIP 302. The `zcashd` output does not truncate null bytes in the hex output.


nuttycom (approved):
utACK

Issue #136, opened 2025-05-12
sync: If one of the other tasks fails, `steady_state` loops instead of aborting


str4d:
Noticed this while debugging an unrelated issue: it seems that when the task handles are aborted, `steady_state` somehow enters a loop where it gets valid but unchanging information from `FetchServiceSubscriber`, which specifically includes its mempool stream always returning `None`. This is a problem because the mempool stream is documented as being closed when a new block is mined; in reality it should not be allowed to open.

Putting aside the Zaino bug, we should figure out why the abort is not ending the steady-state task at its next `await` point.

PR #175, opened 2025-06-26
Recover from `PrevHashMismatch` errors in recovery-range scanning.

nuttycom:


zallet/src/components/sync.rs
```diff
@@ -389,7 +389,21 @@ async fn recover_history(
...
                 )
-            })?;
+            });
+
+            // If a scan error is encountered, truncate to recover.
+            match scan_result {
+                Ok(_) => {}
+                Err(chain::error::Error::Scan(ScanError::PrevHashMismatch { at_height })) => {
+                    db_data
+                        .truncate_to_height(at_height - 10)
```
str4d, line 399: This will break the internal state of the `steady_state` task. And given that it's supposed to be the `steady_state` task that handles reorgs, that suggests there is a different problem causing this bug. My current suspicion is unclean shutdown (I've seen this in a few different ways, such as #136).

Issue #182, opened 2025-07-17
Document migration strategies for zcashd wallet users


str4d:
This issue tracks documentation that may be needed by users migrating from the zcashd wallet.

pacu:
I audited the migration docs against the code on `main` (5aa0ef3) looking for implemented behavior that migrating users would need but that no book page currently covers. Checklist of gaps, for whoever picks up the documentation work:

- [ ] **`--no-scan`** exists on `migrate-zcashd-wallet` (`zallet-core/src/cli.rs:214`, in the CHANGELOG since 0.1.0-alpha.4) but is missing from the flag list in `book/src/cli/migrate-zcashd-wallet.md`.
- [ ] **Sapling viewing-key migration** — since 0.1.0-alpha.3 each key imported to zcashd via `z_importviewingkey` becomes its own view-only account, but the "What is migrated" list in `book/src/zcashd/README.md` doesn't mention viewing keys at all.
- [ ] **Hard-failure scenarios exist only as error strings** (`zallet-core/i18n/en-US/zallet_core.ftl`), in no book page: pre-Sapling wallets with no HD seed are unsupported (`err-migrate-wallet-seed-absent`); non-English mnemonics are rejected (`err-migrate-secret-non-english-mnemonic`); regtest wallets are rejected; and re-running a partial import against the same wallet DB is unsupported — the remedy is a fresh wallet (`err-migrate-wallet-partial-import`). Users only discover these mid-migration.
- [ ] **The `zcash.conf` → `zallet.toml` option mapping** (which options are mapped, warned about, or deliberately ignored — ~160 entries) exists only in `zallet-core/src/commands/migrate_zcash_conf.rs`. `book/src/cli/migrate-zcash-conf.md` is 22 lines and doesn't enumerate any of it; operators can't predict what their config will become without reading source. The trycmd fixtures under `backends/zebra/tests/cmd/` would make good doc examples.
- [ ] **Encrypted `wallet.dat` support** (interactive passphrase prompt, up to 3 attempts, added in 0.1.0-beta.1) is documented only in the CHANGELOG.
- [ ] **Feature availability in shipped binaries** — `book/src/guide/installation/README.md` never states whether the release binaries / Docker image / Debian package are built with `zcashd-import` (they are, per the Dockerfiles, flake.nix, and CI), so an operator can't tell from the docs whether `migrate-zcashd-wallet` is present in what they downloaded.
[...]

Issue #96, opened 2025-04-03
`zallet 0.1.0-alpha.1`: First alpha release


str4d:
This issue can be closed once it is possible to use Zallet as a wallet (rather than a thing you can run and poke). That means:
- Generating or importing spending key material.
- Scanning the chain to discover transactions.
- Viewing transactions.
- Spending funds.

This issue does not require being able to do the above in every payment pool or in every way; there just needs to be _some_ way to do each of the above core wallet functionalities.

PR #186, opened 2025-07-29
Define versioning (tags)

y4ssi:
We should follow [[Semantic Versioning](https://semver.org/)](https://semver.org/) for pre-release versions. For alpha releases, that would look like:

* `0.1.0-alpha`
* `0.1.0-alpha.1`
* `0.1.0-beta`
* `0.1.0-rc.1`
* `0.1.0` (final release)

### DIFF
--- a/.github/workflows/build-and-push-docker-hub.yaml
+++ b/.github/workflows/build-and-push-docker-hub.yaml
@@ -23,7 +23,7 @@ jobs:
         set -Eeuo pipefail
         IFS=$'\n\t'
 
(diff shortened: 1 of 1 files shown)


str4d:
We decided against using suffixes for this. Instead, 1.0.0 will be the first stable release; 0.1.0 will be the first alpha release (#96), and everything in between will be alpha releases until we decide to cut the first beta release (#182).

str4d:
In a pairing with @nuttycom we decided that actually for this first alpha we would use 0.1.0-alpha.1 for the first alpha; if we change our minds later again we can switch course.

str4d (approved):
utACK 664ce647edcef88cc95cf0f7b1061d8ba4a4b6b7

PR #198, opened 2025-08-20
database: Store wallet and version metadata for Zallet

### LINKED ISSUE #171: db: Add an extension table for Zallet version metadata
We should have a table that Zallet writes to after it has finished loading and checking the database. Then in future Zallet versions we can detect what Zallet version last wrote to the database, for determining version-relevant migration information.

We could alternatively just have a database migration for every single version bump, but that seems excessive for patch bumps, whereas a table with a single row we overwrite on each update is fine.

str4d:
Closes #171.


nuttycom (approved):
utACK https://github.com/zcash/wallet/pull/198/commits/c068f70d74e489e202e009e80d01d72d4625cf70

PR #207, opened 2025-08-23
Prepare `zallet rpc` for the alpha

### LINKED ISSUE #113: cli: Add a `zcash-cli` equivalent for interacting with the JSON-RPC API

str4d:
Closes #113.

### DIFF
--- a/book/src/cli/rpc.md
+++ b/book/src/cli/rpc.md
@@ -1,0 +1,3 @@
+# The `rpc` command
+
+TODO
--- a/book/src/cli/rpc.md
+++ b/book/src/cli/rpc.md
@@ -1,3 +1,57 @@
 # The `rpc` command
 
-TODO
+> Available on **crate feature** `rpc-cli` only.
+
+`zallet rpc` lets you communicate with a Zallet wallet's JSON-RPC interface from a
+command-line shell.
+
+- `zallet rpc help` will print a list of all JSON-RPC methods supported by Zallet.
+- `zallet rpc help <method>` will print out a description of `<method>`.
+- `zallet rpc <method>` will call that JSON-RPC method. Parameters can be provided via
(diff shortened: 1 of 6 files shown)


book/src/cli/rpc.md
```diff
@@ -1,3 +1,57 @@
...
+| `zcash-cli -conf=<file>`          | `zallet --config <file> rpc`       |
+| `zcash-cli -datadir=<dir>`        | `zallet --datadir <dir> rpc`       |
+| `zcash-cli -stdin`                | Not implemented                    |
+| `zcash-cli -rpcconnect=<ip>`      | `rpc.bind` setting in config file  |
+| `zcash-cli -rpcport=<port>`       | `rpc.bind` setting in config file  |
+| `zcash-cli -rpcwait`              | Not implemented                    |
+| `zcash-cli -rpcuser=<user>`       | Not implemented                    |
+| `zcash-cli -rpcpassword=<pw>`     | Not implemented                    |
+| `zcash-cli -rpcclienttimeout=<n>` | Not implemented, always 60s        |
+| Hostname, domain, or IP address   | Only IP address                    |
```
nuttycom, line 30: I don't understand this line in the context of the rest of the table.

nuttycom (approved):
utACK 1a9f5e35053dab00143d91ab7d10ba39b5b224d0

PR #215, opened 2025-08-28
Apply a temporary fix for a Zaino bug

str4d:
This can be removed once zingolabs/zaino#495 is fixed and pulled in.


nuttycom (commented):
If the Zaino team expects to turn this around quickly, I'd rather keep this unmerged and use it for local patching where necessary.

str4d:
Replaced by #224.

PR #216, opened 2025-08-28
rpc: Add transparent information to `z_viewtransaction`

### LINKED ISSUE #147: rpc: Consider optionally adding transparent-equivalent details to `z_viewtransaction`
`z_viewtransaction` exposes all shielded information that the wallet is able to decrypt for a transaction, without imposing the perspective of any particular account (other than treating outputs preferentially as "incoming" if any account can decrypt with its ivks, before checking for "outgoing" if any account can decrypt with its ovks). Adding transparent details would effectively save calls to `getrawtransaction` (because this is information that is already visible in the transaction bytes).

This would _not_ be equivalent to calling `gettransaction`, which has a different and Bitcoin-specific interpretation of its data that does not align with the existing shielded details exposed in `z_viewtransaction`.

str4d:
Closes #147.

### DIFF
--- a/book/src/zcashd/json_rpc.md
+++ b/book/src/zcashd/json_rpc.md
@@ -46,6 +46,17 @@ Changes to response:
 ### `z_viewtransaction`
 
 Changes to response:
+- Some top-level fields from `gettransaction` have been added:
+  - `status`
+  - `confirmations`
+  - `blockhash`, `blockindex`, `blocktime`
+  - `version`
+  - `expiryheight`, which is now always included (instead of only when a
+    transaction has been mined).
+  - `fee`, which is now included even if the transaction does not spend any
+    value from any account in the wallet, but can also be omitted if the
+    transparent inputs for a transaction cannot be found.
+  - `generated`
 - Information about all transparent inputs and outputs (which are always visible
   to the wallet) are now included. This causes the following semantic changes:
   - `pool` field on both inputs and outputs can be `"transparent"`.
--- a/book/src/zcashd/json_rpc.md
+++ b/book/src/zcashd/json_rpc.md
@@ -62,8 +62,10 @@ Changes to response:
   - `pool` field on both inputs and outputs can be `"transparent"`.
   - New fields `tIn` and `tOutPrev` on inputs.
   - New field `tOut` on outputs.
-  - `address` field on outputs is no longer only omitted if the output was
-    received on an account-internal address; use `walletInternal` for this.
+  - `address` field on outputs: in `zcashd`, this was omitted only if the output
+    was received on an account-internal address; it is now also omitted if it is
+    a transparent output to a script that doesn't have an address encoding. Use
+    `walletInternal` if you need to identify change outputs.
   - `memo` field on outputs is omitted if `pool = "transparent"`.
   - `memoStr` field on outputs is no longer only omitted if `memo` does not
     contain valid UTF-8.
--- a/book/src/zcashd/json_rpc.md
+++ b/book/src/zcashd/json_rpc.md
@@ -43,6 +43,20 @@ Changes to response:
   listed in a new `derived_transparent` field (an array of objects) instead of
   the `transparent` field.
 
+### `z_viewtransaction`
+
+Changes to response:
+- Information about all transparent inputs and outputs (which are always visible
+  to the wallet) are now included. This causes the following semantic changes:
+  - `pool` field on both inputs and outputs can be `"transparent"`.
+  - New fields `tIn` and `tOutPrev` on inputs.
+  - New field `tOut` on outputs.
+  - `address` field on outputs is no longer only omitted if the output was
+    received on an account-internal address; use `walletInternal` for this.
+  - `memo` field on outputs is omitted if `pool = "transparent"`.
+  - `memoStr` field on outputs is no longer only omitted if `memo` does not
+    contain valid UTF-8.
+
 ### `z_sendmany`
 
 Changes to parameters:
--- a/zallet/src/components/json_rpc/methods/view_transaction.rs
+++ b/zallet/src/components/json_rpc/methods/view_transaction.rs
@@ -32,8 +32,10 @@ pub(crate) struct Transaction {
     /// The transaction ID.
     txid: String,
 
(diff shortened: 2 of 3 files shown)


book/src/zcashd/json_rpc.md
```diff
@@ -43,6 +43,31 @@ Changes to response:
...
+    value from any account in the wallet, but can also be omitted if the
+    transparent inputs for a transaction cannot be found.
+  - `generated`
+- Information about all transparent inputs and outputs (which are always visible
+  to the wallet) are now included. This causes the following semantic changes:
+  - `pool` field on both inputs and outputs can be `"transparent"`.
+  - New fields `tIn` and `tOutPrev` on inputs.
+  - New field `tOut` on outputs.
+  - `address` field on outputs is no longer only omitted if the output was
+    received on an account-internal address; use `walletInternal` for this.
```
nuttycom, line 66: This description is confusing - it requires going and looking up exactly what the semantics zcashd had were, by looking at the implementation there. In general, I think that when describing changes from zcashd, we should also describe what the original semantics were because in a lot of cases they were undocumented.

zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -22,28 +33,94 @@ use crate::components::{
...
+    ///
+    /// Omitted if this transaction is not mined within a block in the current best chain.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    blockhash: Option<String>,
+
+    /// The index of the transaction within its block's `vtx` field.
+    ///
+    /// Omitted if this transaction is not mined within a block in the current best chain.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    blockindex: Option<u16>,
```
nuttycom, line 78: Is there a reason to use u16 here? I ask because the transparent `OutPoint` type uses a u32; I ran into this in the `z_listunspent` transparent additions, where to use a u16 I would have to (fallibly) convert to the narrower representation; since this struct is just used for serialization, why not use a `u32` and avoid that conversion?

zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -257,6 +366,95 @@ pub(crate) fn call(wallet: &DbConnection, txid_str: &str) -> Response {
...
+                t_in: Some(idx),
+                spend: None,
+                action: None,
+                txid_prev,
+                t_out_prev: Some(
+                    input
+                        .prevout()
+                        .n()
+                        .try_into()
+                        .expect("should always be small enough"),
```
nuttycom, line 407: Same with this conversion; I feel like we should just use `u32` instead?

zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -257,6 +366,95 @@ pub(crate) fn call(wallet: &DbConnection, txid_str: &str) -> Response {
...
+                                match wallet.get_transparent_address_metadata(account, &address) {
+                                    Ok(Some(metadata)) => Some(metadata.scope()),
+                                    _ => None,
+                                }
+                            });
+
+                    (
+                        Some(address.encode(wallet.params())),
+                        wallet_scope.is_none(),
+                        matches!(wallet_scope, Some(TransparentKeyScope::INTERNAL)),
```
nuttycom, line 437: Why does this not also count ephemeral addresses as internal?

zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -472,9 +676,141 @@ pub(crate) fn call(wallet: &DbConnection, txid_str: &str) -> Response {
...
+        wallet: &DbConnection,
+        chain: &FetchServiceSubscriber,
+        tx: &zcash_primitives::transaction::Transaction,
+        chain_height: BlockHeight,
+    ) -> Result<Self, SqliteClientError> {
+        let mined_height = wallet.get_tx_height(tx.txid())?;
+
+        let confirmations = {
+            match mined_height {
+                Some(mined_height) => (chain_height + 1 - mined_height) as i32,
```
nuttycom, line 727: We should use i64 here; even though we "know" the cast from block height to i32 is safe, using `i64::from(...)` would be cleaner.

zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -472,9 +676,141 @@ pub(crate) fn call(wallet: &DbConnection, txid_str: &str) -> Response {
...
+                    height: 0,
+                    hash: block_metadata.block_hash().0.to_vec(),
+                })
+                .await
+                .map_err(|_| SqliteClientError::ChainHeightUnknown)?;
+
+            let tx_index = block
+                .vtx
+                .iter()
+                .find(|tx| tx.hash == block_metadata.block_hash().0)
```
nuttycom, line 767: Bug: comparing the transaction id to the block hash?

nuttycom (changes requested):
One blocking bug.

zallet/src/components/json_rpc/methods/view_transaction.rs
str4d: We have consensus checks that Orchard bundles can have at most `2^16` actions, and the consensus transaction size limits have much lower limits for all the protocols, so I've tended to default to `u16` in all new places. But yeah, I guess as long as this is only used for serialization (which should be the case; we are never going to provide client libraries based on our code) then these relaxations should be fine.

zallet/src/components/json_rpc/methods/view_transaction.rs
str4d: Good catch; this is a bug in #36 that got copied over here.

zallet/src/components/json_rpc/methods/view_transaction.rs
str4d: Because they are not internal; they are a different scope, and they are addresses that are used to interact with external parties (in a specific controlled way) and thus are "exposed externally".

Also, per `zcashd`, `wallet_internal` is documented as `"true" if this is a change output`, and ephemeral addresses are not used for change outputs.

PR #193, opened 2025-08-05
rpc (z_listunspent): support the same arguments as `zcashd`s version of the RPC

nuttycom:
This updates the RPC API of `z_listunspent` to include all of the arguments supported by this API in `zcashd`, although at present the `include_watch_only` argument value is ignored.

Depends on zcash/librustzcash#1930
Closes #86.

PR #230, opened 2025-09-04
rpc (z_listunspent): Add transparent UTXOs to `z_listunspent` results.

nuttycom:
This builds atop #193

### DIFF
--- a/zallet/src/components/json_rpc/methods/list_unspent.rs
+++ b/zallet/src/components/json_rpc/methods/list_unspent.rs
@@ -8,9 +8,13 @@ use jsonrpsee::{
 use schemars::JsonSchema;
 use serde::Serialize;
 
+use transparent::keys::TransparentKeyScope;
 use zcash_client_backend::{
     address::UnifiedAddress,
-    data_api::{Account, AccountPurpose, InputSource, WalletRead, wallet::TargetHeight},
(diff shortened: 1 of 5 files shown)


zallet/src/components/json_rpc/methods/list_unspent.rs
```diff
@@ -204,6 +213,64 @@ pub(crate) fn call(
...
+            let wallet_internal = wallet
+                .get_transparent_address_metadata(account_id, utxo.recipient_address())
+                .map_err(|e| {
+                    RpcError::owned(
+                        LegacyCode::Database.into(),
+                        "WalletDb::get_transparent_address_metadata failed",
+                        Some(format!("{e}")),
+                    )
+                })?
+                .is_some_and(|m| m.scope() == TransparentKeyScope::INTERNAL);
```
str4d, line 253: Hmm: what does it mean if we know the account for a transparent address, but we don't have its metadata? That is, should `walletInternal = false` mean either "external" or "we don't know", or should this be considered an error case, or should `walletInternal` be optional (and omitted if we cannot determine it)?

str4d (approved):
utACK 56ab42cb15bcb868c2d3a7463daf4500f1df36e9 with a concern.

PR #238, opened 2025-09-05
rpc: Ignore non-mnemonic zcashd seeds for numeric account parameters

str4d:
Closes zcash/wallet#231.


nuttycom (approved):
utACK; after some consideration I think this is fine, because it will only be an observable difference in behavior for those users that have imported zcashd wallets.

Issue #233, opened 2025-09-04
rpc (z_viewtransaction): Report account UUID with each spend and receive entry.


nuttycom:
The correct way for a caller to determine whether an output is change, from their perspective, is to check the relationship between spends and outputs, potentially in combination with checking the value of the `walletInternal` flag.  The account UUID should replace the ZIP 32 account identifier in outputs, and the UUID for the account that controlled each input should be added to the input's spend entry.

We _may_ additionally, in the future, want to add some metadata about what we are able to determine about spends and/or outputs that do not belong to the wallet, but that's not something that we should do prior to the first Zallet release.

Issue #228, opened 2025-09-04
rpc (z_viewtransaction): Add a top-level map that describes the effect of this transaction on each involved account's balance.


nuttycom:
This should be a new field that contains a dictionary keyed by account UUID, and where the value is a map that describes the balance delta for each supported currency in the wallet; this will initially provide information just for the ZEC balance but after the introduction of ZSAs will include the balance delta for each asset identifier.

PR #239, opened 2025-09-05
rpc: Add `account_uuid` fields to spends and outputs in `z_viewtransaction`

str4d:
Closes zcash/wallet#233.

### DIFF
--- a/zallet/src/components/json_rpc/methods/view_transaction.rs
+++ b/zallet/src/components/json_rpc/methods/view_transaction.rs
@@ -14,7 +14,7 @@ use zcash_address::{
     unified::{self, Encoding},
 };
 use zcash_client_backend::data_api::WalletRead;
-use zcash_client_sqlite::error::SqliteClientError;
+use zcash_client_sqlite::{AccountUuid, error::SqliteClientError};
 use zcash_keys::encoding::AddressCodec;
 use zcash_note_encryption::{try_note_decryption, try_output_recovery_with_ovk};
 use zcash_protocol::{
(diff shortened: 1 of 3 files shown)


zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -302,9 +323,10 @@ pub(crate) async fn call(
             .with_raw(|conn| {
                 conn.query_row(
                     &format!(
-                        "SELECT txid, {output_prefix}_index, address, value
-                        FROM {pool_prefix}_received_notes
+                        "SELECT txid, {output_prefix}_index, accounts.uuid, address, value
+                        FROM {pool_prefix}_received_notes rn
                         JOIN transactions ON tx = id_tx
+                        JOIN accounts ON accounts.id = rn.account_id
```
nuttycom, line 347: This comment isn't really specific to this PR, but I think that taking the approach of looking up the spend by the nullifier runs a small risk of producing inconsistencies between different views in the wallet, because that's not how we look up spends anywhere else. The way I would have approached that particular issue (in the earlier PR, where I thought about this but didn't want to take time asking for changes on) would be to query the `v_received_note_spends` view to find all the outputs known by the wallet to be spent in the transaction in question.

zallet/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -450,27 +489,39 @@ pub(crate) async fn call(
...
+            .iter()
+            .zip(0..)
+            .filter_map(|(output, idx)| {
+                sapling_ivks.iter().find_map(|(account_id, ivk, scope)| {
+                    sapling::note_encryption::try_sapling_note_decryption(
+                        ivk,
+                        output,
+                        zip212_enforcement,
+                    )
+                    .map(|(n, a, m)| {
```
nuttycom, line 529: This is another example; we should already have decrypted all of these outputs, and should be able to just get the data from the database - there's no reason we should be trial-decrypting here again.

nuttycom (approved):
utACK; when we have time I would like this to rely mostly on wallet APIs, instead of the mix of direct database access and manual extraction of raw transaction data that we're doing here.

nuttycom (approved):
re-utACK 65a7f45a7cbd0c164c0b04590f4486244d468fe9

zallet/src/components/json_rpc/methods/view_transaction.rs
str4d: As I've said elsewhere, there *is* a reason: this method provides all information that the wallet can currently decrypt, so if a new key has just been imported but a rescan hasn't proceeded, then this API will still show the relevant spends or outputs in `spends` and `outputs`. After #228 the only thing that won't be accurate until a rescan completes is the `accounts` field, but that doesn't change the fact that it is useful to be able to see the details inside transactions as soon as we have the key material to do so.

PR #244, opened 2025-09-06
Switch to ECC fork of Zaino

str4d:
Includes a fix to a `zaino-state` bug that has completely blocked Zallet from working for several days (which we can't revert back from because the original update of Zaino was needed to fix a different bug). This allows Zallet to function while we wait on a full fix upstream.


nuttycom (approved):
utACK

PR #274, opened 2025-09-16
z_viewtransaction: Fix bugs for coinbase transactions

### LINKED ISSUE #269: z_viewtransaction: Fix it for coinbase transactions
#147 added transparent information to `z_viewtransaction`, and we also added top-level fields from `gettransaction` to it. However, this has created several problems:
- Coinbase transactions have a transparent input that doesn't spend any prior coin, which currently causes `z_viewtransaction` to panic.
- The `confirmations` field pulled over from `gettransaction` is not working for coinbase transactions, and is always set to -1.
- Another problem tracked in #268.

We need to fix `z_viewtransaction` so that it works, and the information it returns for coinbase transactions is equivalent to the information that `zcashd` provided in `gettransaction` in the fields we ported over.

str4d:
Closes #269.


nuttycom (approved):

daira (commented):
Post-hoc ACK

PR #276, opened 2025-09-16
migrate_zcashd_wallet: Print legacy account seedfp & error on duplicate/multiple imports.

### LINKED ISSUE #212: zcashd wallet migration CLI operation should report the legacy seed fingerprint that should be added to the config

nuttycom:
Closes #212.

### DIFF
--- a/zallet/i18n/en-US/zallet.ftl
+++ b/zallet/i18n/en-US/zallet.ftl
@@ -23,6 +23,9 @@
 -allow-multiple-wallet-imports = --allow-multiple-wallet-imports
 -datadir = --datadir
 
+-legacy_pool_seed_fingerprint = legacy_pool_seed_fingerprint
+-zallet_toml = zallet.toml
+
 ## Usage
 
 usage-header = Usage
@@ -73,6 +76,11 @@ migrate-alpha-code =
     this, use '{-allow-alpha-migration}'.
 
 migrate-config-written = {-zallet} config written to {$conf}
+
+migrate-wallet-legacy-seed-fp =
+    Importing zcashd legacy account for seed fingerprint '{$seed_fp}'. If you wish to
+    enable legacy zcashd semantics for wallet RPC methods, you should set
+    '{-legacy_pool_seed_fingerprint}' to this value in '{-zallet_toml}'.",
 
 ## General errors
 
--- a/zallet/i18n/en-US/zallet.ftl
+++ b/zallet/i18n/en-US/zallet.ftl
@@ -20,6 +20,7 @@
 -allow-warnings = --allow-warnings
 -allow-alpha-example = --this-is-alpha-code-and-you-will-need-to-recreate-the-example-later
 -allow-alpha-migration = --this-is-alpha-code-and-you-will-need-to-redo-the-migration-later
+-allow-multiple-wallet-imports = --allow-multiple-wallet-imports
 -datadir = --datadir
 
 ## Usage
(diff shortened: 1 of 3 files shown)


zallet/i18n/en-US/zallet.ftl
```diff
@@ -112,6 +112,12 @@ err-migrate-unknown-zcashd-option = Unknown {-zcashd} option '{$option}'
 err-failed-seed-fingerprinting =
     Zallet was unable to import invalid seed data, likely due to the seed having
     an invalid length.
+err-migrate-wallet-multi-import-disabled =
+    The `--allow-multiple-wallet-imports` flag must be provided to allow the
```
str4d, line 116:
```suggestion
    The `{-allow-multiple-wallet-imports}` flag must be provided to allow the
```
and add a corresponding term at the top of the file (to ensure the term is consistent across translations).

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -192,11 +193,59 @@ impl MigrateZcashdWalletCmd {
...
+                            derivation,
+                            key_source,
+                        } if key_source.as_ref() == Some(&ZCASHD_MNEMONIC_SOURCE.to_string()) => {
+                            found.push(derivation.clone());
+                        }
+                        _ => {}
+                    }
+
+                    Ok::<_, SqliteClientError>(found)
+                })?;
```
str4d, line 222: `try_fold` is an odd way to do this. Why not:
```suggestion
                .filter_map(|account_id| {
                    match db_data
                        .get_account(account_id) {
                        Err(e) => Some(Err(e)),
                        Ok(account) => match account.expect("account exists for just-retrieved id")
                            .source() {
                            AccountSource::Derived {
                                derivation,
                                key_source,
                            } if key_source.as_ref() == Some(&ZCASHD_MNEMONIC_SOURCE.to_string()) => {
                                Some(Ok(derivation.clone()))
                            }
                            _ => None
                        },
                    }
                })
                .collect::<Result<Vec<_>, _>>()?;
```

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -192,11 +193,59 @@ impl MigrateZcashdWalletCmd {
...
+
+        if !existing_zcash_sourced_accounts.is_empty() {
+            if allow_multiple_wallet_imports {
+                if let Some((seed, _)) = mnemonic_seed_data.as_ref() {
+                    let seed_fp =
+                        SeedFingerprint::from_seed(seed.expose_secret()).expect("valid length");
+                    if existing_zcash_sourced_accounts
+                        .iter()
+                        .any(|derivation| derivation.seed_fingerprint() == &seed_fp)
+                    {
```
str4d, line 240: If we don't use `existing_zcash_sourced_accounts` for anything other than this, then we should just collect the `SeedFingerprints` themselves directly into a `HashSet`, and then this could be:
```suggestion
                    if existing_zcash_sourced_accounts.contains(&seed_fp) {
```

str4d (changes requested):
Reviewed 5372b0bcf18fb737eaa9f3b993eff6855ba290bd

zallet/src/commands/migrate_zcashd_wallet.rs
nuttycom: I don't see that this is particularly any clearer? `try_fold` is a very common way of building lists.

str4d (approved):
utACK 5f41bdde9595f9628107a19826fd709a70268f33

zallet/src/commands/migrate_zcashd_wallet.rs
str4d: I disagree for the original list, but after the switch to just collecting fingerprints into a `HashSet` (which has a different cardinality than the original sequence), `try_fold` feels fine.

str4d (approved):
utACK 2391da46cb021da4ec4633ba9a358fea2fb87dc8

Issue #241, opened 2025-09-05
rpc: Ensure that all parameter names inherited from Bitcoin Core match it


str4d:
https://github.com/bitcoin/bitcoin/pull/8811 was never backported to `zcashd`, so Zallet will be the first time that the `zcashd` wallet RPCs have supported named parameters. We should ensure that for Bitcoin Core-inherited RPCs, the parameter names match. And for Zcash RPCs, we should go through and ensure that the paraemter names are consistent and how we want them.

PR #277, opened 2025-09-16
rpc: Adjust JSON-RPC parameter names to match Bitcoin Core

str4d:
This is helpful to reduce the delta for JSON-RPC 2.0 clients.

Closes zcash/wallet#241.


nuttycom (approved):
utACK

Issue #267, opened 2025-09-12
Add license-collection logic to binary creation processes


str4d:
The code distributed in this repository falls under the MIT and Apache 2.0 licenses, with license files ([LICENSE-MIT](https://github.com/zcash/wallet/blob/main/LICENSE-MIT) and [LICENSE-APACHE](https://github.com/zcash/wallet/blob/main/LICENSE-APACHE)) distributed alongside it. Publishing the `zallet` crate to crates.io (once we no longer depend on pinned revisions of dependencies and are able to) also only distributes code from this repository.

However, once we produce binaries (e.g. Debian packages, Docker images), those binaries will also include code from our dependencies, and we need to ensure that their licenses are also provided to downstream users of those binaries.

For `zcashd`, we maintain the [`contrib/debian/copyright`](https://github.com/zcash/zcash/blob/master/contrib/debian/copyright) file by hand; it is included into Debian packages (which IIRC were the only binaries published), but also by way of being checked into the repo, it was published along with the zcashd source code.

Currently in Zallet, we track the licenses of our dependencies with `cargo-deny` in a [`deny.toml`](https://github.com/zcash/wallet/blob/main/deny.toml) file: any dependency that isn't MIT or Apache-2.0 is listed there. But that alone is not sufficient for some licenses at binary distribution time.

We should decide on an approach for Zallet, and implement it, before we start producing binaries (which will likely only be at the first beta release: #182).

str4d:
We ended up managing to do this for the first alpha.

PR #282, opened 2025-09-17
Generate the Debian copyright file including dependency licenses

str4d:
Depends on https://github.com/str4d/rage/pull/584.

Part of zcash/wallet#267.


zallet/build.rs
```diff
@@ -352,3 +356,74 @@ pub(super) static METHODS: ::phf::Map<&str, RpcMethod> = ::phf::phf_map! {
...
+    let mut contents = "Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: zallet
+
+Files:
+ *
+Copyright: 2024-2025, The Electric Coin Company
+License: MIT OR Apache-2.0"
+        .to_string();
+
+    let licensing = embed_licensing::collect(embed_licensing::CollectConfig::default())?;
```
str4d, line 370: The generation does slow down builds a bit, I suspect due to this live-fetching everything each time something changes that re-runs the build script. But we can resolve that with targeted caching later (i.e. ensuring that this only runs if dependencies change).

str4d: Actually, we only need this in release builds, so I'll disable it for debug builds.

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/bd70e5d820e2c50e82ad1320b212e1ee862f4619..c5a4a00de6bbd8c609d2719d49cdc74417d4d6d0) to fix clippy lints and disable generation in debug builds.

nuttycom:
re-utACK c5a4a00de6bbd8c609d2719d49cdc74417d4d6d0

PR #309, opened 2025-10-29
Update and audit dependencies

str4d:
Making incremental progress improving our audit base. This does not update everything that can be at present; I'll wait until we pull in the next Zcash crate releases for that.


nuttycom (approved):
utACK 616500372d587c9ae91ec233fa0dfd35c80066e3

PR #311, opened 2025-10-29
CI updates

str4d:


nullcopy (approved):
> utACK; I think that the expansion to check --all-features is desirable.

Agreed. Just wanted to check it was intentional.

utACK c5f7a6f45b4af1acc1e6050c2f3851e0b5b9fbb1

PR #335, opened 2025-11-12
Prepare for refactoring the sync engine

str4d:
Includes a bugfix to prevent the `recover_history` task from stomping over the chain region that `steady_state` is responsible for.


schell (approved):
LGTM. I would name these changes "Just call it Chain, already."

PR #338, opened 2025-11-17
SLSA-compliant release pipeline with end-to-end verification guide

### LINKED ISSUE #351: Auditable releases
After #313, we can produce release binaries that can be reproduced by anyone. This removes most of the trust in our release pipeline because given both a binary and the source repo, it is in theory possible to reproduce the binary. However, there is also value in proving that we intended to release a given binary, and that we did so at a given point in time, as well as attesting to the state of the build environment. We should update the release pipeline to produce whatever evidence is easy to do.

y4ssi:
- Introduces a tag-triggered release orchestrator that builds the StageX-based multi-arch image, pushes to Docker Hub, signs and attests via Cosign/Sigstore, exports the runtime bundle, and then feeds a new workflow that packages
    binaries plus Debian artifacts, runs smoke tests, signs with the Zcash KMS key, emits SPDX SBOMs, and attaches provenance/signatures to the GitHub Release and apt repository.
  - Replaces the old deb-release job with a reusable binaries-and-deb-release workflow that consumes the exported OCI bundle, supports customizable platform lists, and outputs GPG-signed binaries, .deb packages, and their intoto.jsonl
    attestations.
  - Refactors the Docker image workflow into a reusable workflow_call job that accepts tags/platforms, exports deterministic artifacts, generates SBOMs and provenance per architecture, and exposes metadata needed by downstream jobs.
  - Adds a comprehensive “Supply Chain Security (SLSA)” chapter detailing the workflow architecture, the guarantees (builder identity, provenance predicate, reproducibility), and step-by-step commands to verify images, binaries,
    packages, attestations, and SBOMs with cosign/rekor/slsa-verifier/GPG.
  - Pins critical GitHub Action versions (checkout v5 with commit SHAs, Rust toolchain action) across CI/lint/book/audit workflows to harden provenance throughout the pipeline.

Closes #351.

### DIFF
--- a/.github/workflows/build-and-push-docker-hub.yaml
+++ b/.github/workflows/build-and-push-docker-hub.yaml
@@ -33,7 +33,7 @@ on:
         required: true
         type: string
       dockerhub-namespace:
-        description: "Docker Hub namespace/organization (e.g. electriccoinco)"
+        description: "Docker Hub namespace/organization (e.g. zodlinc)"
         required: true
         type: string
     secrets:


.github/workflows/lints.yml
```diff
@@ -30,18 +30,18 @@ jobs:
     runs-on: ubuntu-latest
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
-      - uses: dtolnay/rust-toolchain@beta
+      - uses: dtolnay/rust-toolchain@4a46ea8263312b0fbcb4c73539e4b66ef8d8474f # beta
```
str4d: Same bug. Remove this change from your commit.

.github/workflows/build-and-push-docker-hub.yaml
```diff
@@ -1,52 +1,460 @@
...
+        required: true
+        type: string
+      image_tags:
+        description: "Comma-separated list of tags"
+        required: true
+        type: string
+      dockerfile:
+        description: "Path to the Dockerfile"
+        required: true
+        type: string
```
str4d: Why do we need this argument? Shouldn't we always be using the Dockerfile from this repository?
```suggestion
```

.github/workflows/build-and-push-docker-hub.yaml
```diff
@@ -1,52 +1,460 @@
...
+      context:
+        description: "Build context"
+        required: true
+        type: string
```
str4d: Same question.
```suggestion
```

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -0,0 +1,522 @@
...
+        required: true
+        type: string
+      oci-artifact-name:
+        description: "Name of the exported binaries artifact to consume"
+        required: true
+        type: string
+      platforms:
+        description: "Comma- or newline-separated list of linux/ARCH platforms to package"
+        required: false
+        default: linux/amd64
```
str4d: Ditto:
```suggestion
        required: true
```

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -0,0 +1,522 @@
...
+        type: string
+      smoke-distro:
+        description: "Debian release used for binary/package smoke tests"
+        required: false
+        default: bullseye
+        type: string
+      apt-distributions:
+        description: "Comma- or newline-separated Debian distributions to mirror/publish to apt.z.cash"
+        required: false
+        default: bullseye,bookworm,trixie
```
str4d: Ditto (and here the default doesn't match `release.yml`, it includes `trixie` which is listed there as a TODO implying it doesn't work yet):
```suggestion
        required: true
```

.github/workflows/build-and-push-docker-hub.yaml
```diff
@@ -1,52 +1,460 @@
...
-    outputs:
-      tags: ${{ steps.version_step.outputs.tags }}
     env:
-      VERSION: ${{ github.ref_name }}
-      SHA: ${{ github.sha }}
+      REGISTRY_HOST: docker.io
+      IMAGE_NAMESPACE: ${{ secrets.dockerhub_registry }}
+      IMAGE_NAME: ${{ inputs.image_name }}
+      BUILD_ARGS_RAW: ${{ inputs['build-args'] }}
+      OCI_ARTIFACT_NAME: ${{ inputs.oci-artifact-name }}
```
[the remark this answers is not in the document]
y4ssi: In the case of `build-args`, it’s set this way to match the downstream action workflow (`docker/build-push-action`)

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -0,0 +1,522 @@
...
+        uses: actions/attest-build-provenance@977bb373ede98d70efdf65b84cb5f73e068dcc2a # v3.0.0
+        with:
+          subject-path: standalone/zallet-${{ env.RELEASE_VERSION }}-${{ matrix.binary_suffix }}
+
+      - name: Save binary attestation bundle
+        env:
+          BUNDLE_PATH: ${{ steps.binary-attest.outputs['bundle-path'] }}
+        run: |
+          cp -- "$BUNDLE_PATH" "standalone/zallet-${RELEASE_VERSION}-${{ matrix.binary_suffix }}.intoto.jsonl"
+
```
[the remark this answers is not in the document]
y4ssi: You're right. The current attestation for the standalone binary only captures that this workflow extracted and packaged it, not the actual build environment (which is in the Docker image with provenance: mode=max). There's also no explicit link between the binary and the source OCI artifact digest.

I'll add the Docker image digest reference to make the provenance chain traceable.

.github/workflows/build-and-push-docker-hub.yaml
```diff
@@ -1,52 +1,341 @@
...
+          push: true
+          platforms: ${{ steps.platforms.outputs.platforms }}
+          tags: ${{ env.tags_fixed }}
+          sbom: true
+          provenance: mode=max
+          outputs: |
+            type=registry,rewrite-timestamp=true,force-compression=true
+
+      - name: Build export stage (local artifacts)
+        id: build_export
```
str4d: Does this reuse the artifacts from the previous step? If so, rename this for clarity (and remove the `id` which is unused below):
```suggestion
      - name: Extract artifacts from export stage
```

If it doesn't, why not?

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -0,0 +1,618 @@
...
+        type: string
+      smoke-distro:
+        description: "Debian release used for binary/package smoke tests"
+        required: true
+        default: bullseye
+        type: string
+      apt-distributions:
+        description: "Comma- or newline-separated Debian distributions to mirror/publish to apt.z.cash"
+        required: true
+        default: bullseye,bookworm
```
str4d: Ditto:
```suggestion
```

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -0,0 +1,618 @@
...
+
+
+      - name: Build package (cargo deb --no-build)
+        run: |
+          cargo deb --no-build --target ${{ matrix.target }}
+
+      - name: Rename package to not collide across the matrix
+        run: |
+          mv target/debian/zallet_*.deb ./
+          mv zallet_*.deb "zallet_${RELEASE_VERSION}_${{ matrix.arch }}.deb"
```
str4d: Bug: this re-introduces an earlier bug we had with the old `deb-release.yml` (fixed in 7ad338b3a1d9bd2d7d6cde70d9f1d56c453ec8ca) in that it does not include `${{ matrix.debian }}`, so builds for different Debian distros but the same arch will collide.

Either use the previous Bash logic we had in `deb-release.yml`:
```suggestion
          mv zallet_*.deb "$(basename -s .deb zallet_*.deb)_${{ matrix.debian }}.deb"
```
or explicitly include the Debian version:
```suggestion
          mv zallet_*.deb "zallet_${RELEASE_VERSION}_${{ matrix.arch }}_${{ matrix.debian }}.deb"
```

In either case, every reference to the .deb file in subsequent steps needs to be fixed to account for this extra suffix (like we also did in `deb-release.yml`).

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -0,0 +1,618 @@
...
+      - name: Upload artifacts for publish job
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        with:
+          name: release-${{ matrix.platform_key }}
+          path: release/${{ matrix.platform_key }}/*
+
+  publish:
+    name: Create/Update GitHub Release (aggregate)
+    needs: binaries_deb_release
+    if: ${{ github.ref_type == 'tag' }}
```
str4d: Does this work now that this job is triggered by `workflow_call`? (In `deb-release` it was triggered by the tag directly, but that is now handled by `release.yml`.)

nuttycom:
@y4ssi Is this ready for re-review? Please re-request review from @str4d when it's ready.

PR #357, opened 2026-01-30
rpc: Implement verifymessage (Issue #92)

### LINKED ISSUE #92: rpc: Implement `verifymessage`

emersonian:
This is my first PR to zallet, I am very open to constructive feedback and appreciate your review. I tried to keep this as close as possible to the zcashd implementation.

An open question for me is whether any of the signature verification logic should be upstreamed into librustzcash. To me it feels like core Zcash functionality to standardize for other developers interacting with transparent addresses. But implementing it here in a wallet first seems like the quickest path to keep deprecation moving along.

Resolves #92
Resolves COR-266

### DIFF
--- a/zallet/src/components/json_rpc/methods/verify_message.rs
+++ b/zallet/src/components/json_rpc/methods/verify_message.rs
@@ -8,10 +8,10 @@ use secp256k1::{
 };
 use serde::Serialize;
 use std::io::Write;
-use zcash_primitives::transaction::util::sha256d::HashWriter;
 use transparent::address::TransparentAddress;
 use zcash_encoding::CompactSize;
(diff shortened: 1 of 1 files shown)


nullcopy (changes requested):
Thanks for this! I've reviewed your code and compared it to the original cpp code, and I only have a few small suggestions.

In response to your question:
> An open question for me is whether any of the signature verification logic should be upstreamed into librustzcash

I lean towards yes, but IMO that can happen in a second PR. This is a good addition in the meantime. I'll let @nuttycom or @str4d chime in on that point, though.

emersonian:
Sure no problem. Thanks for the quick reviews you rock!

Issue #101, opened 2025-04-08
rpc: Implement `z_getbalances`


str4d:
This is the RPC proposed in https://github.com/zcash/zcash/issues/6356 and drafted in https://github.com/zcash/zcash/pull/6389 that was never completed in `zcashd`. We should ensure it exists from the beginning in Zallet.

PR #362, opened 2026-02-03
rpc: Add `z_getbalances`

str4d:
Closes zcash/wallet#101.
Closes COR-270.

### DIFF
--- a/zallet/src/components/json_rpc/methods/get_balances.rs
+++ b/zallet/src/components/json_rpc/methods/get_balances.rs
@@ -29,6 +29,16 @@ pub(crate) struct Balances {
     /// or no legacy transparent funds are present.
     #[serde(skip_serializing_if = "Option::is_none")]
     legacy_transparent: Option<TransparentBalance>,
+
+    /// The balance of transparent funds held by legacy watch-only transparent addresses.
+    ///
+    /// All funds held in legacy transparent addresses are treated as though they are
+    /// associated with a single spending authority.
+    ///
+    /// Omitted if `features.legacy_pool_seed_fingerprint` is unset in the Zallet config,
+    /// or no legacy transparent funds are present.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    legacy_transparent_watchonly: Option<TransparentBalance>,
 }
 
 #[derive(Clone, Debug, Serialize, JsonSchema)]
@@ -38,9 +48,20 @@ struct AccountBalance {
 
     /// The balance held by the account in the transparent pool.
     ///
+    /// This includes all funds held by transparent addresses derived from the account's
+    /// viewing key, and excludes all funds held by watch-only standalone transparent
+    /// addresses imported into the account.
+    ///
     /// Omitted if no transparent funds are present.
     #[serde(skip_serializing_if = "Option::is_none")]
     transparent: Option<TransparentBalance>,
+
+    /// The balance held by each watch-only standalone transparent address imported into
+    /// the account.
+    ///
+    /// Omitted if the account has no standalone transparent addresses.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    transparent_watchonly: Vec<StandaloneTransparentAddressBalance>,
 
     /// The balance held by the account in the Sapling shielded pool.
     ///
@@ -56,6 +77,15 @@ struct AccountBalance {
 
     /// The total funds in all pools held by the account.
     total: Balance,
+}
+
+#[derive(Clone, Debug, Serialize, JsonSchema)]
+struct StandaloneTransparentAddressBalance {
+    /// The standalone transparent address.
+    address: String,
+
+    #[serde(flatten)]
+    balance: TransparentBalance,
 }
 
 #[derive(Clone, Debug, Serialize, JsonSchema)]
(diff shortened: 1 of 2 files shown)


zallet/src/components/json_rpc/methods/get_balances.rs
```diff
@@ -0,0 +1,223 @@
...
+    ///
+    /// All funds held in legacy transparent addresses are treated as though they are
+    /// associated with a single spending authority.
+    ///
+    /// Omitted if `features.legacy_pool_seed_fingerprint` is unset in the Zallet config,
+    /// or no legacy transparent funds are present.
+    legacy_transparent: Option<TransparentBalance>,
+
+    /// The total of all funds for which this wallet controls spending keys.
+    total: Balance,
```
str4d: A downside to the current design including `total` fields for each account as well as the wallet balance, is that it is very duplicative:
```json
$ zallet -d local/wallet-testnet rpc z_getbalances 1 true
{
  "accounts": [
    {
      "account_uuid": "46416286-4799-4ac4-be4b-dd879274d340",
      "sapling": {
        "value": 4.69997000,
        "valueZat": 469997000
      },
      "total": {
        "value": 4.69997000,
        "valueZat": 469997000
      }
    }
  ],
  "total": {
    "value": 4.69997000,
    "valueZat": 469997000
  }
}
```

But maybe this is only visually annoying, and doesn't actually matter? From an automated consumption perspective, this mostly just doubles the JSON-RPC response size, which is linear in the number of accounts in the wallet, so likely not large.

daira: It's fine I think. But personally, I would just drop "value" and only include "valueZat". Decimals were always a mistake.

str4d: With the above changes, an offline-for-some-time syncing wallet looks like this:
```json
$ zallet -d local/wallet-testnet rpc z_getbalances 1 true
{
  "accounts": [
    {
      "account_uuid": "46416286-4799-4ac4-be4b-dd879274d340",
      "sapling": {
        "spendable": {
          "valueZat": 0
        },
        "pending": {
          "valueZat": 469997000
        }
      },
      "orchard": {
        "spendable": {
          "valueZat": 0
        },
        "pending": {
          "valueZat": 960000000
        }
      },
      "total": {
        "spendable": {
          "valueZat": 0
        },
        "pending": {
          "valueZat": 1429997000
        }
      }
    }
  ]
}
```

And once synced, we see this:
```json
$ zallet -d local/wallet-testnet rpc z_getbalances 1 true
{
  "accounts": [
    {
      "account_uuid": "46416286-4799-4ac4-be4b-dd879274d340",
      "sapling": {
        "spendable": {
          "valueZat": 469997000
        }
      },
      "orchard": {
        "spendable": {
          "valueZat": 960000000
        }
      },
      "total": {
        "spendable": {
          "valueZat": 1429997000
        }
      }
    }
  ]
}
```

So with these changes, the only guaranteed-to-show balance field is `['accounts'][N]['total']['spendable']['valueZat']`, and then only if `['accounts']` is non-empty (which it will be in almost every case).

zallet/src/components/json_rpc/methods/get_balances.rs
```diff
@@ -0,0 +1,275 @@
...
+    sapling: Option<Balance>,
+
+    /// The balance held by the account in the Orchard shielded pool.
+    ///
+    /// Omitted if no Orchard funds are present.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    orchard: Option<Balance>,
+
+    /// The total funds in all pools held by the account.
+    total: Balance,
```
nuttycom:
```suggestion
```

str4d:
[Force-pushed](https://github.com/zcash/wallet/compare/6e6900bb83ef278ef2c682bca8adb00ce72bc3bd..024a1774abf55e869e5abba0359005d3a4a4dd6d) to address review comments.

zallet/src/components/json_rpc/methods/get_balances.rs
```diff
@@ -0,0 +1,275 @@
...
+
+pub(crate) fn call(
+    wallet: &DbConnection,
+    minconf: Option<u32>,
+    include_watchonly: Option<bool>,
+) -> Response {
+    match include_watchonly {
+        Some(true) => Ok(()),
+        None | Some(false) => Err(LegacyCode::Misc
+            .with_message("include_watchonly argument must be set to true (for now)")),
```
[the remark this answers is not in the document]
nuttycom: I like the idea of splitting `legacy_transparent` into two parts. When I think more about importing watch-only addresses, I wonder if a better way to go about it would be to require that an account be specified when importing an address. I think that might be more useful than treating imported addresses as part of an undifferentiated watch-only pool of funds; the caller can always do sums, and this gives the user of the API the choice to either plumb them all into a single bucket, or to split them up however they like.

Here's a concrete use case: a user with a transparent-only Ledger device could choose to import all of the distinct transparent addresses produced by that device into a single logical account (because, indeed, that's how they're treated for spending purposes.) In the case of legacy imported addresses, we could import them all into the single legacy account. 

We also need to make sure we have the ability to *remove* an imported watchonly address.

zallet/src/components/json_rpc/methods/get_balances.rs
```diff
@@ -0,0 +1,233 @@
...
+    }?;
+
+    let confirmations_policy = match minconf {
+        Some(minconf) => match NonZeroU32::new(minconf) {
+            Some(c) => ConfirmationsPolicy::new_symmetrical(c, false),
+            None => ConfirmationsPolicy::new_symmetrical(NonZeroU32::MIN, true),
```
nuttycom: nit: document the rationale here (user provided an explicit zero, so we are defaulting to "minimum possible confirmations.")

str4d: Yep, this is a limitation of how `ConfirmationsPolicy` was designed, and something I recall saying at the time would need to be changed to enable properly handling fully-transparent logic. Once that work is done, this would change to actually use `minconf 0` for determining fully-transparent spends (while "zero-conf" shielded funds would always show as `pending` because they can never be spent as such).

zallet/src/components/json_rpc/methods/get_balances.rs
```diff
@@ -0,0 +1,233 @@
...
+                orchard: opt_balance_from(account.orchard_balance())?,
+                total: balance_from(account)?,
+            })
+        })
+        .collect::<RpcResult<_>>()?;
+
+    Ok(Balances {
+        accounts,
+        // TODO: Fetch legacy transparent balance once supported.
+        legacy_transparent: None,
```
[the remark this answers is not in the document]
str4d: Opened #384

zallet/src/components/json_rpc/methods/get_balances.rs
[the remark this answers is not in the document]
str4d: I realised that if we still allow importing standalone transparent addresses, then even with legacy mode disabled, we need to distinguish watch-only transparent balances. So I'm also adding a `transparent_watchonly` to the accounts balance. Should this just be a sum over all the imported transparent addresses within the account (kind of like the legacy balance was for the entire wallet), or should this instead show the balance per transparent address? I'm doing the latter for now.

PR #364, opened 2026-02-03
rpc: Implement asm fields of getrawtransaction (#235)

### LINKED ISSUE #235: rpc: Implement `asm` fields of `getrawtransaction`

emersonian:
Resolves #235

### DIFF
--- a/zallet/src/components/json_rpc/methods/get_raw_transaction.rs
+++ b/zallet/src/components/json_rpc/methods/get_raw_transaction.rs
@@ -12,6 +12,7 @@ use zcash_protocol::{
     consensus::{self, BlockHeight},
     value::ZatBalance,
 };
+use zcash_script::script::{Asm, Code};
 
 use crate::components::{
     database::DbConnection,
@@ -648,7 +649,8 @@ pub(crate) async fn call(
 
 impl TransparentInput {
     fn encode(tx_in: &TxIn<transparent::bundle::Authorized>, is_coinbase: bool) -> Self {
-        let script_hex = hex::encode(&tx_in.script_sig().0.0);
+        let script_bytes = &tx_in.script_sig().0.0;
+        let script_hex = hex::encode(script_bytes);
 
         if is_coinbase {
             Self {
@@ -659,14 +661,15 @@ impl TransparentInput {
                 sequence: tx_in.sequence(),
             }
         } else {
+            // For scriptSig, we pass `true` since there may be signatures
+            let asm = to_zcashd_asm(&Code(script_bytes.to_vec()).to_asm(true));
+
             Self {
                 coinbase: None,
                 txid: Some(TxId::from_bytes(*tx_in.prevout().hash()).to_string()),
                 vout: Some(tx_in.prevout().n()),
                 script_sig: Some(TransparentScriptSig {
-                    // TODO: Implement this
-                    //       https://github.com/zcash/wallet/issues/235
-                    asm: "TODO: Implement this".into(),
+                    asm,
                     hex: script_hex,
                 }),
                 sequence: tx_in.sequence(),
@@ -677,16 +680,19 @@ impl TransparentInput {
 
 impl TransparentOutput {
     fn encode((tx_out, n): (&TxOut, u16)) -> Self {
+        let script_bytes = &tx_out.script_pubkey().0.0;
+
+        // For scriptPubKey, we pass `false` since there are no signatures
+        let asm = to_zcashd_asm(&Code(script_bytes.to_vec()).to_asm(false));
+
+        // Detect the script type using zcash_script's solver.
+        let (kind, req_sigs) = detect_script_type_and_sigs(script_bytes);
+
         let script_pub_key = TransparentScriptPubKey {
-            // TODO: Implement this
-            //       https://github.com/zcash/wallet/issues/235
-            asm: "TODO: Implement this".into(),
-            hex: hex::encode(&tx_out.script_pubkey().0.0),
-            // TODO: zcashd relied on initialization behaviour for the default value
-            //       for null-data or non-standard outputs. Figure out what it is.
-            //       https://github.com/zcash/wallet/issues/236
-            req_sigs: 0,
-            kind: "nonstandard",
+            asm,
+            hex: hex::encode(script_bytes),
+            req_sigs,
+            kind,
             addresses: vec![],
         };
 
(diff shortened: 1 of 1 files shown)


zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -648,7 +649,8 @@ pub(crate) async fn call(
 
 impl TransparentInput {
     fn encode(tx_in: &TxIn<transparent::bundle::Authorized>, is_coinbase: bool) -> Self {
-        let script_hex = hex::encode(&tx_in.script_sig().0.0);
+        let script_bytes = &tx_in.script_sig().0.0;
```
oxarbitrage: Optional:
```suggestion
        let script_bytes = &tx_in.script_sig().0.to_bytes();
```

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +791,245 @@ impl OrchardAction {
         }
     }
 }
+
+/// Converts zcash_script asm output to zcashd-compatible format.
+///
+/// The zcash_script crate outputs "OP_1" through "OP_16" and "OP_1NEGATE",
+/// but zcashd outputs "1" through "16" and "-1" respectively.
```
oxarbitrage: We need a link to the zcashd code for this.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +791,245 @@ impl OrchardAction {
...
+fn to_zcashd_asm(asm: &str) -> String {
```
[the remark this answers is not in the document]
emersonian: I think consistency with bitcoind's script representation is nice so that its wide ecosystem's tools are easy to point at Zcash, so I would keep the to_zcashd_asm around here. I'm not sure if zcashd backwards compatibility is a goal of the asm methods in Zebra, will defer to your team on that.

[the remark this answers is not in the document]
conradoplg: The goal was zcash_script to produce the same asm as zcashd does, so I agree this should upstream this fix. I created an issue for this in zcash_script https://github.com/ZcashFoundation/zcash_script/issues/288

That being said this is not a blocker since it will keep working when we change zcash_script.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -659,14 +661,15 @@ impl TransparentInput {
                 sequence: tx_in.sequence(),
             }
         } else {
+            // For scriptSig, we pass `true` since there may be signatures
+            let asm = to_zcashd_asm(&Code(script_bytes.to_vec()).to_asm(true));
```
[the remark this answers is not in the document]
emersonian: I'm uncertain about that, I think that switching to false for scriptSig would drop sighash type decoding from the asm output. Signatures would show as raw hex instead of having [ALL], [NONE|ANYONECANPAY] suffixes, which is what zcashd produces (it passes true for scriptSig at [src/core_write.cpp#L166](https://github.com/zcash/zcash/blob/v6.11.0/src/core_write.cpp#L166)).

If we want to intentionally diverge from zcashd here I'd like for the maintainers to discuss and make an explicit decision. I'll leave this PR matching zcashd for now but am happy to update if we decide to drop sighash decoding.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
emersonian: I'm going to avoid using to_bytes for now to prevent an additional allocation but if I'm over-optimizing let me know and I'm happy to refactor as requested.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +791,245 @@ impl OrchardAction {
...
+
+/// Converts zcash_script asm output to zcashd-compatible format.
+///
+/// The zcash_script crate outputs "OP_1" through "OP_16" and "OP_1NEGATE",
+/// but zcashd outputs "1" through "16" and "-1" respectively.
+fn to_zcashd_asm(asm: &str) -> String {
+    asm.split(' ')
+        .map(|token| match token {
+            "OP_1NEGATE" => "-1",
+            "OP_1" => "1",
```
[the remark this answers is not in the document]
emersonian: Good catch, resolved.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
oxarbitrage: Fair enough.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +791,245 @@ impl OrchardAction {
...
+        .and_then(|component| zcash_script::solver::standard(&component))
+        .map(|script_kind| match script_kind {
+            zcash_script::solver::ScriptKind::PubKeyHash { .. } => ("pubkeyhash", 1),
+            zcash_script::solver::ScriptKind::ScriptHash { .. } => ("scripthash", 1),
+            zcash_script::solver::ScriptKind::MultiSig { required, .. } => ("multisig", required),
+            zcash_script::solver::ScriptKind::NullData { .. } => ("nulldata", 0),
+            zcash_script::solver::ScriptKind::PubKey { .. } => ("pubkey", 1),
+        })
+        .unwrap_or(("nonstandard", 0))
+}
```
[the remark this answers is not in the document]
str4d: I was going to say that this could be replaced by https://github.com/ZcashFoundation/zcash_script/pull/291, but due to that PR being based on `main` (and thus 0.5) instead of the 0.4 series (which is identical to 0.5 due to a SemVer mistake), and merged before I could review it and point this out, we cannot use it without someone doing additional backporting work.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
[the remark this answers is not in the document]
str4d: We should not diverge from `zcashd` here (in particular because this encoding was not defined by `zcashd`, but by Bitcoin Core).

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +791,256 @@ impl OrchardAction {
...
+///
+/// Reference: https://github.com/zcash/zcash/blob/v6.11.0/src/script/script.cpp#L19-L40
+///
+/// TODO: Remove this function once zcash_script is upgraded past 0.4.x,
+///       as `to_asm()` will natively output zcashd-compatible format.
+///       See https://github.com/ZcashFoundation/zcash_script/pull/289
+fn to_zcashd_asm(asm: &str) -> String {
+    asm.split(' ')
+        .map(|token| match token {
+            "OP_0" => "0",
```
str4d:
```suggestion
```

@oxarbitrage's suggestion was unnecessary, because @conradoplg had already special-cased `OP_0` in `zcash_script` when originally implementing the ASM renderer (just not the rest of the small-number opcodes).

Non-blocking because it should just be superfluous matching.

str4d (changes requested):
Reviewed 14832268faa4e25d79300dd3b9569fc1644b63da.

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +788,258 @@ impl OrchardAction {
...
+            ("82", "[NONE|ANYONECANPAY]"),
+            ("83", "[SINGLE|ANYONECANPAY]"),
+        ];
+
+        for (sighash_byte, expected_suffix) in test_cases {
+            let sig_hex = format!("48{}{}", base_sig, sighash_byte);
+            let sig_bytes = hex::decode(&sig_hex).unwrap();
+            let asm = Code(sig_bytes).to_asm(true);
+            assert!(
+                asm.contains(expected_suffix),
```
str4d: Since we know this is a suffix:
```suggestion
                asm.ends_with(expected_suffix),
```

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +788,258 @@ impl OrchardAction {
...
+        // A simple signature with SIGHASH_ALL (0x01) suffix
+        // This is a minimal DER signature followed by 0x01
+        let sig_with_sighash = hex::decode(
+            "483045022100ab4c5228e6f8290a5c7e1c4afedbb32b6a6e95b9f873d2e1d5f6a8c3b4e7f09102205d6a8c3b4e7f091ab4c5228e6f8290a5c7e1c4afedbb32b6a6e95b9f873d2e1d01"
+        ).unwrap();
+
+        // With sighash decode (for scriptSig), should show [ALL]
+        let asm_with_decode = Code(sig_with_sighash.clone()).to_asm(true);
+        assert!(
+            asm_with_decode.contains("[ALL]"),
```
str4d: Ditto:
```suggestion
            asm_with_decode.ends_with("[ALL]"),
```

zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -782,3 +788,258 @@ impl OrchardAction {
...
+            "scriptSig should decode sighash suffix"
+        );
+
+        // Without sighash decode (for scriptPubKey), should show raw hex
+        let asm_without_decode = Code(sig_with_sighash).to_asm(false);
+        assert!(
+            !asm_without_decode.contains("[ALL]"),
```
str4d: Here by contrast, since we're checking for absence, `contains` is fine.

str4d (approved):
utACK 027565af082a0dbe01ed5c975b9deffa45b53f41 with suggestions.

Issue #237, opened 2025-09-04
chain_view: Migrate to `ChainIndex` trait


str4d:
The `zaino-state` crate is gaining a new `ChainIndex` trait, and associated indexer flow, that better serves the needs of a full-node wallet like Zallet. Once it is finished, we will migrate to it.

str4d:
Current blockers as of zingolabs/zaino@da2717f44b040f346543ed96c3758ad9e876ae6e:
- No index of subtree roots in a way that is accessible via `ChainIndex`.

PR #367, opened 2026-02-04
rpc: Add `getwalletstatus`

### LINKED ISSUE #316: Add a `walletstatus` RPC method
The motivation for this proposal comes from testing needs, but I don’t think the solution should be limited to that use case.

We want to eliminate the (ab)use of sleep-based synchronization in the Zebra wallet tests (https://github.com/ZcashFoundation/zebra/pull/10045). The issue appears when Zebra mines a new block and there’s a delay before Zallet detects and processes it. To avoid guessing timing with sleeps, we need an RPC method that reports the wallet’s processed tip height, so the tests can actively poll until both components are in sync.

Zebra already provides `healthy` and `ready` HTTP endpoints (https://github.com/ZcashFoundation/zebra/blob/main/book/src/user/health.md), but this is different: we need to see the wallet’s actual best tip height, not just a binary "OK" once sync is complete. We need visibility into where the wallet currently is in the chain.

A dedicated RPC method seems reasonable for this. Existing calls such as `getwalletinfo` don’t expose the necessary state. That said, I’m open to discussion if there’s a preferred alternative.

Draft proposal (open for changes):

```rust
pub(crate) struct WalletStatus {
    /// The chain tip height the wallet has observed.
    observed_tip: u32,
    /// The most recent block height fully processed by the wallet.
    processed_height: u32,
    /// A timestamp for the wallet's status (e.g., time of last update).
    timestamp: u64,
}
```

str4d:
This is intended for use with https://github.com/zcash/integration-tests to enable monitoring the sync status of Zallet processes in regtest mode.

ZIT-Revision: 5f84b1fa54ac9b1ebfc64ec602ac214edc884d01
Closes zcash/wallet#316.
Closes COR-338.

### DIFF
--- a/zallet/src/components/json_rpc/methods/get_wallet_status.rs
+++ b/zallet/src/components/json_rpc/methods/get_wallet_status.rs
@@ -1,0 +1,179 @@
+use documented::Documented;
+use jsonrpsee::core::RpcResult;
+use rand::rngs::OsRng;
+use schemars::JsonSchema;
+use serde::Serialize;
+use zaino_state::{FetchServiceSubscriber, LightWalletIndexer};
+use zcash_client_backend::data_api::{
+    BlockMetadata, WalletRead, WalletSummary, scanning::ScanRange, wallet::ConfirmationsPolicy,
+};
+use zcash_client_sqlite::{AccountUuid, WalletDb, error::SqliteClientError, util::SystemClock};
+use zcash_primitives::block::BlockHash;
+
+use crate::{
+    components::{database::DbConnection, json_rpc::server::LegacyCode},
+    network::Network,
+};
+
+/// Response to a `getwalletstatus` RPC request.
+pub(crate) type Response = RpcResult<ResultType>;
+pub(crate) type ResultType = GetWalletStatus;
+
+/// The wallet status information.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct GetWalletStatus {
+    /// The backing full node's view of the chain tip.
+    node_tip: ChainTip,
+
+    /// The wallet's view of the chain tip.
+    ///
+    /// This should only diverge from `node_tip` for very short periods of time.
+    ///
+    /// Omitted if the wallet has just been started for the first time and has not yet
+    /// begun syncing.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    wallet_tip: Option<ChainTip>,
+
+    /// The height to which the wallet is fully synced.
+    ///
+    /// The wallet only has a partial view of chain data above this height.
+    ///
+    /// Omitted if the wallet does not have any accounts or birthday data and thus has
+    /// nowhere to sync from, or if the wallet birthday itself has not yet been synced.
+    /// The latter occurs when a recovered wallet first starts and is scanning the chain
+    /// tip region.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fully_synced_height: Option<u32>,
+
+    /// The height up to which the wallet has full information.
+    ///
+    /// Omitted if the wallet is fully synced (which requires that `fully_synced_height`
+    /// is equal to `wallet_tip.height`).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    sync_work_remaining: Option<SyncWorkRemaining>,
+}
+
+#[derive(Clone, Debug, Serialize, JsonSchema)]
+struct ChainTip {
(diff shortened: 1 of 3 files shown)


zallet/src/components/json_rpc/methods/get_wallet_status.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+#[derive(Clone, Debug, Serialize, JsonSchema)]
+struct SyncWorkRemaining {
+    unscanned_blocks: u32,
+
+    // TODO: Replace these with accurate unscanned note counts, which we can determine
+    // because Zallet tracks the chain tip very closely.
+    progress_numerator: u64,
+    progress_denominator: u64,
```
str4d: These are temporary because it's all I have access to currently. Given that Zallet is intended to track the backing node's chain tip as closely as possible, once #237 is done we will always know the size of the note commitment trees, and thus it should always be possible to determine how many leaves within them have not yet been appended (and thus are unscanned).

zallet/src/components/json_rpc/methods/get_wallet_status.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// The height to which the wallet is fully synced.
+    ///
+    /// The wallet only has a partial view of chain data above this height.
+    ///
+    /// Omitted if the wallet does not have any accounts or birthday data and thus has
+    /// nowhere to sync from, or if the wallet birthday itself has not yet been synced.
+    /// The latter occurs when a recovered wallet first starts and is scanning the chain
+    /// tip region.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fully_synced_height: Option<u32>,
```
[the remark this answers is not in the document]
str4d: Having it be present-but-null vs not-present-if-null is a different question than the one I was asking.

My question was, should `None` be an allowable value here, or should we instead return something else?

oxarbitrage:
Pushed a commit addressing the open review threads, and rebased the branch on `main` (it was ~2 months behind):

- **Progress fields** — grouped `progress_numerator`/`progress_denominator` into a `Progress { numerator, denominator }` struct, per the suggestion above (serializes as `{"numerator": N, "denominator": D}`).
- **Approximate progress** — kept the fields (so the QA framework has a usable signal now) but documented `progress` as an approximation derived from scanned block ranges, noting it will become an exact unscanned-note count once #237 lands.
- **`fully_synced_height`** — keeping `Option`/`None` as discussed; the field doc already describes when it is omitted.
- Added the missing `CHANGELOG` entry for `getwalletstatus`.

`cargo fmt` and `cargo clippy --all-features --all-targets` are clean on the rebased branch.

@str4d since this touches your PR — please take a look. As I have now contributed code here, it would be good to have a separate reviewer before merge.

zallet/src/components/json_rpc/methods/get_wallet_status.rs
oxarbitrage: Resolving for now: kept the fields so clients/QA have a usable progress signal, and documented `progress` as an approximation that will be refined to exact unscanned-note counts once #237 lands. Easy to revisit then.

dannywillems:
Rebased on top of master.

zallet/src/components/json_rpc/methods/get_wallet_status.rs
```diff
@@ -0,0 +1,192 @@
...
+#[derive(Clone, Debug, Serialize, JsonSchema)]
+struct SyncWorkRemaining {
+    /// The number of blocks within the wallet's view of the chain that have not yet been
+    /// scanned.
+    unscanned_blocks: u32,
+
+    /// Approximate sync progress, as a `numerator / denominator` fraction.
+    ///
+    /// This is currently derived from scanned block ranges. Once the wallet always tracks
+    /// the note commitment tree sizes (zcash/wallet#237), this will be refined to an exact
```
dannywillems, line 74: > Once the wallet always tracks the note commitment tree sizes (zcash/wallet#237), this will be refined to an exact count of the unscanned note commitments.

This looks to me an additional good UX. I am opening a follow-up issue to keep track of this. Maybe we should have both (block + NC tree sizes).

dannywillems: See https://github.com/zcash/wallet/issues/469

dannywillems (approved):
LGTM on the idea.
Checking the counterparty on integration-tests.

nullcopy (approved):
utACK 7404cbf5b044500f4f5a9233e855c56a137f6feb

PR #368, opened 2026-02-04
rpc: Implement decodescript (#94)

### LINKED ISSUE #94: rpc: Implement `decodescript`

emersonian:
Resolves #94
Resolves COR-268

to_zcashd_asm is duplicated from #235, we can clean that up once either is merged, or adjust if maintainers want to take a different approach.

### DIFF
--- a/zallet/src/components/json_rpc/methods/decode_script.rs
+++ b/zallet/src/components/json_rpc/methods/decode_script.rs
@@ -1,0 +1,331 @@
+use documented::Documented;
+use jsonrpsee::core::RpcResult;
+use ripemd::Ripemd160;
+use schemars::JsonSchema;
+use secp256k1::PublicKey;
+use serde::Serialize;
+use sha2::{Digest, Sha256};
+use transparent::address::TransparentAddress;
+use zcash_keys::encoding::AddressCodec;
+use zcash_script::{
+    script::{Asm, Code},
+    solver::{self, ScriptKind},
+};
+
+use crate::{components::json_rpc::server::LegacyCode, network::Network};
+
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// The result of decoding a script.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct ResultType {
+    /// String representation of the script public key.
+    asm: String,
+    /// The type of script.
+    ///
+    /// One of: `pubkeyhash`, `scripthash`, `pubkey`, `multisig`, `nulldata`, `nonstandard`.
+    #[serde(rename = "type")]
+    kind: &'static str,
+    /// The required number of signatures.
+    ///
+    /// Omitted for scripts that don't contain identifiable addresses (such as
+    /// non-standard or null-data scripts).
+    #[serde(rename = "reqSigs", skip_serializing_if = "Option::is_none")]
+    req_sigs: Option<u8>,
+    /// The addresses associated with this script.
+    ///
+    /// Omitted for scripts that don't contain identifiable addresses (such as
+    /// non-standard or null-data scripts).
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    addresses: Vec<String>,
+    /// The P2SH address for this script.
+    p2sh: String,
+}
+
+pub(super) const PARAM_HEXSTRING_DESC: &str = "The hex-encoded script.";
+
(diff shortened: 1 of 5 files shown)


zallet/src/components/json_rpc/methods/decode_script.rs
```diff
@@ -0,0 +1,320 @@
...
+fn calculate_p2sh_address(script_bytes: &[u8], params: &Network) -> String {
+    let hash = hash160(script_bytes);
+    TransparentAddress::ScriptHash(hash).encode(params)
+}
+
+/// Detects the script type and extracts associated information.
+///
+/// Returns a tuple of (type_name, required_sigs, addresses).
+fn detect_script_info(
+    script_bytes: &[u8],
```
str4d:
```suggestion
    script_code: &Code,
```

zallet/src/components/json_rpc/methods/decode_script.rs
```diff
@@ -0,0 +1,320 @@
...
+    #[serde(rename = "type")]
+    kind: &'static str,
+    /// The required number of signatures.
+    ///
+    /// Omitted for scripts that don't require signatures.
+    #[serde(rename = "reqSigs", skip_serializing_if = "Option::is_none")]
+    req_sigs: Option<u8>,
+    /// The addresses associated with this script.
+    ///
+    /// Empty for non-standard scripts and nulldata scripts.
```
str4d:
```suggestion
    /// Omitted for non-standard scripts and nulldata scripts.
```

zallet/src/components/json_rpc/methods/decode_script.rs
```diff
@@ -0,0 +1,320 @@
...
+             53ae",
+            pk = ZCASHD_PUBLIC_KEY
+        );
+        let result = call(&mainnet(), &script_hex).unwrap();
+
+        assert_eq!(result.kind, "multisig");
+        assert_eq!(result.req_sigs, Some(2));
+        assert_eq!(result.addresses.len(), 3);
+        // All addresses should be the same since we used the same pubkey
+        assert!(result.addresses.iter().all(|a| a.starts_with("t1")));
```
str4d: This is not checking that all addresses are the same. Instead you could check that with something like:
```suggestion
        // All addresses should be the same since we used the same pubkey
        assert_eq!(result.addresses.iter().collect::<HashSet>().len(), 1);
```

zallet/src/components/json_rpc/methods/decode_script.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    #[test]
+    fn p2sh_address_calculation() {
+        // Verify P2SH address is correctly calculated for a P2PKH script
+        let script_hex = format!("76a914{ZCASHD_PUBLIC_KEY_HASH}88ac");
+        let result = call(&mainnet(), &script_hex).unwrap();
+
+        // P2SH address should be computed for any script type
+        assert!(result.p2sh.starts_with("t3"));
+    }
```
str4d: This test is not verifying that the P2SH address is correctly calculated for a P2PKH script; it is only checking that some P2SH address is returned.

str4d (changes requested):
Reviewed dc4e63891f0fcab9d458d7743a7232828824f800

zallet/src/components/json_rpc/methods/decode_script.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    #[test]
+    fn p2sh_address_calculation() {
+        // Verify P2SH address is correctly calculated for a P2PKH script
+        let script_hex = format!("76a914{ZCASHD_PUBLIC_KEY_HASH}88ac");
+        let result = call(&mainnet(), &script_hex).unwrap();
+
+        // P2SH address should be computed for any script type
+        let script_bytes = hex::decode(&script_hex).unwrap();
+        let expected_p2sh = calculate_p2sh_address(&script_bytes, &mainnet());
```
str4d: This test is now just a tautology, since it's calling the same computation code either way. This should instead be pinning the correct address, which was ideally computed via some other impl. I am not a fan of tautological tests because they make it look like there is good test coverage but don't actually catch bugs.

Non-blocking, we can fix this later.

str4d (approved):
utACK 3a3fa903c45d43891b4345e56e876e20ef4026e8 with test feedback.

PR #369, opened 2026-02-04
rpc: Implement decoderawtransaction (#93)

### LINKED ISSUE #93: rpc: Implement `decoderawtransaction`

emersonian:
Resolves #93

### DIFF
--- a/zallet/src/components/json_rpc/methods/get_raw_transaction.rs
+++ b/zallet/src/components/json_rpc/methods/get_raw_transaction.rs
@@ -184,7 +184,7 @@ pub(crate) struct Transaction {
 }
 
 #[derive(Clone, Debug, Serialize, JsonSchema)]
-struct TransparentInput {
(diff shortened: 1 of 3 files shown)


zallet/src/components/json_rpc/methods/get_raw_transaction.rs
```diff
@@ -37,7 +37,7 @@ pub(crate) enum ResultType {
 
 /// Verbose information about a transaction.
 #[derive(Clone, Debug, Serialize, JsonSchema)]
-pub(crate) struct Transaction {
+pub(super) struct Transaction {
```
str4d: Clippy is complaining that this is too private. And given that this PR doesn't actually use this type yet, there's no need to alter this.
```suggestion
pub(crate) struct Transaction {
```

str4d (approved):
Re-utACK a0ef77ca0bef5421c971a3649a8eb860c05d8767

PR #370, opened 2026-02-04
rpc: Add z_converttex RPC endpoint (#327)

### LINKED ISSUE #327: rpc: Add `z_converttex` RPC endpoint.
This is the [RPC](https://zcash.github.io/rpc/z_converttex.html) proposed in https://github.com/zcash/zcash/pull/6891. It was implemented to provide a "full-node backed" implementation of tex address conversion which was mainly required by `Binance` as a way to double-check their own implementation of this functionality. 

Several wallet developers have had doubts on how to handle TEX addresses despite this being explained with reference implementations on ZIP-320. Consider implementing this RPC into Zallet.

emersonian:
Resolves #327

### DIFF
--- a/zallet/i18n/en-US/zallet.ftl
+++ b/zallet/i18n/en-US/zallet.ftl
@@ -243,6 +243,11 @@ rec-privpol-privacy-weakening =
     to '{$policy}' or weaker if you wish to allow this transaction to proceed
     anyway.
 
+## RPC method errors
+
+err-rpc-convert-tex-invalid-address = Invalid address
+err-rpc-convert-tex-not-p2pkh = Address is not a transparent p2pkh address
+
 ## RPC CLI errors
 
 err-rpc-cli-conn-failed = Failed to connect to the Zallet wallet's JSON-RPC port.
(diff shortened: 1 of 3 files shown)


zallet/src/components/json_rpc/methods/convert_tex.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::{components::json_rpc::server::LegacyCode, network::Network};
+
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// The TEX address encoding of the input transparent P2PKH address.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+#[serde(transparent)]
+pub(crate) struct ResultType(String);
+
+pub(super) const PARAM_TRANSPARENT_ADDRESS_DESC: &str = "The transparent P2PKH address to convert.";
```
nuttycom: Consider internationalizing the English strings used in this file by adding them to https://github.com/zcash/wallet/blob/main/zallet/i18n/en-US/zallet.ftl and using the `fl!` macro, instead of a constant.

zallet/i18n/en-US/zallet.ftl
```diff
@@ -243,6 +243,11 @@ rec-privpol-privacy-weakening =
     to '{$policy}' or weaker if you wish to allow this transaction to proceed
     anyway.
 
+## RPC method errors
+
+err-rpc-convert-tex-invalid-address = Invalid address
+err-rpc-convert-tex-not-p2pkh = Address is not a transparent p2pkh address
```
nuttycom: I confirmed that these error messages match zcashd's error strings.

PR #371, opened 2026-02-10
Simulate wallets for stateful testing

emersonian:
Zallet developers need to be able to write stateful integration tests to match the level of test coverage that zcashd has. Our most essential RPC methods do not yet have integration test coverage (tests using actual wallet state).

Today our integration testing process is looking very manual, potentially burning a lot of dev hours: https://github.com/zcash/wallet/pull/364#issuecomment-3864359194

This PR is a first attempt at implementing an in-memory SQLite wallet for integration testing purposes.

I think we should create an issue to properly implement integration testing sooner than later.

I'm happy to iterate on this first attempt based on your feedback, or to close this PR. I'm mostly opening it to raise this as a conversation topic.

### DIFF
--- a/zallet/src/components.rs
+++ b/zallet/src/components.rs
@@ -16,6 +16,10 @@ pub(crate) mod tracing;
 #[cfg(zallet_build = "wallet")]
 pub(crate) mod keystore;
 
+#[cfg(test)]
+#[cfg(zallet_build = "wallet")]
+pub(crate) mod testing;
+
 /// A handle to a background task spawned by a component.
 ///
 /// Background tasks in Zallet are either one-shot (expected to terminate before Zallet),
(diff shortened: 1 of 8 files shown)


oxarbitrage (commented):
Thanks for the PR, it seems to be some good work here, but it’s currently trying to do too much at once, which makes it hard to identify and evaluate the core idea.

I’d recommend splitting this into smaller, focused PRs:

- Separate the new RPC method: the `dump_priv_key` RPC should live in its own PR so it can be reviewed independently.

- Reduce the test scope for now: instead of adding many tests, please include a single test that illustrates how the framework is intended to be used. Additional coverage can be added incrementally once the framework itself is approved.

- Add a high-level explanation of the framework: a short description of what the framework does, how it’s meant to be used, and what problem it’s solving would really help reviewers understand the intent and design.

PR #372, opened 2026-02-10
i18n: Internationalize CLI-visible strings

nuttycom:
Replace hardcoded English strings in CLI commands and startup/init error paths with fl!() macro invocations backed by Fluent (.ftl) message entries. JSON-RPC error messages are left unchanged to avoid breaking API consumers who may parse error strings.

### DIFF
--- a/zallet/i18n/en-US/zallet.ftl
+++ b/zallet/i18n/en-US/zallet.ftl
@@ -30,6 +30,7 @@
 -legacy_pool_seed_fingerprint = legacy_pool_seed_fingerprint
 -zallet_toml = zallet.toml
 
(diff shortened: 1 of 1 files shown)


schell (approved):
utACK [cd54dcd](https://github.com/zcash/wallet/pull/372/commits/cd54dcdeba5139db5ed3fd4fb2b13fdf2eddc477)

zallet/i18n/en-US/zallet.ftl
str4d:
```suggestion

-cfg-rpc-auth = rpc.auth
```
(just above the other similar config terms).

PR #378, opened 2026-02-17
cli: Add hidden regtest-only command for creating account and miner address

str4d:
This enables the chain-caching logic within the JSON-RPC tests to obtain a miner address that can then be configured within Zebra before the latter is started.


nullcopy (approved):
utACK https://github.com/zcash/wallet/pull/378/commits/f258a0ded7ff15e23d2a7edb3cada9750ede0cd5

str4d:
Self-ACK on my trivial fix for the feature-flagging bug. Merging so I can continue work on the integration-tests repo.

PR #400, opened 2026-03-20
Implement `z_importkey` and `z_exportkey` for Sapling

### LINKED ISSUE #69: rpc: Implement `z_exportkey`

### LINKED ISSUE #79: rpc: Implement `z_importkey`
Only Sapling support (`zcashd` only supported Sprout and Sapling).

oxarbitrage:
Adds two new RPC methods:

- `z_importkey`: Imports a Sapling extended spending key into the wallet. The key is encrypted with age and stored in the keystore. A UFVK is derived and imported into the wallet database so the sync engine can track transactions. The rescan parameter controls whether historical blocks are scanned ("yes", "no", or "whenkeyisnew"), and start_height sets the birthday for the imported account. Real treestates are fetched from the chain indexer for non-zero start heights.

- `z_exportkey`: Exports the spending key for a Sapling payment address. Requires the wallet to be unlocked. Looks up the key by iterating standalone sapling DFVKs and using decrypt_diversifier to match the address.

Also adds `fetch_account_birthday` to `json_rpc::utils` as a shared helper for building an AccountBirthday from a chain treestate.

Includes unit tests for parameter validation, key encoding/decoding roundtrips, and address parsing.

Close https://github.com/zcash/wallet/issues/69
Close https://github.com/zcash/wallet/issues/79
Closes COR-254 and COR-259

### DIFF
--- a/zallet/src/components/json_rpc/methods/import_key.rs
+++ b/zallet/src/components/json_rpc/methods/import_key.rs
@@ -1,0 +1,287 @@
+use documented::Documented;
+use jsonrpsee::core::RpcResult;
+use schemars::JsonSchema;
+use serde::Serialize;
+use zaino_state::FetchServiceSubscriber;
+use zcash_client_backend::data_api::{AccountPurpose, WalletRead, WalletWrite};
+use zcash_keys::{encoding::decode_extended_spending_key, keys::UnifiedFullViewingKey};
+use zcash_protocol::consensus::{BlockHeight, NetworkConstants};
+
+use crate::components::{
+    database::DbConnection,
+    json_rpc::{server::LegacyCode, utils::fetch_account_birthday},
+    keystore::KeyStore,
+};
+
+/// Response to a `z_importkey` RPC request.
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// Result of importing a spending key.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct ResultType {
+    /// The type of the imported address (always "sapling").
+    address_type: String,
+
+    /// The Sapling payment address corresponding to the imported key.
+    address: String,
+}
+
+pub(super) const PARAM_KEY_DESC: &str =
+    "The spending key (only Sapling extended spending keys are supported).";
+pub(super) const PARAM_RESCAN_DESC: &str = "Whether to rescan the blockchain for transactions (\"yes\", \"no\", or \"whenkeyisnew\"; default is \"whenkeyisnew\"). When rescan is enabled, the wallet's background sync engine will scan for historical transactions from the given start height.";
+pub(super) const PARAM_START_HEIGHT_DESC: &str = "Block height from which to begin the rescan (default is 0). Only used when rescan is \"yes\" or \"whenkeyisnew\" (for a new key).";
+
+/// Validates the `rescan` parameter.
+///
+/// Returns the validated rescan value, or an RPC error if the value is invalid.
+fn validate_rescan(rescan: Option<&str>) -> RpcResult<&str> {
+    match rescan {
+        None | Some("whenkeyisnew") => Ok("whenkeyisnew"),
+        Some("yes") => Ok("yes"),
+        Some("no") => Ok("no"),
+        Some(_) => Err(LegacyCode::InvalidParameter
+            .with_static("Invalid rescan value. Must be \"yes\", \"no\", or \"whenkeyisnew\".")),
+    }
+}
+
+/// Decodes a Sapling extended spending key and derives the default payment address.
+///
+/// Returns the decoded key and the encoded payment address string.
+fn decode_key_and_address(
+    hrp_spending_key: &str,
+    hrp_payment_address: &str,
+    key: &str,
+) -> RpcResult<(sapling::zip32::ExtendedSpendingKey, String)> {
+    let extsk = decode_extended_spending_key(hrp_spending_key, key).map_err(|e| {
+        LegacyCode::InvalidAddressOrKey.with_message(format!("Invalid spending key: {e}"))
+    })?;
+
+    let (_, payment_address) = extsk.default_address();
+
+    let address =
+        zcash_keys::encoding::encode_payment_address(hrp_payment_address, &payment_address);
+
+    Ok((extsk, address))
+}
+
+pub(crate) async fn call(
+    wallet: &mut DbConnection,
+    keystore: &KeyStore,
+    chain: FetchServiceSubscriber,
+    key: &str,
+    rescan: Option<&str>,
+    start_height: Option<u64>,
+) -> Response {
+    let rescan = validate_rescan(rescan)?;
+
+    // Resolve and validate start_height, defaulting to 0 (genesis).
+    let start_height = BlockHeight::from_u32(
+        u32::try_from(start_height.unwrap_or(0))
+            .map_err(|_| LegacyCode::InvalidParameter.with_static("Block height out of range."))?,
+    );
+
+    let chain_tip = wallet
+        .chain_height()
+        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?;
+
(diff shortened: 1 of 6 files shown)


ValarDragon:
I suggest we add the import key logic here, so zcashd folks can migrate

zallet/src/components/json_rpc/utils.rs
```diff
@@ -44,6 +52,63 @@ pub(super) async fn ensure_wallet_is_unlocked(keystore: &KeyStore) -> RpcResult<
...
+        ));
+    }
+
+    let treestate_height = height.saturating_sub(1);
+    let treestate = chain
+        .fetcher
+        .get_treestate(treestate_height.to_string())
+        .await
+        .map_err(|e| {
+            LegacyCode::InvalidParameter.with_message(format!(
```
nullcopy: `s/LegacyCode::InvalidParameter/RpcErrorCode::InternalError/`

zallet/src/components/json_rpc/methods/import_key.rs
```diff
@@ -0,0 +1,287 @@
...
+        u32::try_from(start_height.unwrap_or(0))
+            .map_err(|_| LegacyCode::InvalidParameter.with_static("Block height out of range."))?,
+    );
+
+    let chain_tip = wallet
+        .chain_height()
+        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?;
+
+    if let Some(tip) = chain_tip {
+        if start_height > tip {
```
nullcopy: (not blocking) might be nice to skip this check if rescan="no".

zallet/src/components/json_rpc/methods/import_key.rs
```diff
@@ -0,0 +1,287 @@
...
+        .is_none();
+
+    if is_new_key {
+        // Determine the birthday height based on the rescan parameter:
+        // - "yes" or "whenkeyisnew" → use start_height so the sync engine scans
+        //   historical blocks from that point.
+        // - "no" → use the current chain tip so the sync engine only tracks new
+        //   transactions going forward.
+        //
+        // TODO: When rescan is "yes" and the key already exists, zcashd would force a
```
nullcopy, line 127: The recently added `rewind_to_height()` method should do this for you now

zallet/src/components/json_rpc/methods/import_key.rs
```diff
@@ -0,0 +1,287 @@
...
+        // rescan from start_height. We currently skip this because zcash_client_sqlite
+        // does not expose a way to reset scan ranges for an existing account.
+        let effective_height = match rescan {
+            "yes" | "whenkeyisnew" => start_height,
+            "no" => chain_tip.unwrap_or(BlockHeight::from_u32(0)),
```
nullcopy: Not sure how I feel about silently failing over to 0 here

nullcopy: Thinking on it some more, I'm actually ok with it. I would support logging a warning here, though (not blocking)

zallet/src/components/json_rpc/utils.rs
```diff
@@ -44,6 +52,63 @@ pub(super) async fn ensure_wallet_is_unlocked(keystore: &KeyStore) -> RpcResult<
...
+/// the chain indexer so the sync engine can validate note commitment tree continuity.
+#[cfg(zallet_build = "wallet")]
+pub(super) async fn fetch_account_birthday(
+    wallet: &DbConnection,
+    chain: &FetchServiceSubscriber,
+    height: BlockHeight,
+) -> RpcResult<AccountBirthday> {
+    if height == BlockHeight::from_u32(0) {
+        return Ok(AccountBirthday::from_parts(
+            ChainState::empty(BlockHeight::from_u32(0), BlockHash([0; 32])),
```
nullcopy, line 144: Should this at least include the real genesis hash?

nullcopy (approved):
Looks good to me! I left a few minor comments, but nothing blocking (modulo merge conflicts)

utACK 532371ebcd3e72161539f224a4704ed5428b3e73

zallet/src/components/json_rpc/utils.rs
oxarbitrage: Done — now `InternalError` (kept the message via `ErrorObjectOwned`, since `ErrorCode` has no `with_message`).

zallet/src/components/json_rpc/methods/import_key.rs
oxarbitrage: Done — the `start_height > tip` check is now skipped when `rescan="no"`.

zallet/src/components/json_rpc/methods/import_key.rs
oxarbitrage: Deferred — `rewind_to_height` rewinds the whole wallet (all accounts), not just this key, so I left the TODO (now referencing it) rather than trigger a global rewind on import.

zallet/src/components/json_rpc/methods/import_key.rs
oxarbitrage: Superseded by your follow-up below — added a warning log.

oxarbitrage: Done — `tracing::warn!` when the `rescan="no"` path falls back to genesis.

zallet/src/components/json_rpc/utils.rs
oxarbitrage: Left as zeros, with a comment explaining why: the chain state is the state *before* the birthday, and nothing precedes genesis, so the zero parent hash is the correct anchor — the real genesis hash would be the hash of block 0 itself.

oxarbitrage:
Thanks for the review, and apologies for the delay in getting back to this!

Rebased onto `main` and resolved the `keystore.rs` conflict from the migration refactor: the encrypt helpers are now `Encryptor` methods on `main`, so I dropped the branch's duplicate free functions and kept only the new `decrypt_standalone_sapling_extsk`; I also dropped the `zcashd-import` cfg from `Encryptor::encrypt_standalone_sapling_key`, since the always-built `z_importkey` path needs it.

Addressed the review comments (replied inline on each thread):
- `fetch_account_birthday`: treestate-fetch failure now returns `InternalError` instead of `InvalidParameter`.
- `import_key`: skip the `start_height > tip` check when `rescan="no"` (start_height is unused there); `tracing::warn!` when the `rescan="no"` path falls back to genesis; refined the existing-key-rescan TODO to note `rewind_to_height` (deferred, since it rewinds the whole wallet rather than just this key).
- Genesis `ChainState`: left as `BlockHash([0; 32])` with a comment explaining why (no block precedes genesis, so the zero parent hash is the correct anchor).

Re @nuttycom's concern: `z_exportkey`'s help text now warns that it is not a complete backup (Orchard funds under the same spending authority are not exported). Added a `CHANGELOG` entry for both methods.

Note: the two CI failures (`Test NU7` and `Latest build`) look pre-existing and dependency-related (`zebra-chain` not covering `NetworkUpgrade::Nu7`, and `orchard = "^0.13"` failing to resolve), not from this PR.

oxarbitrage:
Validated this PR end-to-end against the dedicated integration test (zcash/integration-tests#76, `wallet_import_export_key.py`) by running it through the interop CI (`ZIT-Revision`), building zallet from this branch:

- ✅ `wallet_import_export_key.py | True | 16 s` — covers import → `listaddresses` (sapling UA under `imported_watchonly`) → re-import idempotency → `z_exportkey` round-trip → cross-wallet isolation → `rescan` param values → and the error paths (out-of-range height, invalid rescan, invalid key, address-not-in-wallet).

The only red shard in that run was an unrelated, pre-existing failure on `nuparams.py` (`getblocksubsidy` subsidy assertion), which this PR's diff doesn't touch — tracking that separately on the integration-tests side.

This is approved + mergeable + green, and now interop-validated. **Planning to merge shortly.**

Issue #217, opened 2025-08-28
rpc: Design and implement `z_sendfromaccount`


str4d:
`z_sendmany` has weird semantics inherited from several iterations of wallet design. In particular, you have to specify an address to send from, but for UAs we use those as representatives of their corresponding accounts (rather than only selecting funds that were sent to the specified addresses).

We should instead create a new RPC method that takes one (or more?) account UUID(s) as the source of funds.

Additionally, we should consider whether this RPC should be designed around proposals:
- `z_sendfromaccount` would return a proposal object, that includes information about the privacy implications of the transaction that the caller can review.
  - This does not necessarily have to be the same `Proposal` protobuf as the mobile SDKs use, because Zallet can internally cache proposed transactions. It could instead directly convey just the (economic and privacy) effects of the proposed transaction.
  - Zallet could prepare and create proofs for the proposed transaction asynchronously while waiting for spending approval.
- A second RPC would take a pointer (UUID?) to a proposal, and calling it would constitute acceptance of the transaction. Any required signatures would be applied at this time, and then the transaction would be broadcast.

This has similarities to / overlaps with the plan to add RPC methods for working with PCZTs (#99).

nuttycom:
The original specification of z_sendfromaccount is not ideal for fully-automated workflows. I think that `z_sendfromaccount` should be similar in spirit to `z_sendmany` in terms of how it uses the privacy policy, but we should additionally provide `z_proposetransaction` and `z_finalizetransaction`; `z_sendfromaccount` can then be implemented in terms of these operations under the hood.

Arguments to z_proposetransaction and z_sendfromaccount:
* account UUID
* fund source (orchard | sapling | any_transparent | {specific taddrs})
* recipient information (same as z_sendmany)
* minconf (same as z_sendmany)

High-level design:

* z_proposetransaction should return both the PCZT for detailed inspection, and the privacy policy required to execute the transaction 
* z_finalizetransaction should take a PCZT and the privacy policy (as an acknowledgement of the transaction's privacy implications)
* z_sendfromaccount elides the interstitial PCZT step, but additionally takes the privacy policy as acknowledgement that the user understands the privacy implications of the rest of their request. Unlike `z_sendmany`, `z_sendfromaccount` will not use the background processing system; it will perform the entire operation in one shot.

PR #402, opened 2026-03-20
feat: COR-264 add z_shieldcoinbase to WalletRpc

### LINKED ISSUE #89: rpc: Implement `z_shieldcoinbase`

### LINKED ISSUE #436: Transparent ingestion path does not record `tx_index`, breaking `TransparentOutputFilter::CoinbaseOnly` consumers
`zallet`'s UTXO ingestion path doesn't set the transaction index for UTXOs it discovers through `zaino`'s `z_get_address_utxos`.

`zallet/src/components/sync.rs:540` ingests transparent UTXOs via:
```rust
let output = WalletTransparentOutput::from_parts(
    OutPoint::new(txid.0, index.index()),
    TxOut::new(...),
    Some(BlockHeight::from_u32(mined_height.0)),  // mined_height only
)...
db_data.put_received_transparent_utxo(&output)?;
```
Both `WalletTransparentOutput::from_parts` (in `zcash_client_backend 0.22`) and `put_received_transparent_utxo` (in `zcash_client_sqlite 0.20`) accept only `mined_height`. They have no parameter for block-level transaction index. So `tx_index` stays `NULL` for every transparent-only transaction the wallet learns about via Zaino's `z_get_address_utxos`.
The other ingestion path (`decrypt_and_store_transaction` -> `store_decrypted_tx` -> `put_tx_meta`) DOES set `tx_index`, but that path runs only for transactions with shielded outputs the wallet can decrypt. A pure-transparent coinbase tx in regtest never goes through it.

This is currently _not_ blocking an implementation for 
* #402 
as there are **two** possible fixes: 
   - `zallet`: extend `zallet`'s sync to fetch the full block (not just `z_get_address_utxos`) for any block containing a wallet-relevant transparent UTXO, then derive `tx_index` from the block ordering and write it via a richer `put_received_transparent_utxo` API.
   - `librustzcash`: propose adding `tx_index` (or a `block_index: u16`) to `WalletTransparentOutput::from_parts` upstream and threading it through to `put_transparent_output` in `zcash_client_sqlite`. This is the cleaner fix because it acknowledges that any wallet using `TransparentOutputFilter::CoinbaseOnly` must record this datum.

schell:
COR-264

## Summary

Adds the `z_shieldcoinbase` JSON-RPC method, integrated with the upstream `propose_shielding_coinbase` API ([librustzcash#2353](https://github.com/zcash/librustzcash/pull/2353), merged). Matches `zcashd`'s pre-flight return shape (`{ remainingUTXOs, remainingValue, shieldingUTXOs, shieldingValue, opid }`). End-to-end verified against [`zcash/integration-tests#80`](https://github.com/zcash/integration-tests/pull/80) (all 10 tests pass).

The `memo` and `limit` parameters are honored: `memo` is attached to the resulting shielded payment, and `limit` caps the number of selected coinbase UTXOs to the highest-value `n` of those eligible. The proposal produces no transparent or shielded change, preserving the privacy invariant that a shielded change output would otherwise let `toaddress` learn the sender's total selected-coinbase value.

## Commits

1. `feat(rpc): add z_shieldcoinbase` — the RPC method itself, initially built against `propose_shielding` + `TransparentOutputFilter::CoinbaseOnly`.
2. `fix(rpc): accumulate standalone transparent keys per address` — replaces an overwriting `keys.insert(...)` with `entry().or_default().push(...)` so duplicate addresses don't drop signing keys. Applied in both `z_sendmany` and `z_shieldcoinbase`.
3. `feat(rpc): match zcashd's z_shieldcoinbase pre-flight return shape` — returns the 5-field object instead of a bare opid string.
4. `fix(sync): record tx_index for coinbase transactions during transparent ingest` — `fetch_transparent_utxos` now routes each observed tx through `decrypt_and_store_transaction`, which sets `tx_index = 0` for coinbase. Required for `TransparentOutputFilter::CoinbaseOnly` to work. Fixes #436.
5. `fix(rpc): only decrypt standalone keys for standalone-imported addresses` — pre-filters the proposal-input loop to skip HD-derived addresses, which have no standalone-key entry. Without this, every `z_shieldcoinbase`/`z_sendmany` call with HD-derived inputs failed with `"Query returned no rows"`.
6. `deps + feat: integrate propose_shielding_coinbase upstream` — bumps librustzcash crates to a `[patch.crates-io]` pin at `zcash/librustzcash@541f5b562e` (PR #2353 merge), absorbs the API changes that came along (`WalletTransparentOutput<AccountId>` generic, `WalletUtxo` removal, `rewind_to_height` → `rewind_to_chain_state`, `from_parts` extra args), and rewrites `z_shieldcoinbase` to call `propose_shielding_coinbase`. The `memo` and `limit` parameters are now honored, and the `_memo` / `_limit` TODOs are removed.

Fixes #89
Fixes #436

The `[patch.crates-io]` block in commit 6 is temporary — drop it once librustzcash cuts a release containing `propose_shielding_coinbase`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ be considered breaking changes.
   - `verifymessage`
   - `z_converttex`
   - `z_importaddress`
+  - `z_shieldcoinbase`
 
 ### Changed
 - `getrawtransaction` now correctly reports the fields `asm`, `reqSigs`, `kind`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,8 @@ be considered breaking changes.
   `z_shieldcoinbase` (and any other consumer of
   `TransparentOutputFilter::CoinbaseOnly`) to correctly identify coinbase
   outputs.
-- `z_sendmany` and `z_shieldcoinbase` no longer fail with `Query returned no
-  rows` when a proposal includes inputs at HD-derived transparent addresses.
+- `z_sendmany` no longer fails with `Query returned no rows` when a proposal
+  includes inputs at HD-derived transparent addresses.
   The keystore's standalone-key decryption is now invoked only for addresses
   that were imported standalone; HD-derived addresses are signed for using
   the account's unified spending key.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,9 @@ be considered breaking changes.
   not defined.
 - Zallet now refuses to open wallet databases from incompatible earlier alpha
   releases instead of attempting to migrate them.
-- `z_sendmany` and `z_shieldcoinbase` no longer drop standalone transparent
-  signing keys when the same address backs multiple proposal inputs. Keys
-  are now accumulated per address rather than overwritten.
+- `z_sendmany` no longer drop standalone transparent signing keys when the same
+  address backs multiple proposal inputs. Keys are now accumulated per address
+  rather than overwritten.
 - Transparent UTXO ingestion now records `tx_index` for coinbase transactions
   by routing each observed transaction through `decrypt_and_store_transaction`
   in addition to `put_received_transparent_utxo`. This enables
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,11 @@ be considered breaking changes.
   `z_shieldcoinbase` (and any other consumer of
   `TransparentOutputFilter::CoinbaseOnly`) to correctly identify coinbase
   outputs.
+- `z_sendmany` and `z_shieldcoinbase` no longer fail with `Query returned no
+  rows` when a proposal includes inputs at HD-derived transparent addresses.
+  The keystore's standalone-key decryption is now invoked only for addresses
+  that were imported standalone; HD-derived addresses are signed for using
+  the account's unified spending key.
 
 ## [0.1.0-alpha.3] - 2025-12-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ be considered breaking changes.
   not defined.
 - Zallet now refuses to open wallet databases from incompatible earlier alpha
   releases instead of attempting to migrate them.
+- `z_sendmany` and `z_shieldcoinbase` no longer drop standalone transparent
+  signing keys when the same address backs multiple proposal inputs. Keys
+  are now accumulated per address rather than overwritten.
 
 ## [0.1.0-alpha.3] - 2025-12-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,12 @@ be considered breaking changes.
 - `z_sendmany` and `z_shieldcoinbase` no longer drop standalone transparent
   signing keys when the same address backs multiple proposal inputs. Keys
   are now accumulated per address rather than overwritten.
+- Transparent UTXO ingestion now records `tx_index` for coinbase transactions
+  by routing each observed transaction through `decrypt_and_store_transaction`
+  in addition to `put_received_transparent_utxo`. This enables
+  `z_shieldcoinbase` (and any other consumer of
+  `TransparentOutputFilter::CoinbaseOnly`) to correctly identify coinbase
+  outputs.
 
 ## [0.1.0-alpha.3] - 2025-12-15
 
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -543,26 +543,37 @@ pub(crate) trait WalletRpc {
     ) -> z_send_many::Response;
 
     /// Shields coinbase UTXOs by sending them from a transparent address (or all
-    /// wallet taddrs) to a shielded address.
+    /// wallet taddrs) to a shielded address within the same account.
     ///
     /// This is an asynchronous operation; it returns an operation ID that can
     /// be used with `z_getoperationstatus` or `z_getoperationresult`.
-    ///                                                                                                                     
-    /// # Arguments                                                                                                         
-    /// - `fromaddress` (string, required): The transparent address to sweep
-    ///   from, or `"*"` to sweep from all wallet taddrs.
-    /// - `toaddress` (string, required): The shielded address to receive the funds.                                        
-    /// - `fee` (numeric, optional): The fee amount in ZEC to attach to this
-    ///   transaction. If not provided, the default ZIP-317 fee is used.                                                                                                            
-    /// - `limit` (numeric, optional): Limit on the maximum number of UTXOs to
-    ///   shield. Set to 0 to use as many as will fit in the transaction.                                                                                                         
-    /// - `memo` (string, optional): Encoded as hex. This will be stored in the
-    ///   memo field of the new note.
+    ///
+    /// **Note:** This method's behavior differs from `zcashd`. The `toaddress`
+    /// must be a wallet-owned shielded address and is used to select the account
+    /// whose transparent UTXOs will be shielded. The resulting transaction sends
+    /// funds to the account's internal shielded address, which may differ from
+    /// `toaddress`. The `fromaddress` must also belong to the same account.
+    ///
+    /// # Arguments
+    /// - `fromaddress` (string, required): A wallet-owned transparent address to
+    ///   sweep from, or `"*"` to sweep from all taddrs belonging to the same
+    ///   account as `toaddress`. Must belong to the same account as `toaddress`.
+    /// - `toaddress` (string, required): A wallet-owned shielded address (Sapling,
+    ///   Orchard, or Unified with shielded receivers) used to identify the account.
+    ///   Funds are shielded into the account's internal shielded address, which may
+    ///   differ from this address.
+    /// - `fee` (optional): If provided, this parameter must be `null`; any
+    ///   non-null value will be rejected. The value of this field is ignored,
+    ///   as Zallet always calculates and applies the ZIP-317 fee internally.
+    /// - `limit` (numeric, optional): Accepted for compatibility but currently
+    ///   ignored; it does not constrain how many UTXOs are shielded.
+    /// - `memo` (string, optional): Accepted for compatibility but currently
+    ///   ignored; it is not stored in the memo field of any new note.
     /// - `privacy_policy` (string, optional): Policy for what information
-    ///   leakage is acceptable.                           
-    ///                                                                                                                     
-    /// # Returns                                                                                                           
-    /// - An operation ID for tracking the async operation.                                                                 
+    ///   leakage is acceptable.
+    ///
+    /// # Returns
+    /// - An operation ID for tracking the async operation.
     #[method(name = "z_shieldcoinbase")]
     async fn z_shieldcoinbase(
         &self,
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -1,3 +1,4 @@
+//! Contains general and wallet-specific RPC trait methods.
 use async_trait::async_trait;
 use jsonrpsee::{
     core::{JsonValue, RpcResult},
@@ -60,6 +61,8 @@ mod view_transaction;
 mod z_get_total_balance;
 #[cfg(zallet_build = "wallet")]
 mod z_send_many;
+#[cfg(zallet_build = "wallet")]
+mod z_shieldcoinbase;
 
 /// The general JSON-RPC interface, containing the methods provided in all Zallet builds.
 #[rpc(server)]
@@ -538,6 +541,38 @@ pub(crate) trait WalletRpc {
         fee: Option<JsonValue>,
         privacy_policy: Option<String>,
     ) -> z_send_many::Response;
+
+    /// Shields coinbase UTXOs by sending them from a transparent address (or all
+    /// wallet taddrs) to a shielded address.
+    ///
+    /// This is an asynchronous operation; it returns an operation ID that can
+    /// be used with `z_getoperationstatus` or `z_getoperationresult`.
+    ///                                                                                                                     
+    /// # Arguments                                                                                                         
+    /// - `fromaddress` (string, required): The transparent address to sweep
+    ///   from, or `"*"` to sweep from all wallet taddrs.
+    /// - `toaddress` (string, required): The shielded address to receive the funds.                                        
+    /// - `fee` (numeric, optional): The fee amount in ZEC to attach to this
+    ///   transaction. If not provided, the default ZIP-317 fee is used.                                                                                                            
+    /// - `limit` (numeric, optional): Limit on the maximum number of UTXOs to
+    ///   shield. Set to 0 to use as many as will fit in the transaction.                                                                                                         
+    /// - `memo` (string, optional): Encoded as hex. This will be stored in the
+    ///   memo field of the new note.
+    /// - `privacy_policy` (string, optional): Policy for what information
+    ///   leakage is acceptable.                           
+    ///                                                                                                                     
+    /// # Returns                                                                                                           
+    /// - An operation ID for tracking the async operation.                                                                 
+    #[method(name = "z_shieldcoinbase")]
+    async fn z_shieldcoinbase(
+        &self,
+        fromaddress: String,
+        toaddress: String,
+        fee: Option<JsonValue>,
+        limit: Option<u32>,
+        memo: Option<String>,
+        privacy_policy: Option<String>,
+    ) -> z_shieldcoinbase::Response;
 }
 
 pub(crate) struct RpcImpl {
@@ -546,7 +581,6 @@ pub(crate) struct RpcImpl {
     keystore: KeyStore,
     chain: Chain,
 }
-
 impl RpcImpl {
     /// Creates a new instance of the general RPC handler.
     pub(crate) fn new(
@@ -852,6 +886,33 @@ impl WalletRpcServer for WalletRpcImpl {
                     amounts,
                     minconf,
                     fee,
+                    privacy_policy,
+                )
+                .await?,
+            )
+            .await)
+    }
+
+    async fn z_shieldcoinbase(
+        &self,
+        fromaddress: String,
+        toaddress: String,
+        fee: Option<JsonValue>,
+        limit: Option<u32>,
+        memo: Option<String>,
+        privacy_policy: Option<String>,
+    ) -> z_shieldcoinbase::Response {
+        Ok(self
+            .start_async(
+                z_shieldcoinbase::call(
+                    self.wallet().await?,
+                    self.keystore.clone(),
+                    self.chain().await?,
+                    fromaddress,
+                    toaddress,
+                    fee,
+                    limit,
+                    memo,
                     privacy_policy,
                 )
                 .await?,
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -62,6 +62,8 @@ mod z_get_total_balance;
 mod z_import_address;
 #[cfg(zallet_build = "wallet")]
 mod z_send_many;
+#[cfg(zallet_build = "wallet")]
+mod z_shieldcoinbase;
 
 /// The general JSON-RPC interface, containing the methods provided in all Zallet builds.
 #[rpc(server)]
@@ -561,6 +563,50 @@ pub(crate) trait WalletRpc {
         fee: Option<JsonValue>,
         privacy_policy: Option<String>,
     ) -> z_send_many::Response;
+
+    /// Shields coinbase UTXOs by sending them from a transparent address (or all
+    /// wallet taddrs) to a shielded address within the same account.
+    ///
+    /// This is an asynchronous operation; it returns an operation ID that can
+    /// be used with `z_getoperationstatus` or `z_getoperationresult`.
+    ///
+    /// **Note:** This method's behavior differs from `zcashd`. The `toaddress`
+    /// must be a wallet-owned shielded address and is used to select the account
+    /// whose transparent UTXOs will be shielded. The resulting transaction sends
+    /// funds to the account's internal shielded address, which may differ from
+    /// `toaddress`. The `fromaddress` must also belong to the same account.
+    ///
+    /// # Arguments
+    /// - `fromaddress` (string, required): A wallet-owned transparent address to
+    ///   sweep from, or `"*"` to sweep from all taddrs belonging to the same
+    ///   account as `toaddress`. Must belong to the same account as `toaddress`.
+    /// - `toaddress` (string, required): A wallet-owned shielded address (Sapling,
+    ///   Orchard, or Unified with shielded receivers) used to identify the account.
+    ///   Funds are shielded into the account's internal shielded address, which may
+    ///   differ from this address.
+    /// - `fee` (optional): This parameter must be omitted or set to `null`;
+    ///   any other value will be rejected. When accepted, the value of this field
+    ///   is ignored, as Zallet always calculates and applies the ZIP-317 fee
+    ///   internally.
+    /// - `limit` (numeric, optional): Accepted for compatibility but currently
+    ///   ignored; it does not constrain how many UTXOs are shielded.
+    /// - `memo` (string, optional): Accepted for compatibility but currently
+    ///   ignored; it is not stored in the memo field of any new note.
+    /// - `privacy_policy` (string, optional): Policy for what information
+    ///   leakage is acceptable.
+    ///
+    /// # Returns
+    /// - An operation ID for tracking the async operation.
+    #[method(name = "z_shieldcoinbase")]
+    async fn z_shieldcoinbase(
+        &self,
+        fromaddress: String,
+        toaddress: String,
+        fee: Option<JsonValue>,
+        limit: Option<u32>,
+        memo: Option<String>,
+        privacy_policy: Option<String>,
+    ) -> z_shieldcoinbase::Response;
 }
 
 pub(crate) struct RpcImpl {
@@ -891,6 +937,33 @@ impl WalletRpcServer for WalletRpcImpl {
                     amounts,
                     minconf,
                     fee,
+                    privacy_policy,
+                )
+                .await?,
+            )
+            .await)
+    }
+
+    async fn z_shieldcoinbase(
+        &self,
+        fromaddress: String,
+        toaddress: String,
+        fee: Option<JsonValue>,
+        limit: Option<u32>,
+        memo: Option<String>,
+        privacy_policy: Option<String>,
+    ) -> z_shieldcoinbase::Response {
+        Ok(self
+            .start_async(
+                z_shieldcoinbase::call(
+                    self.wallet().await?,
+                    self.keystore.clone(),
+                    self.chain().await?,
+                    fromaddress,
+                    toaddress,
+                    fee,
+                    limit,
+                    memo,
                     privacy_policy,
                 )
                 .await?,
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -596,7 +596,16 @@ pub(crate) trait WalletRpc {
     ///   leakage is acceptable.
     ///
     /// # Returns
-    /// - An operation ID for tracking the async operation.
+    /// An object matching `zcashd`'s `z_shieldcoinbase` shape:
+    /// - `remainingUTXOs` (numeric): Number of coinbase UTXOs eligible for
+    ///   shielding that were not selected. Zallet currently ignores `limit`,
+    ///   so in practice this is `0` whenever the proposal succeeded.
+    /// - `remainingValue` (numeric, ZEC): Total value of those UTXOs.
+    /// - `shieldingUTXOs` (numeric): Number of coinbase UTXOs being shielded
+    ///   by this operation.
+    /// - `shieldingValue` (numeric, ZEC): Total value being shielded.
+    /// - `opid` (string): Operation id to pass to `z_getoperationstatus` or
+    ///   `z_getoperationresult` to retrieve the final result.
     #[method(name = "z_shieldcoinbase")]
     async fn z_shieldcoinbase(
         &self,
@@ -953,21 +962,19 @@ impl WalletRpcServer for WalletRpcImpl {
         memo: Option<String>,
         privacy_policy: Option<String>,
     ) -> z_shieldcoinbase::Response {
-        Ok(self
-            .start_async(
-                z_shieldcoinbase::call(
-                    self.wallet().await?,
-                    self.keystore.clone(),
-                    self.chain().await?,
-                    fromaddress,
-                    toaddress,
-                    fee,
-                    limit,
-                    memo,
-                    privacy_policy,
-                )
-                .await?,
-            )
-            .await)
+        let (preflight, context, fut) = z_shieldcoinbase::call(
+            self.wallet().await?,
+            self.keystore.clone(),
+            self.chain().await?,
+            fromaddress,
+            toaddress,
+            fee,
+            limit,
+            memo,
+            privacy_policy,
+        )
+        .await?;
+        let opid = self.start_async((context, fut)).await;
+        Ok(z_shieldcoinbase::ShieldCoinbaseResult::new(preflight, opid))
     }
 }
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -564,54 +564,65 @@ pub(crate) trait WalletRpc {
         privacy_policy: Option<String>,
     ) -> z_send_many::Response;
 
-    /// Shields coinbase UTXOs by sending them from a transparent address (or all
-    /// wallet taddrs) to a shielded address within the same account.
+    /// Shields coinbase UTXOs from a single wallet-owned source into a
+    /// shielded address.
     ///
-    /// This is an asynchronous operation; it returns an operation ID that can
-    /// be used with `z_getoperationstatus` or `z_getoperationresult`.
-    ///
-    /// **Note:** This method's behavior differs from `zcashd`. The `toaddress`
-    /// must be a wallet-owned shielded address and is used to select the account
-    /// whose transparent UTXOs will be shielded. The resulting transaction sends
-    /// funds to the account's internal shielded address, which may differ from
-    /// `toaddress`. The `fromaddress` must also belong to the same account.
+    /// This is an asynchronous operation; it returns an operation id that
+    /// can be used with `z_getoperationstatus` or `z_getoperationresult`.
     ///
     /// # Arguments
-    /// - `fromaddress` (string, required): A wallet-owned transparent address to
-    ///   sweep from, or `"*"` to sweep from all taddrs belonging to the same
-    ///   account as `toaddress`. Must belong to the same account as `toaddress`.
-    /// - `toaddress` (string, required): A wallet-owned shielded address (Sapling,
-    ///   Orchard, or Unified with shielded receivers) used to identify the account.
-    ///   Funds are shielded into the account's internal shielded address, which may
-    ///   differ from this address.
-    /// - `fee` (optional): This parameter must be omitted or set to `null`;
-    ///   any other value will be rejected. When accepted, the value of this field
-    ///   is ignored, as Zallet always calculates and applies the ZIP-317 fee
-    ///   internally.
-    /// - `limit` (numeric, optional): If supplied, caps the number of selected
-    ///   coinbase UTXOs to the highest-value `n` of those eligible. Recommended
-    ///   for wallets with many eligible coinbase UTXOs: without it, a single
-    ///   transaction is built containing all eligible UTXOs, which can exceed
-    ///   transaction-size limits at broadcast time. Zallet logs a warning (but
-    ///   does not enforce a hard cap) when the proposal selects more than ~400
-    ///   inputs.
-    /// - `memo` (string, optional): If supplied, stored in the memo field of
-    ///   the resulting shielded payment. Must be a hex-encoded string (up to
-    ///   1024 hex characters = 512 bytes).
+    /// - `fromaddress`: Either a single transparent address owned by this
+    ///   wallet, or an account UUID (string) to sweep every coinbase UTXO
+    ///   across that account's transparent receivers. Arrays of addresses
+    ///   are not accepted; pass the account UUID instead if you need to
+    ///   sweep across multiple receivers. Unlike `zcashd`, the wildcard
+    ///   `"*"` (sweep all wallet t-addrs) is rejected with an
+    ///   `InvalidParameter` error: cross-account sweeps would correlate
+    ///   otherwise-unrelated accounts on-chain, so callers must scope the
+    ///   sweep to a single account by passing its UUID.
+    /// - `toaddress`: Any Zcash shielded address (Sapling, Orchard, or
+    ///   Unified with a shielded receiver) that will receive the shielded
+    ///   funds. Need not belong to this wallet. Transparent / TEX
+    ///   destinations are rejected by the backend.
+    /// - `fee` (numeric, optional): If set, it must be null. Zallet always
+    ///   uses a fee calculated according to ZIP 317; the parameter is
+    ///   accepted for positional compatibility with `zcashd`'s
+    ///   `z_shieldcoinbase` only.
+    /// - `limit` (numeric, optional): If supplied, caps the number of
+    ///   selected coinbase UTXOs to the highest-value `n` of those
+    ///   eligible. Recommended for wallets with many eligible coinbase
+    ///   UTXOs: without it, a single transaction is built containing all
+    ///   eligible UTXOs, which can exceed transaction-size limits at
+    ///   broadcast time.
+    /// - `memo` (string, optional): If supplied, stored in the memo field
+    ///   of the resulting shielded payment. Hex-encoded, up to 1024 hex
+    ///   characters (= 512 bytes).
     /// - `privacy_policy` (string, optional): Policy for what information
-    ///   leakage is acceptable.
+    ///   leakage is acceptable. Coinbase shielding always reveals the
+    ///   source transparent address(es), so only two policy values are
+    ///   accepted:
+    ///   - `"AllowRevealedSenders"`: Allow revealing the source transparent
+    ///     address. Sufficient when sweeping from a single t-addr.
+    ///   - `"AllowLinkingAccountAddresses"`: Additionally allow linking
+    ///     multiple source t-addrs on-chain. Required when sweeping from
+    ///     an account UUID that expands to >1 transparent receiver with
+    ///     eligible coinbase UTXOs.
+    ///
+    ///   When omitted or `null`, the default is chosen from `fromaddress`:
+    ///   `"AllowRevealedSenders"` for a single t-addr, or
+    ///   `"AllowLinkingAccountAddresses"` for an account UUID. Any other
+    ///   policy name (including stricter values like `"FullPrivacy"` and
+    ///   looser ones like `"NoPrivacy"`) is rejected.
     ///
     /// # Returns
     /// An object matching `zcashd`'s `z_shieldcoinbase` shape:
-    /// - `remainingUTXOs` (numeric): Number of coinbase UTXOs eligible for
-    ///   shielding that were not selected. Non-zero when `limit` is smaller
-    ///   than the count of eligible coinbase UTXOs.
+    /// - `remainingUTXOs` (numeric): Eligible-but-not-selected coinbase UTXO
+    ///   count.
     /// - `remainingValue` (numeric, ZEC): Total value of those UTXOs.
-    /// - `shieldingUTXOs` (numeric): Number of coinbase UTXOs being shielded
-    ///   by this operation.
+    /// - `shieldingUTXOs` (numeric): Number of coinbase UTXOs being
+    ///   shielded by this operation.
     /// - `shieldingValue` (numeric, ZEC): Total value being shielded.
-    /// - `opid` (string): Operation id to pass to `z_getoperationstatus` or
-    ///   `z_getoperationresult` to retrieve the final result.
+    /// - `opid` (string): Operation id.
     #[method(name = "z_shieldcoinbase")]
     async fn z_shieldcoinbase(
         &self,
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -589,7 +589,12 @@ pub(crate) trait WalletRpc {
     ///   is ignored, as Zallet always calculates and applies the ZIP-317 fee
     ///   internally.
     /// - `limit` (numeric, optional): If supplied, caps the number of selected
-    ///   coinbase UTXOs to the highest-value `n` of those eligible.
+    ///   coinbase UTXOs to the highest-value `n` of those eligible. Recommended
+    ///   for wallets with many eligible coinbase UTXOs: without it, a single
+    ///   transaction is built containing all eligible UTXOs, which can exceed
+    ///   transaction-size limits at broadcast time. Zallet logs a warning (but
+    ///   does not enforce a hard cap) when the proposal selects more than ~400
+    ///   inputs.
     /// - `memo` (string, optional): If supplied, stored in the memo field of
     ///   the resulting shielded payment. Must be a hex-encoded string (up to
     ///   1024 hex characters = 512 bytes).
--- a/zallet/src/components/json_rpc/methods.rs
+++ b/zallet/src/components/json_rpc/methods.rs
@@ -588,18 +588,19 @@ pub(crate) trait WalletRpc {
     ///   any other value will be rejected. When accepted, the value of this field
     ///   is ignored, as Zallet always calculates and applies the ZIP-317 fee
     ///   internally.
-    /// - `limit` (numeric, optional): Accepted for compatibility but currently
-    ///   ignored; it does not constrain how many UTXOs are shielded.
-    /// - `memo` (string, optional): Accepted for compatibility but currently
-    ///   ignored; it is not stored in the memo field of any new note.
+    /// - `limit` (numeric, optional): If supplied, caps the number of selected
+    ///   coinbase UTXOs to the highest-value `n` of those eligible.
+    /// - `memo` (string, optional): If supplied, stored in the memo field of
+    ///   the resulting shielded payment. Must be a hex-encoded string (up to
+    ///   1024 hex characters = 512 bytes).
     /// - `privacy_policy` (string, optional): Policy for what information
     ///   leakage is acceptable.
     ///
     /// # Returns
     /// An object matching `zcashd`'s `z_shieldcoinbase` shape:
     /// - `remainingUTXOs` (numeric): Number of coinbase UTXOs eligible for
-    ///   shielding that were not selected. Zallet currently ignores `limit`,
-    ///   so in practice this is `0` whenever the proposal succeeded.
+    ///   shielding that were not selected. Non-zero when `limit` is smaller
+    ///   than the count of eligible coinbase UTXOs.
     /// - `remainingValue` (numeric, ZEC): Total value of those UTXOs.
     /// - `shieldingUTXOs` (numeric): Number of coinbase UTXOs being shielded
     ///   by this operation.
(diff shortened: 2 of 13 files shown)


ValarDragon:
What is needed to help unblock this? This is as far as I'm tracking the last piece to unblock miners from being Zebra/Zallet based.

Linking for reference this commit which is being used by some mainnet miners right now to build z_shieldcoinbase: https://github.com/zkclaudecoder/zallet/commit/634fb799aac2aa5c29e9195971fca47771e04c7b

schell:
@ValarDragon thanks for the link to that commit :) I'm reviewing it right now 👍

schell:
The reference implementation @ValarDragon pointed me to also "repurposes" the `toaddress` parameter to use as account selection, and then sends all UTXOs to the internal Orchard receiver.  

In the reference, the proposal is created by hand, without going through `propose_shielding`, and it contains no explicit payment outputs. The entire shielded output is created as change. 

After review I think there are two bugs in the linked reference impl. 

1. The reference takes coinbase maturity into account by only considering transparent outputs that have >= 100 confirmations, but it selects any transparent output after that, not just coinbase. Maybe this is a feature, not a bug :shrug: - but would be confusing for users, at best.
2. In the case of `fromaddress = *`, the reference selects outputs from all accounts, but it only signs the UTXOs with the account selected by toaddress - which isn't necessarily the account the UTXOs belong to. I'm not sure what happens at runtime, but I bet `create_proposed_transactions` fails unless there happens to be a match between the accounts. A match might be likely though, since the RPC call is being used as a replacement for the single-account `zcashd` wallet.

I do like the way the reference impl's pre-flight return value has the same shape as the original `zcashd` impl, so I think I'll implement that here.

I think the most important thing to make headway here is to get buy-in from @nuttycom on the patch branch (https://github.com/zcash/wallet/pull/428) to unblock testing for coinbase-only selection and then I'll work on the pre-flight return value and possibly `fromaddress` stuff. 

Then I think this would be ready for review.

schell:
@ValarDragon One thing @nullcopy pointed out internally is that if `fromaddress = *`, sweeping through all _accounts_ introduces a privacy leak. In the reference implementation you linked, the end result, (if it actually worked - unlikely because of the signing bug), would be a single transaction with transparent inputs from multiple accounts, signed and broadcast as one atomic on-chain event. 

But even sweeping each account and creating one transaction per account could associate the accounts to the wallet by timing, so I think my current approach of `fromaddress = *` being limited to the single account selected by `toaddress` is the correct path. I'll make a note to mention this privacy issue in the documentation, because it's still up to the user to ensure their own privacy here.

nuttycom:
> @ValarDragon One thing @nullcopy pointed out internally is that if `fromaddress = *`, sweeping through all _accounts_ introduces a privacy leak. In the reference implementation you linked, the end result, (if it actually worked - unlikely because of the signing bug), would be a single transaction with transparent inputs from multiple accounts, signed and broadcast as one atomic on-chain event.
> 
> But even sweeping each account and creating one transaction per account could associate the accounts to the wallet by timing, so I think my current approach of `fromaddress = *` being limited to the single account selected by `toaddress` is the correct path. I'll make a note to mention this privacy issue in the documentation, because it's still up to the user to ensure their own privacy here.

What's the current behavior of zcashd here? I think that your interpretation of what we should do here is correct, but I want to consider whether we should also deprecate `z_shieldcoinbase` at the same time as we do `z_sendmany`, and instead ensure that #217 takes into account the need for coinbase shielding; ideally, you'd never take a `to_address` argument because for coinbase shielding specifically you'd shield to the internal key within the account; for other coinbase spends, so long as the recipient is shielded it doesn't have to be an address internal to the wallet; the only requirement is that the shielding operation cannot produce transparent **change**. Let's document on #217 what we want the behavior to be for coinbase spends, and think about what the API should look like to enable that.

schell:
@nuttycom in `zcashd` the situation is much simpler since there's only one account. `toaddress` can be any address.

I think deprecation is a good idea.

ValarDragon:
> (1) The reference takes coinbase maturity into account by only considering transparent outputs that have >= 100 confirmations, but it selects any transparent output after that, not just coinbase. Maybe this is a feature, not a bug 🤷 > - but would be confusing for users, at best.
> (2) In the case of fromaddress = *, the reference selects outputs from all accounts, but it only signs the UTXOs with the account selected by toaddress - which isn't necessarily the account the UTXOs belong to. I'm not sure what happens at runtime, but I bet create_proposed_transactions fails unless there happens to be a match between the accounts. A match might be likely though, since the RPC call is being used as a replacement for the single-account zcashd wallet.

Pretty sure the guy mining it had a new address for mining, so maybe thats why (1) just worked for him. Definitely a bug.

On (2) I think its probably just replacing a single account wallet. So maybe we need to set the fromAddress?

The current behavior of mainnet mining pools as far as I understand is that they are really just shielded pool tx washing.

- coinbase. ( https://cipherscan.app/tx/cee5e0ae7ddaa19376e02045a78bee0d66fa5b6f831dc113e75755bc936217b1 , payout to this [addr](https://cipherscan.app/tx/cee5e0ae7ddaa19376e02045a78bee0d66fa5b6f831dc113e75755bc936217b1) )
- Coinbase t->z tx ( https://cipherscan.app/tx/2a5f036fa20c874f4e35ac18defbd45311e8e64176c01218878bb4a7800e2026 )
- z -> transparent mining pool operator addr: ( https://cipherscan.app/tx/4669ab48da529f07bf3404d5b9a15109422191a29fdaac3e254359f761530171 )
- transparent -> transparent pay out to everyone (https://cipherscan.app/tx/589998735f239e565c76cdc5ee725c74479b9230b0538643a5746901040ad6cd)

This mining pool is easy and just uses one addresses. All other mining pools I've looked at do some accumulation in shielded, and then do the shielded to transparent in a second address, then the t->t payout for everyone in mining pools

ValarDragon:
I'm unopinonated on eventual deprecation

I do think its worth getting this as an API where from makes sense. FromAccount makes sense to me as either:
* `fromAddress=*` allowed,  but making that _only_ spend your coinbase outputs across all addresses. (I think its a bad in principle, but will never happen in practice privacy leak. I don't think users will have multiple coinbase tx recipients on same device)
* `fromAddress` must be specified. I think thats easy enough for people to adapt to as well

Would be great if we can get it in, then start talking to the mining pools to move over. (To both Zebra + Orchard)

zallet/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -385,7 +386,7 @@ pub(crate) async fn call(
                             }
                             _ => LegacyCode::Database.with_message(e.to_string()),
                         })?;
-                    keys.insert(address, vec![secret_key]);
+                    keys.entry(address).or_default().push(secret_key);
```
schell: I had cargo culted this for the first iteration of `z_shieldcoinbase` and just realized we probably shouldn't be clobbering previously stored keys in either RPC call.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,546 @@
...
+use schemars::JsonSchema;
+use secrecy::ExposeSecret;
+use serde::Serialize;
+use transparent::address::TransparentAddress;
+use zaino_state::FetchServiceSubscriber;
+use zcash_address::unified;
+use zcash_client_backend::{
+    data_api::{
+        Account, InputSource, TransparentOutputFilter, WalletRead,
+        wallet::{
```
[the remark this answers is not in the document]
schell: `Account` and `InputSource` are definitely necessary.

`Application` is _not_ necessary - but it also isn't triggering a warning or a clippy lint, which is interesting.

zallet/src/components/sync.rs
```diff
@@ -533,21 +533,75 @@ pub(crate) async fn fetch_transparent_utxos(
...
+    // bundle reports `is_coinbase()`. The path is idempotent for non-coinbase txs
+    // (where `tx_index` stays NULL), and for already-recorded coinbase txs the
+    // upsert preserves the existing value.
+    //
+    // Fetch failures are logged and skipped rather than propagated so that a
+    // transient indexer error cannot wedge transparent sync; subsequent polls
+    // will retry.
+    for (txid, mined_height) in tx_heights {
+        let tx_obj = match chain.get_raw_transaction(txid.to_string(), Some(1)).await {
+            Ok(zebra_rpc::methods::GetRawTransaction::Object(tx_obj)) => tx_obj,
```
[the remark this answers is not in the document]
schell: We are now skipping transactions that have already been processed.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,383 @@
...
+            }
+        }
+        // Sapling addresses are always valid shielded destinations.
+        Address::Sapling(_) => {}
+    }
+
+    // Look up the account that owns the destination address.
+    let account = get_account_for_address(wallet.as_ref(), &to_address).map_err(|_| {
+        LegacyCode::InvalidAddressOrKey
+            .with_static("Invalid parameter: toaddress does not belong to this wallet.")
```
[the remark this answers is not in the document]
schell: The error returned by `get_account_for_address` is fine.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,383 @@
...
+        })?;
+
+        let transparent_addr = match from_address {
+            Address::Transparent(addr) => addr,
+            _ => {
+                return Err(LegacyCode::InvalidAddressOrKey.with_static(
+                    "Invalid from address: z_shieldcoinbase only supports transparent source addresses.",
+                ));
+            }
+        };
```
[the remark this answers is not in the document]
schell: I think if a user uses a TEX address here it would most likely be a mistake. Also, `zcashd` doesn't accept TEX addresses as `fromaddress`, if I understand correctly.

@nuttycom @nullcopy - care to chime in?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,383 @@
...
+                LegacyCode::WalletUnlockNeeded.with_message(e.to_string())
+            }
+            _ => LegacyCode::Database.with_message(e.to_string()),
+        })?;
+    let usk = UnifiedSpendingKey::from_seed(
+        wallet.params(),
+        seed.expose_secret(),
+        derivation.account_index(),
+    )
+    .map_err(|e| LegacyCode::InvalidAddressOrKey.with_message(e.to_string()))?;
```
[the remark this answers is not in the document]
schell: This might be useful, but I'm pretty sure `z_send_many` is going to get an overhaul, so might not be worth it.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,546 @@
...
+                            {
+                                LegacyCode::WalletUnlockNeeded.with_message(e.to_string())
+                            }
+                            _ => LegacyCode::Database.with_message(e.to_string()),
+                        })?;
+                    keys.entry(address).or_default().push(secret_key);
+                }
+            }
+        }
+        keys
```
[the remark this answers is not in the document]
schell: This was cargo culted from `z_send_many` so maybe we ship it and then circle back with a ticket. It doesn't seem to be a problem.

zallet/src/components/json_rpc/methods.rs
```diff
@@ -561,6 +563,50 @@ pub(crate) trait WalletRpc {
...
+    /// `toaddress`. The `fromaddress` must also belong to the same account.
+    ///
+    /// # Arguments
+    /// - `fromaddress` (string, required): A wallet-owned transparent address to
+    ///   sweep from, or `"*"` to sweep from all taddrs belonging to the same
+    ///   account as `toaddress`. Must belong to the same account as `toaddress`.
+    /// - `toaddress` (string, required): A wallet-owned shielded address (Sapling,
+    ///   Orchard, or Unified with shielded receivers) used to identify the account.
+    ///   Funds are shielded into the account's internal shielded address, which may
+    ///   differ from this address.
```
nullcopy: So, the address is used only for the purposes of selecting an account, but the address that receives funds may differ from the provided address? Why not provide an account UUID if all you need is to identify an account?

I doubt this would be a practical concern for anyone, but I can see it causing confusion if, e.g. a mining pool builds a tool that tracks specific addresses or something

zallet/src/components/json_rpc/methods.rs
```diff
@@ -561,6 +563,50 @@ pub(crate) trait WalletRpc {
...
+    /// - `fee` (optional): This parameter must be omitted or set to `null`;
```
nullcopy: Why have this argument if it's required unused?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,381 @@
...
+
+    // TODO(schell): `propose_shielding` does not accept a memo parameter. The memo is
+    // accepted here for API compatibility but is currently ignored. Once the backend
+    // supports attaching a memo to shielding transactions, wire it through.
+    let _memo = memo;
+
+    // TODO(schell): `propose_shielding` does not support a UTXO limit parameter. The
+    // limit is accepted here for API compatibility but is currently ignored. Consider
+    // pre-filtering UTXOs or extending the backend API to support this.
+    let _limit = limit;
```
nullcopy: We should have an issue to track this TODO

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,381 @@
...
+        Some(s) => PrivacyPolicy::from_str(s).ok_or_else(|| {
+            LegacyCode::InvalidParameter.with_message(format!("Unknown privacy policy {s}"))
+        }),
+        None => Ok(PrivacyPolicy::AllowRevealedSenders),
+    }?;
+
+    // TODO(schell): `propose_shielding` does not accept a memo parameter. The memo is
+    // accepted here for API compatibility but is currently ignored. Once the backend
+    // supports attaching a memo to shielding transactions, wire it through.
+    let _memo = memo;
```
nullcopy: We should have an issue to track this TODO

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,381 @@
...
+
+    // Parse the privacy policy.
+    // Default to AllowRevealedSenders since shielding always reveals the transparent sender.
+    let privacy_policy = match privacy_policy.as_deref() {
+        Some("LegacyCompat") => Err(LegacyCode::InvalidParameter
+            .with_static("LegacyCompat privacy policy is unsupported in Zallet")),
+        Some(s) => PrivacyPolicy::from_str(s).ok_or_else(|| {
+            LegacyCode::InvalidParameter.with_message(format!("Unknown privacy policy {s}"))
+        }),
+        None => Ok(PrivacyPolicy::AllowRevealedSenders),
```
nullcopy: It seems like `AllowRevealedSenders` is the only policy we support or ever plan to support. Is this correct? If so, why even have a `privacy_policy` arg in the API?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    // Look up the account that owns the destination address.
+    let account = get_account_for_address(wallet.as_ref(), &to_address)?;
+
+    // Resolve the transparent source addresses.
+    // TODO(schell): When fromaddress is "*", we currently only sweep transparent
+    // addresses belonging to the same account as toaddress. Check with teammates
+    // whether "*" should sweep across all wallet accounts instead (matching zcashd's
+    // single-keypool model more closely).
```
nullcopy: Where did you land on this question. IMO sweeping should be restricted to account scope. Sweeping across accounts would not only reduce privacy, but may limit mining software, e.g. if a pool wants to operate multiple mining cohorts from a single Zallet instance.

CHANGELOG.md
```diff
@@ -33,6 +34,20 @@ be considered breaking changes.
   standalone imported transparent keys (e.g. from a `zcashd` migration).
 - No longer crashes in regtest mode when a Sapling or NU5 activation height is
   not defined.
+- `z_sendmany` and `z_shieldcoinbase` no longer drop standalone transparent
```
nullcopy: `z_shieldcoinbase` did not exist before this PR, so I don't think it should show up in these three fixes

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,541 @@
...
+/// Mirrors the JSON object returned by `zcashd`'s `z_shieldcoinbase`:
+/// `{ remainingUTXOs, remainingValue, shieldingUTXOs, shieldingValue, opid }`.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct ShieldCoinbaseResult {
+    /// Number of coinbase UTXOs that were eligible for shielding but were not
+    /// selected by this operation.
+    ///
+    /// Note: Zallet currently ignores the `limit` parameter, so in practice
+    /// this is `0` whenever the proposal succeeded. The field is preserved
+    /// for compatibility with `zcashd`-shape clients.
```
nullcopy: This can't always be zero, right? What if the account holds more UTXOs than can be spent in a single transaction?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,541 @@
...
+        DustOutputPolicy::default(),
+        APP.config().note_management.split_policy(),
+    );
+
+    let input_selector = GreedyInputSelector::new();
+
+    // Create the shielding proposal. Uses Zatoshis::ZERO as the shielding threshold
+    // to shield all available coinbase UTXOs. Passes TransparentOutputFilter::CoinbaseOnly
+    // to ensure only coinbase UTXOs are selected for shielding.
+    let proposal = propose_shielding::<_, _, _, _, Infallible>(
```
nullcopy: What happens if the number of UTXOs exceeds the allowable transaction size? Does `propose_shielding` error? Or proceed with as many as it could?

I'm willing to bet mining pools have this problem.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,541 @@
...
+            .filter(|pool| pool == &&PoolType::ORCHARD)
+            .count()
+            + step
+                .balance()
+                .proposed_change()
+                .iter()
+                .filter(|change| change.output_pool() == PoolType::ORCHARD)
+                .count();
+
+        let orchard_actions = orchard_spends.max(orchard_outputs);
```
nullcopy: What is the scenario where orchard spends would be anything other than zero?

nullcopy (changes requested):
Several of my comments circle around consistency with zcashd's implementation of this method. Notably:
* `toaddress` is used for something other than its originally intended purpose.
* `fee` is an option that must not be provided
* `privacy_policy` is not something we ever plan to be configurable.

The `toaddress` one is the biggest sticking point, IMO. Any tool that was written to use zcashd's implementation of this method now will need to be reviewed to ensure it didn't _expect_ `toaddress` to be funded after shielding. IMO it would be better to shape the API to what we actually want from it, and force consumers to explicitly review how their tools use this method, since it would be unwise to just keep using it assuming it's the same as zcashd.

Reviewed at e5fc378bebb1df228f768cbc5128195aa9543d62

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
schell: I netted out at your opinion. Sweeping accounts introduces  a correlation even if we execute separate transactions per account. I'll remove the TODO 🙇

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
ValarDragon: Agreed, this should be fixed, though perhaps we do it as a follow-up? Most mining pools don't operate near this limit from what I can tell. (Largest number of combines I've seen via manual inspection is ~25). Im sure bigger must exist, and this will get ran into though

nullcopy: I'm fine with deferring to a later fix. Let's be sure to document the limitation and throw up an issue to track the fix

zallet/src/components/json_rpc/methods.rs
schell: I've opened [librustzcash#2353](https://github.com/zcash/librustzcash/pull/2353) to address this. With that PR we'll be able to use `toaddress` as the actual address that receives the funds. It will also still be used to select the account.

zallet/src/components/json_rpc/methods.rs
schell: Well, it's not required - they can omit the parameter as it is optional. There's two calling contexts here - positional and named. 

In the positional case, removing this parameter entirely causes the other positions to shift left, so if they make a call to `z_shieldcoinbase` using the old convention and provide a value for `fee`, the values would be bound incorrectly. That _may_ be easy to figure out if the values happen not to serialize correctly or are invalid, but that's not great.

In the named case I think `jsonrpsee` would just ignore the extra field. But that's also not great, because then the user doesn't get feedback that the parameter is invalid when non-null.

The current `z_send_many` set the precedent here, and if we're trying to keep to the old interface, I think it makes sens.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
schell: [librustzcash#2353](https://github.com/zcash/librustzcash/pull/2353) adds the optional memo to `propose_shielding`.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
schell: [librustzcash#2353](https://github.com/zcash/librustzcash/pull/2353) adds the optional limit to `propose_shielding_coinbase`.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
schell: Now we have `propose_shielding_coinbase`, which contains the optional memo param.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
schell: The only reason I see to keep the arg is to provide the same interface as `zcashd`.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
schell: It would proceed with all the UTXOs and it would be rejected at broadcast time, most likely. We could set an implicit limit, but I think it might be better to simply log a warning, let the transaction fail, and then the user could set an explicit `limit`, now that the param is available and respected.

zallet/src/components/json_rpc/methods.rs
```diff
@@ -561,6 +563,65 @@ pub(crate) trait WalletRpc {
...
+    /// **Note:** This method's behavior differs from `zcashd`. The `toaddress`
+    /// must be a wallet-owned shielded address and is used to select the account
+    /// whose transparent UTXOs will be shielded. The resulting transaction sends
+    /// funds to the account's internal shielded address, which may differ from
+    /// `toaddress`. The `fromaddress` must also belong to the same account.
+    ///
+    /// # Arguments
+    /// - `fromaddress` (string, required): A wallet-owned transparent address to
+    ///   sweep from, or `"*"` to sweep from all taddrs belonging to the same
+    ///   account as `toaddress`. Must belong to the same account as `toaddress`.
```
nullcopy: `toaddress` is not necessarily controlled by this wallet, so it does not make sense to require `fromaddress` to be in the same account as `toaddress`

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,508 @@
...
+        LegacyCode::InvalidParameter.with_message(format!(
+            "Invalid parameter, unknown address format: {toaddress}"
+        ))
+    })?;
+
+    // Look up the account that owns the destination address. This enforces the
+    // zallet-level "same account" requirement; the backend additionally
+    // enforces "must be a shielded address" via
+    // `ProposalError::ShieldingRequiresShieldedRecipient`.
+    let account = get_account_for_address(wallet.as_ref(), &to_address)?;
```
nullcopy: This fails, because `toaddress` need not be owned by our account

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,508 @@
...
+    //   anchor the zallet-level same-account constraint: zallet's
+    //   `z_shieldcoinbase` requires `toaddress` to belong to a wallet account
+    //   (matching `zcashd`'s behavior), on top of the backend's requirement
+    //   that it be shielded.
+    let to_address = Address::decode(wallet.params(), &toaddress).ok_or_else(|| {
+        LegacyCode::InvalidParameter.with_message(format!(
+            "Invalid parameter, unknown address format: {toaddress}"
+        ))
+    })?;
+    let to_zcash_address: ZcashAddress = toaddress.parse().map_err(|_| {
```
nullcopy, line 169: Do we really need to parse `toaddress` into two more representations of the same data?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,508 @@
...
+            _ => {
+                return Err(LegacyCode::InvalidAddressOrKey.with_static(
+                    "Invalid from address: z_shieldcoinbase only supports transparent source addresses.",
+                ));
+            }
+        };
+
+        // Verify the transparent address belongs to the same account as toaddress.
+        let from_account =
+            get_account_for_address(wallet.as_ref(), &Address::Transparent(transparent_addr))?;
```
nullcopy: This check does not make sense. The `toaddress` may be an address not owned by any account in our wallet.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: This won't work, because `to_address` need not belong to our wallet

CHANGELOG.md
```diff
@@ -35,6 +36,20 @@ be considered breaking changes.
   not defined.
 - Zallet now refuses to open wallet databases from incompatible earlier alpha
   releases instead of attempting to migrate them.
+- `z_sendmany` and `z_shieldcoinbase` no longer drop standalone transparent
```
nullcopy: My [original feedback](https://github.com/zcash/wallet/pull/402/changes#r3210032644) here is still unresolved: `z_shieldcoinbase` did not exist before this PR, so it doesn't make sense to have fixes for it in the CHANGELOG

nullcopy (changes requested):
This is a lot closer, but there's still some things that need to change. Namely around the `toaddress`/`fromaddresses` relationship.

NACK 0343ee7c14e94c30342f6e46561b75b94bb58053

zallet/src/components/json_rpc/methods.rs
```diff
@@ -561,6 +563,65 @@ pub(crate) trait WalletRpc {
...
+    /// - `toaddress` (string, required): A wallet-owned shielded address (Sapling,
+    ///   Orchard, or Unified with shielded receivers) used to identify the account.
```
nullcopy: It cannot be used to identify the account, because it may not be owned by any account in our wallet

nullcopy:
I've updated this PR with the following:
* rebase onto main
* pushed [dbf1e0964dfd3dea7f43e50dea33fdaadee190f6](https://github.com/zcash/wallet/pull/402/commits/6c6c1485eac75c596806ca6ea26c165c2725ae4e) to address review feedback
* pushed [3c20d92e523225d81cb04e378009b7a4ed95727b](https://github.com/zcash/wallet/pull/402/commits/85caee08f5028f752c066bf6a8d384f4a4300641) to target librustzcash with [bug-fix](https://github.com/zcash/librustzcash/pull/2376) for propose_shielding_coinbase fee calculation error
* push https://github.com/zcash/wallet/pull/402/commits/4441e01da70c4c82f89d21cf394cc18a2242807a to fix bug in `z_viewtransaction` which was preventing integration tests from validating this PR

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/4441e01da70c4c82f89d21cf394cc18a2242807a..2e0b6920b3fce258217b435c7db2d23985a7a7f3) to fix rust-fmt complaints

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,504 @@
...
+    .map_err(|e| {
+        LegacyCode::Wallet.with_message(format!("Failed to propose shielding transaction: {e}"))
+    })?;
+
+    // The shielding operation always reveals the transparent sender(s). No
+    // other privacy weakening is required, and we deliberately don't allow
+    // callers to opt into it.
+    crate::components::json_rpc::payments::enforce_privacy_policy(
+        &proposal,
+        PrivacyPolicy::AllowRevealedSenders,
```
czarcas7ic: I think this rejects a valid `fromaddresses` case. If the caller passes an account UUID, we expand it to all transparent receivers for that account:

https://github.com/zcash/wallet/blob/2e0b6920b3fce258217b435c7db2d23985a7a7f3/zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs#L313-L318

This also matches the RPC docs, which says the UUID form sweeps coinbase UTXOs across the account’s transparent receivers:

https://github.com/zcash/wallet/blob/2e0b6920b3fce258217b435c7db2d23985a7a7f3/zallet/src/components/json_rpc/methods.rs#L574-L577

If the selected UTXOs come from more than one t-address, the shared privacy check treats that as linking account addresses and rejects unless `AllowLinkingAccountAddresses` is allowed:

https://github.com/zcash/wallet/blob/2e0b6920b3fce258217b435c7db2d23985a7a7f3/zallet/src/components/json_rpc/payments.rs#L240-L246

So an account with coinbase UTXOs on two t-addresses is initially accepted, and then fail here with `LinkingAccountAddresses`. If we actually want account wide sweeping, I think this needs `AllowLinkingAccountAddresses`. Otherwise we should reject multiple source addresses before building the proposal.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,504 @@
...
+
+    // Pre-flight numerics.
+    let (shielding_utxos, shielding_value_zats) = sum_selected_inputs(&proposal)?;
+    let target_height = proposal.min_target_height();
+    let (total_utxos, total_value_zats) = enumerate_eligible(
+        wallet.as_mut(),
+        &from_addrs,
+        target_height,
+        confirmations_policy,
+    )?;
```
czarcas7ic: I think there is an edge case that can make the preflight totals disagree with the actual proposal. The proposal path selects coinbase UTXOs with `ConfirmationsPolicy::MIN` in [librustzcash](https://github.com/nullcopy/librustzcash/blob/e41d4aa9a84cef5d89aac894974c81e1469db2dc/zcash_client_backend/src/data_api/wallet/input_selection.rs#L1219-L1227), but the preflight enumerate function here uses Zallet’s configured confirmation policy:

https://github.com/zcash/wallet/blob/2e0b6920b3fce258217b435c7db2d23985a7a7f3/zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs#L197-L202

https://github.com/zcash/wallet/blob/2e0b6920b3fce258217b435c7db2d23985a7a7f3/zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs#L400-L415

With defaults I don't think this causes issues, but with a stricter config like `untrusted_confirmations > 100`, the proposal can select UTXOs that preflight does not count and then hit the internal accounting error below. I think selection and preflight should use the same eligibility rule.

To be clear I don't think this is blocking, just might save someone from annoying debugging down the line.

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/2e0b6920b3fce258217b435c7db2d23985a7a7f3..9ac9bfa2981134c626fc2489ec1ee0aed45433ad) to address feedback and update librustzcash crate patch

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: This was a good catch. I added a new test here: https://github.com/zcash/integration-tests/pull/95

That test confirmed the bug, which I've now fixed in this PR. Thank you!

zallet/src/components/json_rpc/methods.rs
```diff
@@ -561,6 +563,49 @@ pub(crate) trait WalletRpc {
...
+    /// - `remainingValue` (numeric, ZEC): Total value of those UTXOs.
+    /// - `shieldingUTXOs` (numeric): Number of coinbase UTXOs being
+    ///   shielded by this operation.
+    /// - `shieldingValue` (numeric, ZEC): Total value being shielded.
+    /// - `opid` (string): Operation id.
+    #[method(name = "z_shieldcoinbase")]
+    async fn z_shieldcoinbase(
+        &self,
+        fromaddresses: JsonValue,
+        toaddress: String,
```
nuttycom, line 630: For API compatibility, we *must* include the positional "fee" field, but it can be required to be set to `null`.

nuttycom: I think maybe we also need to include the `privacyPolicy` field, although I'm less concerned about it.

nullcopy: Ah, I didn't realize that was a requirement. In that case, it's probably also not ok that I expanded `fromaddresses` to accept an array instead of just a single address.

I've added these fields back, and reverted `fromaddress` to accept either a single t-addr or an account UUID, which should be API compatible.

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/9ac9bfa2981134c626fc2489ec1ee0aed45433ad..6bd5ff1e6651c79bebce0684477fdce54eae4dd5) to address feedback from @nuttycom and add unit tests for the helpers this API relies on.

zallet/src/components/json_rpc/methods/view_transaction.rs
nuttycom, line 351: Let's not forget about this TODO; I don't like having SQL in this codebase. Can you check whether there's an issue for this follow-up?

zallet/src/components/sync.rs
```diff
@@ -533,21 +534,113 @@ pub(crate) async fn fetch_transparent_utxos(
...
     for utxo in utxos {
         let (address, txid, index, script, value_zat, mined_height) = utxo.into_parts();
         debug!("{address} has UTXO in tx {txid} at index {}", index.index());
 
+        let mined_height = BlockHeight::from_u32(mined_height.0);
+        // We pass `None` for `recipient_account`, `recipient_key_scope`, and
+        // `funding_account` here: the sqlite layer's `put_received_transparent_utxo`
+        // re-derives the recipient account/scope from the wallet's address->account
+        // mapping using the UTXO's recipient address, and the funding account is
+        // not known at sync ingest time.
```
nuttycom: I'm skeptical of the justification here. However, I think that this will be resolved once we land the rest of str4d's scanning PRs, so I won't block on it now.

zallet/src/components/sync.rs
```diff
@@ -533,21 +534,113 @@ pub(crate) async fn fetch_transparent_utxos(
...
+                )
+                .optional()
+            })
+            .map_err(zcash_client_sqlite::error::SqliteClientError::from)?;
+        if tx_index_known.unwrap_or(false) {
+            continue;
+        }
+        let tx_obj = match chain.get_raw_transaction(txid.to_string(), Some(1)).await {
+            Ok(zebra_rpc::methods::GetRawTransaction::Object(tx_obj)) => tx_obj,
+            Ok(zebra_rpc::methods::GetRawTransaction::Raw(_)) => {
```
nuttycom, line 592: I don't fully understand the difference between Object and Raw here. Under what circumstances is each received? This API is kind of mysterious.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+
+pub(crate) type ResultType = ShieldCoinbaseResult;
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// Pre-flight numeric fields, computed before the async portion runs.
+///
+/// Held as a separate type so that the [`OperationId`] (only available after
+/// the async operation is registered) can be joined with these values to
+/// produce the final [`ShieldCoinbaseResult`].
+pub(crate) struct Preflight {
```
nuttycom, line 104: I'm surprised this needs to be pub(crate)/pub(super).

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+    // non-null value so callers can't silently believe they're configuring
+    // either.
+    if fee.is_some() {
+        return Err(LegacyCode::InvalidParameter
+            .with_static("Zallet always calculates fees internally; the fee field must be null."));
+    }
+    if privacy_policy.is_some() {
+        return Err(LegacyCode::InvalidParameter.with_static(
+            "Zallet does not support configuring the privacy policy for z_shieldcoinbase; \
+             the privacy_policy field must be null.",
```
nuttycom:
```suggestion
             the privacy_policy argument must be absent or the field set to null.",
```

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+        ))
+    })?;
+
+    // Parse the memo parameter (hex-encoded).
+    let memo = memo.as_deref().map(parse_memo).transpose()?;
+    let limit_usize = limit.map(|n| n as usize);
+
+    // Resolve `fromaddress` to the source account + its source transparent
+    // addresses (one address for a t-addr input, all account receivers for a
+    // UUID input).
```
nuttycom, line 186: Was the addition of the account UUID something that was specified in the issue, as a replacement for the `*` that zcashd used?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+        LegacyCode::InvalidAddressOrKey.with_message(format!(
+            "No payment source found for account {}.",
+            account_id.expose_uuid(),
+        ))
+    })?;
+    let seed = keystore
+        .decrypt_seed(derivation.seed_fingerprint())
+        .await
+        .map_err(|e| match e.kind() {
+            crate::error::ErrorKind::Generic if e.to_string() == "Wallet is locked" => {
```
nuttycom, line 291: The fact that we're having to match on a string here says that we have our error types wrong in the keystore. Add an issue for follow-up.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+    )
+    .map_err(|e| {
+        LegacyCode::Wallet.with_message(format!("Failed to propose shielding transaction: {e}"))
+    })?;
+
+    // The shielding operation always reveals the transparent sender(s). When the caller asks us to
+    // sweep multiple source addresses (i.e. via an account UUID expanded to >1 receivers), the
+    // resulting transaction also links those addresses on-chain, so we must additionally allow
+    // that.
+    crate::components::json_rpc::payments::enforce_privacy_policy(
```
nuttycom: Hmm, I had forgotten that we had actually implemented privacy policies for z_sendmany. Given that we have this, we should probably parse and use the privacy policy that was passed in.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+        &proposal,
+        PrivacyPolicy::AllowLinkingAccountAddresses,
+    )?;
+
+    // Pre-flight numerics
+    let (shielding_utxos, shielding_value_zats) = sum_selected_inputs(&proposal)?;
+    let target_height = proposal.min_target_height();
+    let (total_utxos, total_value_zats) =
+        enumerate_eligible(wallet.as_mut(), &from_addrs, target_height)?;
```
nuttycom, line 245: I'm confused by what's happening here. Why are we doing UTXO selection again here, separate from the proposal? The "pre-flight numerics" comment is not at all clarifying. 

EDIT: I see, this is basically just for the computation of the remaining utxos value. Please add some comments to clarify here.

nuttycom: Ideally what we'd do here is get note/utxo locking merged, and then we wouldn't have to do these subtractions; we'd just get unlocked notes from the query. That can be follow-up though. Add a task to fix this up after locking is implemented, and have it blocked by https://github.com/zcash/librustzcash/issues/2161

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,601 @@
...
+        return Ok(FromAddressInput::AccountUuid(uuid));
+    }
+    match Address::decode(params, fromaddress) {
+        Some(Address::Transparent(addr)) => Ok(FromAddressInput::TransparentAddress(addr)),
+        Some(_) => Err(LegacyCode::InvalidAddressOrKey.with_message(format!(
+            "Invalid `fromaddress`: only transparent addresses are accepted (got a \
+             non-transparent Zcash address): {fromaddress}",
+        ))),
+        None => Err(LegacyCode::InvalidParameter.with_message(format!(
+            "Invalid `fromaddress`: expected a wallet-owned transparent address or an \
```
nuttycom, line 374: Existing callers that use `*` will get this error. We might want to handle `*` explicitly, instead of falling into this bucket.

nuttycom (approved):
utACK; several comments but none are necessarily blocking. Please check on the difference between the GetRawTransaction result types before merging though.

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: It was not specified in the issue, but the issue didn't specify anything.

Without an account UUID, we cannot know which account to sweep from (we decided not to allow sweeping across accounts). The alternative I considered was to add a second optional 'fromaccount' field to be mutually exclusive to `fromaddress`, and using that in place of wildcard addresses. The approach in code today felt closest to the spirit of the original API, but I'm not married to this if you feel another approach would be better.

zallet/src/components/json_rpc/methods/view_transaction.rs
nullcopy: I couldn't find one, so I created one now: https://github.com/zcash/librustzcash/issues/2388

zallet/src/components/sync.rs
nullcopy: The docs at `get_raw_transaction` just link to the zcashd docs: https://zcash.github.io/rpc/getrawtransaction.html

According to those, `Raw` is returned when `verbose=0`, and `Object` is returned when `verbose=1`. Since we pass `verbose=1`, we shouldn't receive a raw response. Instead of warning, perhaps we should fail loudly, as this is a sign the node RPC is misbehaving according to our expectations?

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: `Preflight` is returned by `pub(crate) call()`, so I matched it's visibility

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: Agreed: https://github.com/zcash/wallet/issues/448

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: See https://github.com/zcash/wallet/pull/402#discussion_r3318651240

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: Sure, I can do that. The only two allowed privacy policies will be `AllowLinkingAccountAddresses` and `AllowRevealedSenders` though

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
nullcopy: Agreed. I will add an explicit error for this.

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/6bd5ff1e6651c79bebce0684477fdce54eae4dd5..9818694deeb2bba57fe3cfe4557025e423391131) to address feedback from @nuttycom

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,735 @@
...
+        )
+    })?;
+
+    if shielding_utxos > COINBASE_INPUTS_WARN_THRESHOLD {
+        warn!(
+            "z_shieldcoinbase: proposal selected {} coinbase UTXOs, which exceeds the \
+             soft warning threshold of {}. The resulting transaction may exceed \
+             network/mempool size limits at broadcast time. If broadcast fails, retry \
+             with a `limit` parameter to shield in smaller batches.",
+            shielding_utxos, COINBASE_INPUTS_WARN_THRESHOLD,
```
daira, line 269: This warning is only useful if the `limit` parameter is not already given.

zallet/src/components/sync.rs
```diff
@@ -533,21 +534,113 @@ pub(crate) async fn fetch_transparent_utxos(
...
+        // Instead, check `tx_index` directly: if it's already non-NULL, the
+        // tx has been enhanced and either is or isn't coinbase definitively.
+        // Reorgs reset `tx_index` to NULL, so re-confirmation correctly re-runs
+        // the second pass.
+        //
+        // The downside is that non-coinbase transparent-only txs (whose
+        // `tx_index` stays NULL) are re-fetched on every poll. This is
+        // acceptable: in steady state most observed UTXOs are coinbase
+        // (mining) or already-enhanced (previously seen), and a failed
+        // re-fetch is logged-and-skipped so it can't wedge sync.
```
daira: This is the kind of argument that breaks if there is a DoS attack.

CHANGELOG.md
```diff
@@ -35,6 +36,20 @@ be considered breaking changes.
   not defined.
 - Zallet now refuses to open wallet databases from incompatible earlier alpha
   releases instead of attempting to migrate them.
+- `z_sendmany` no longer drop standalone transparent signing keys when the same
+  address backs multiple proposal inputs. Keys are now accumulated per address
+  rather than overwritten.
+- Transparent UTXO ingestion now records `tx_index` for coinbase transactions
+  by routing each observed transaction through `decrypt_and_store_transaction`
+  in addition to `put_received_transparent_utxo`. This enables
```
daira, line 44: This is a significant change; have we thought through all the consequences?

daira (approved):
utACK

CHANGELOG.md
nullcopy: I reviewed this change on @schell's initial addition, and I couldn't come up with a reason against it. I have less context here than you or @nuttycom though, so I would appreciate it if you can think of possible implications I should investigate more deeply

nullcopy: You had a good point about a DoS vector in this process. I've added a mitigation for that.

zallet/src/components/sync.rs
nullcopy: Good point. I've gone ahead and added a `raw IS NOT NULL` gate to prevent re-fetching transactions who've been processed once (`raw` is always written on any prior `put_tx_data`)

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/1f11140661b2c7fdfc61ae6b3497186b1149eab5..b41a29647c206ebe8edb85f5b7ae926d2b01fbe0) to address feedback from @daira

zallet/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -0,0 +1,735 @@
...
+        assert!(
+            msg.contains("account UUID"),
+            "error message should direct the caller to use an account UUID, got: {msg}",
+        );
+    }
+
+    fn taddr_input() -> FromAddressInput {
+        match parse_fromaddress(&mainnet(), MAINNET_P2PKH).unwrap() {
+            input @ FromAddressInput::TransparentAddress(_) => input,
+            other => panic!("Expected TransparentAddress input, got {other:?}"),
```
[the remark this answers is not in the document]
nuttycom: Account UUIDS are not sensitive data.

zallet/src/components/json_rpc/methods.rs
```diff
@@ -561,6 +563,76 @@ pub(crate) trait WalletRpc {
...
+    ///   sweep across multiple receivers. Unlike `zcashd`, the wildcard
+    ///   `"*"` (sweep all wallet t-addrs) is rejected with an
+    ///   `InvalidParameter` error: cross-account sweeps would correlate
+    ///   otherwise-unrelated accounts on-chain, so callers must scope the
+    ///   sweep to a single account by passing its UUID.
+    /// - `toaddress`: Any Zcash shielded address (Sapling, Orchard, or
+    ///   Unified with a shielded receiver) that will receive the shielded
+    ///   funds. Need not belong to this wallet. Transparent / TEX
+    ///   destinations are rejected by the backend.
+    /// - `fee` (numeric, optional): If set, it must be null. Zallet always
```
dannywillems, line 587: > If set, it must be null

The implementation is the following:
```
    // `fee` exists for positional compatibility with `zcashd`'s
    // `z_shieldcoinbase` only — Zallet always computes the fee internally.
    // Reject any non-null value so callers can't silently believe they're
    // configuring it.
    if fee.is_some() {
        return Err(LegacyCode::InvalidParameter
            .with_static("Zallet always calculates fees internally; the fee field must be null."));
    }
```

Therefore, it must always be `None`; it cannot be set to any value.

dannywillems (approved):
tACK b41a29647c206ebe8edb85f5b7ae926d2b01fbe0, with the following suggestions https://github.com/zcash/wallet/pull/453.

I also considered the wallet being in alpha, therefore not ready for production. There will probably be other changes coming, and I approve to unblock future work.

This has been tested with the patch https://github.com/zcash/integration-tests/pull/98.

zallet/src/components/json_rpc/methods.rs
nullcopy: I'm not sure I understand the question.

We leave the fee argument in the API, to be backwards compatible with z_shieldcoinbase as implemented in zcashd, but if the caller attempts to specify a fee this check provides the caller with a helpful error explaining that Zallet does not allow fee customization

dannywillems: The documentation mentions:
> If set, it must be null

The parameter `fee` is of type `Option<JsonValue>`.
By reading the documentation, I understand that either `None` or `Some(JsValue(0))` - i.e. `if set, it must be null`.
However, there is only one value that can be accepted without raising an exception, precisely `None`.

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/b41a29647c206ebe8edb85f5b7ae926d2b01fbe0..2509285d267d1c4f54a5aa2adf62774778b80d4c) to apply suggestions from @dannywillems

schell:
utACK [bea08cc](https://github.com/zcash/wallet/pull/402/commits/bea08cc063157f2584842812b4a3cd60a879db7f)

dannywillems (approved):
utACK 7916aeccebe40a58b0d8229f023b662f5380d690

nullcopy (approved):
utACK 7916aeccebe40a58b0d8229f023b662f5380d690

PR #406, opened 2026-03-27
feat(rpc): Implement z_importaddress

nullcopy:
This PR implements `z_importaddress` for standalone P2PKH and P2SH address import. I didn't see an existing implementation of `z_importaddress` in zcashd, so I tried to use my best judgement when designing it here.

The commit patches librustzcash to use [the latest commit](https://github.com/zcash/librustzcash/commit/62ee526f83f84c01d26bd99325131e4b22c18205), which supports the new standalone P2SH address import API.

Supercedes https://github.com/zcash/wallet/pull/399 
Resolves https://github.com/zcash/wallet/issues/272
Resolves COR-326

### DIFF
--- a/zallet/src/components/sync.rs
+++ b/zallet/src/components/sync.rs
@@ -154,7 +154,6 @@ impl WalletSync {
         let poll_transparent_task = crate::spawn!("Poll transparent", async move {
             poll_transparent(
                 chain_subscriber,
-                &params,
                 db_data.as_mut(),
                 poll_tip_change_signal_receiver,
             )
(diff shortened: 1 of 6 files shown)


zallet/src/components/sync.rs
```diff
@@ -503,13 +502,63 @@ async fn recover_history(
...
+    db_data: &mut DbConnection,
+) -> Result<(), SyncError> {
+    let params = db_data.params();
+
+    // Collect all of the wallet's non-ephemeral transparent addresses.
+    //
+    // TODO: This is likely to be append-only unless we add support for removing an
+    // account from the wallet, so we could implement a more efficient strategy here
+    // with some changes to the `WalletRead` API. For now this is fine.
+    let addresses = db_data
```
nuttycom: Compare this to the operation of https://github.com/zcash/zcash-devtool/blob/bf89be602ce40ddff5310b4eee758cc9bffe894e/src/commands/wallet/enhance.rs#L96. This manual reimplementation is fine, but I wonder if it won't potentially miss some of the graph traversal effects we get from the transparent data retrieval queue. I guess that this will be replaced once we have transparent scanning, though, so no big deal for right now.

zallet/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -380,7 +380,7 @@ pub(crate) async fn call(
                             }
                             _ => LegacyCode::Database.with_message(e.to_string()),
                         })?;
-                    keys.insert(address, secret_key);
+                    keys.insert(address, vec![secret_key]);
```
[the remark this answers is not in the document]
nullcopy: Good point. I'd considered this logically independent, but didn't consider it needed the patch changes.

I'll pull the patch changes into their own commit, and then squash the logical changes together

zallet/src/components/sync.rs
nullcopy: Agreed, a retrieval queue would be better. I'll leave it as is for now, since this will be replaced soon anyways.

PR #420, opened 2026-04-04
cli: Add --no-scan flag to migrate-zcashd-wallet

### LINKED ISSUE #414: Unable to migrate from zcashd without Zebra connection
Currently, to migrate a zcashd wallet into Zallet, an active connection to Zebra is required. IMO it should be possible (perhaps as a non-default option) to migrate just the key material and defer the scan/recovery process.

To solve this, I propose adding a `--keys-only` option to the `migrate-zcashd-wallet` command, which performs the import without scanning.

---
*Created by nullcopy via [Dagny](https://dagny.co)*

nullcopy:
Allows importing only keys and accounts without querying the chain for transaction data or tree state.

Resolves: https://github.com/zcash/wallet/issues/414
Resolves: [COR-1178](https://linear.app/zodl/issue/COR-1178/unable-to-migrate-from-zcashd-without-zebra-connection)

### DIFF
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -16,7 +16,7 @@ use zaino_state::{FetchServiceError, LightWalletIndexer, ZcashIndexer};
 use zcash_client_backend::{
     data_api::{
         Account as _, AccountBirthday, AccountPurpose, AccountSource, WalletRead, WalletWrite as _,
-        Zip32Derivation,
+        Zip32Derivation, chain::ChainState,
     },
     decrypt_transaction,
 };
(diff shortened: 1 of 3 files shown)


daira (approved):
utACK (cursory review).

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -320,21 +333,43 @@ impl MigrateZcashdWalletCmd {
...
-        )
-        .await?;
+                .map(|h| BlockHeight::from_u32(h.saturating_sub(1000)))
+                .map(|h| std::cmp::max(h, sapling_activation))
+                .unwrap_or(sapling_activation);
+
+            AccountBirthday::from_parts(
+                ChainState::empty(birthday_height, BlockHash([0; 32])),
+                None,
+            )
```
nuttycom, line 371: This is going to mean rescanning all of sandblasting, but since it's an explicit opt-out (and mostly used for regtest) it's fine.

nuttycom (commented):
post-hoc utACK

PR #422, opened 2026-04-07
fix: Use available transaction data to estimate exposure heights

nullcopy:
In --no-scan mode, main_chain_block_heights was left empty because resolving block hashes to heights requires a chain subscriber. This caused all mined transactions to be treated as mempool transactions, preventing update_gap_limits from setting exposed_at_height for addresses. As a result, listaddresses only showed the first address per account.

Use transaction expiry heights (available in wallet.dat) as approximate block heights for mined transactions. The expiry height is always in the same consensus epoch as the mined height (enforced by zcashd), so it produces correct BranchIds and marks transactions as mined. For pre-Overwinter transactions with zero expiry, fall back to the wallet birthday height.

Resolves: https://github.com/zcash/wallet/issues/421
Resolves: [COR-1182](https://linear.app/zodl/issue/COR-1182/bug-zcashd-wallet-migration-transaction-buffer-does-not-update-key)

### DIFF
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -311,6 +311,61 @@ impl MigrateZcashdWalletCmd {
                     }
                 }
             }
+        } else {
+            // In no-scan mode, we don't have a chain subscriber to resolve block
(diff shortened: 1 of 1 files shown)


nullcopy:
Superceded by https://github.com/zcash/wallet/pull/423

PR #423, opened 2026-04-07
fix: Use zcashd transaction data to estimate address exposures

nullcopy:
This PR augments the `migrate-zcashd-wallet` command's `--no-scan` mode to use transaction data stored in the wallet to infer address exposure.

Previously in `--no-scan` mode, addresses were imported, but not marked as exposed.

Resolves: https://github.com/zcash/wallet/issues/421
Resolves: [COR-1182](https://linear.app/zodl/issue/COR-1182/bug-zcashd-wallet-migration-transaction-buffer-does-not-update-key)

### DIFF
--- a/zallet/src/cli.rs
+++ b/zallet/src/cli.rs
@@ -199,7 +199,7 @@ pub(crate) struct MigrateZcashdWalletCmd {
     /// Skip chain scanning during migration. Keys, accounts, and transaction data are
     /// still imported, but block heights and tree state are not resolved from the chain.
     /// Useful when the corresponding chain data is not available.
-    #[arg(long, conflicts_with = "buffer_wallet_transactions")]
+    #[arg(long)]
     pub(crate) no_scan: bool,
 
     /// Temporary flag ensuring any alpha users are aware the migration is not stable.
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -11,6 +11,7 @@ use bip0039::{English, Mnemonic};
 use secp256k1::PublicKey;
 use secrecy::{ExposeSecret, SecretVec};
 use shardtree::error::ShardTreeError;
+use transparent::address::TransparentAddress;
 use zaino_fetch::jsonrpsee::response::block_header::GetBlockHeader;
 use zaino_state::{FetchServiceError, LightWalletIndexer, ZcashIndexer};
 use zcash_client_backend::{
@@ -26,7 +27,7 @@ use zcash_keys::keys::{
     zcashd::{PathParseError, ZcashdHdDerivation},
 };
 use zcash_primitives::{block::BlockHash, transaction::Transaction};
-use zcash_protocol::consensus::{BlockHeight, BranchId, NetworkType, Parameters};
+use zcash_protocol::consensus::{BlockHeight, BranchId, NetworkType, NetworkUpgrade, Parameters};
 use zewif_zcashd::{BDBDump, ZcashdDump, ZcashdParser, ZcashdWallet};
 use zip32::{AccountId, fingerprint::SeedFingerprint};
 
@@ -81,6 +82,7 @@ impl AsyncRunnable for MigrateZcashdWalletCmd {
             wallet,
             self.buffer_wallet_transactions,
             self.allow_multiple_wallet_imports,
+            self.no_scan,
         )
         .await?;
 
@@ -223,6 +225,7 @@ impl MigrateZcashdWalletCmd {
         wallet: ZcashdWallet,
         buffer_wallet_transactions: bool,
         allow_multiple_wallet_imports: bool,
+        no_scan: bool,
     ) -> Result<(), MigrateError> {
         let mut db_data = db.handle().await?;
         let network_params = *db_data.params();
@@ -282,11 +285,14 @@ impl MigrateZcashdWalletCmd {
             (None, None)
         };
         let sapling_activation = network_params
-            .activation_height(zcash_protocol::consensus::NetworkUpgrade::Sapling)
+            .activation_height(NetworkUpgrade::Sapling)
             .expect("Sapling activation height is defined.");
 
-        // Collect an index from txid to block height for all transactions known to the wallet that
-        // appear in the main chain.
+        // Collect an index from block hash to block height for all transactions known to the
+        // wallet that appear in the main chain. This only runs when we have a chain subscriber;
+        // without one, all transactions are stored as unmined and a later scan will assign
+        // accurate heights. Address exposure is handled separately via
+        // `mark_transparent_addresses_exposed` below.
         info!(
             "Wallet contains {} transactions",
             wallet.transactions().len(),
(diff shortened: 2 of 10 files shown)


zallet/src/components/database/connection.rs
```diff
@@ -213,6 +213,37 @@ impl DbConnection {
...
+        <WalletDb<rusqlite::Connection, Network, SystemClock, OsRng> as WalletRead>::Error,
+    > {
+        self.with_raw_mut(|conn, _| {
+            let tx = conn.transaction()?;
+            {
+                let mut stmt = tx.prepare_cached(
+                    "UPDATE transactions
+                     SET mined_height = NULL
+                     WHERE txid = :txid
+                       AND block IS NULL",
```
schell: I have a little concern here that we're getting into the database directly instead of going through a `zcash_backend_sqlite` interface (WalletDb), which makes this access dependent on the names and types of tables. 

But I guess this is not really a big deal because this is just for migrating from `zcashd`?

nullcopy: All of the zcashd migration logic is contained within Zallet, so it felt best to keep this here with the rest. Perhaps there's an argument to move the migration mechanics into librustzcash though. I think that would be for another PR though, if at all

schell (approved):

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -311,6 +311,59 @@ impl MigrateZcashdWalletCmd {
...
+                    }
+                }
+            }
+
+            // Pass 2: Handle mined transactions with zero expiry height. These are pre-Overwinter
+            // transactions or coinbase transactions, which predate the expiry mechanism. Their
+            // branch ID is embedded in the serialized transaction data (not derived from block
+            // height). We use the minimum known height from Pass 1, falling back to Sapling
+            // activation as a conservative last resort (pre-Overwinter transactions necessarily
+            // predate Sapling activation).
```
nuttycom: This is incorrect; zero expiry heights are valid for all transactions; they have never been disallowed. A zero expiry height simply means that the transaction will not expire, though it may have its anchor invalidated by a reorg or, in extreme cases, the transaction version might no longer be supported by the network.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -311,6 +311,59 @@ impl MigrateZcashdWalletCmd {
...
+            let fallback_height = main_chain_block_heights
+                .values()
+                .min()
+                .copied()
+                .unwrap_or(sapling_activation);
+            for (_, wallet_tx) in wallet.transactions().iter() {
+                let block_hash = BlockHash(*wallet_tx.hash_block().as_ref());
+                if block_hash.0 != [0; 32] {
+                    if let Entry::Vacant(entry) = main_chain_block_heights.entry(block_hash) {
+                        entry.insert(fallback_height);
```
nuttycom: Falling back to earliest transaction in the wallet, or the activation height, can definitely lead to the transaction being deserialized with the wrong consensus branch ID.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -694,6 +747,28 @@ impl MigrateZcashdWalletCmd {
             info!("Storing {} decrypted transactions", decrypted_txs.len());
             // TODO: Use chunking here if we add support for resuming migrations.
             db_data.store_decrypted_txs(decrypted_txs)?;
+
+            // In no-scan mode, the mined heights we just wrote are estimates derived
+            // from transaction expiry heights. They served their purpose of triggering
+            // `update_gap_limits` (which populates `addresses.exposed_at_height`), but
```
nuttycom: The fact that these heights are being used for exposure information wasn't evident to me earlier; it should be documented that that's what we're using those heights for.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -694,6 +747,28 @@ impl MigrateZcashdWalletCmd {
...
+            // we must not leave them in `transactions.mined_height`: `put_block`'s
+            // fix-up UPDATE trusts that column and would spuriously associate a
+            // transaction with the wrong block if a later scan happens to reach a
+            // block at an estimated height. Clear them now so the next real scan can
+            // populate them with accurate values.
+            if chain_subscriber.is_none() {
```
nuttycom: What changes to `librustzcash` would be necessary for us to be able to avoid setting incorrect mined_height values? Those are so central to so much logic that I'm very skeptical that we should use estimates for them.

zallet/src/cli.rs
```diff
@@ -199,7 +199,7 @@ pub(crate) struct MigrateZcashdWalletCmd {
     /// Skip chain scanning during migration. Keys, accounts, and transaction data are
     /// still imported, but block heights and tree state are not resolved from the chain.
     /// Useful when the corresponding chain data is not available.
-    #[arg(long, conflicts_with = "buffer_wallet_transactions")]
+    #[arg(long)]
```
nuttycom, line 202: The `no_scan` flag isn't used anywhere else in this PR; that's weird.

nullcopy: The `--no-scan` flag switches the chain subscriber above (see L65). This PR simply adds address exposure marking when in `--no-scan` mode

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: Good catch. This is copy-pasta documentation. The actual reason to fall-back to pass 2 is that pass 1 was using expiry height as a proxy to estimate mined height. If no expiry data is provided, we fall back to pass 2, which is less precise (and flawed, as you point out below).

I'll fix this comment.

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: That's a good point I hadn't considered. That might kill this whole 'pass 2' approach 🤔

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: Hmm, good question. I need to think on this some more, but my first idea is to add a `WalletWrite` method which I can use to tell the wallet "I observed address A", so that it can run through the address derivations and mark that and all prior addresses as exposed.

This doesn't solve the problem about not having an accurate exposure height, but it would eliminate these hacky workarounds... I need to think on this some more

nullcopy:
I have [a PR to add an address exposure API](https://github.com/zcash/librustzcash/pull/2305) to librustzcash. That should obviate the need for a lot of the complexity in this PR

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: This has been removed

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: Per this prompt, I've added the `WalletWrite::mark_transparent_addresses_as_exposed()` method to zcash_client_backend (https://github.com/zcash/librustzcash/pull/2305) which enables me to explicitly set exposure heights, instead of working around it with these temporary database updates. Much cleaner now.

nullcopy:
Closing this, as subsumed by https://github.com/zcash/wallet/pull/430

PR #428, opened 2026-04-08
deps: Migrate to latest releases of the Zcash crates

schell:
Migrates Zallet to use the published crates.io releases of the Zcash crates that match what was previously pinned via `[patch.crates-io]` to commit `0268bbc` of librustzcash. All 12 librustzcash crates (`equihash 0.3.0`, `f4jumble 0.1.1`, `zcash_address 0.11.0`, `zcash_client_backend 0.22.0`, `zcash_client_sqlite 0.20.0`, `zcash_encoding 0.4.0`, `zcash_history 0.4.0`, `zcash_keys 0.13.0`, `zcash_primitives 0.27.0`, `zcash_proofs 0.27.0`, `zcash_protocol 0.8.0`, `zcash_transparent 0.7.0`) have been published to crates.io at versions matching the git rev's content.

## Changes

- **`Cargo.toml`**: Drop the 12 librustzcash entries from `[patch.crates-io]`. The `age`, `zewif`, `zewif-zcashd`, and `zaino-*` patches remain since those crates are not published at the revisions we need. Add two new `[patch.crates-io]` entries pointing `core2` (^0.3 and ^0.4) at local shim crates that re-export `corez` (see Notes for review below).
- **Workspace dep version specifiers** (in `[workspace.dependencies]`): Already at the now-published versions from earlier commits in this PR — no further changes.
- **`zallet/src/components/database/connection.rs`** (from prior commits): Forwarders for three trait methods that became required at this librustzcash revision:
  - `WalletRead::find_account_for_address`
  - `WalletWrite::rewind_to_height`
  - `WalletWrite::mark_transparent_addresses_exposed`
- **`Cargo.lock`**: Regenerated. The librustzcash crates now resolve from `registry+...` instead of `git+...`.
- **`vendor/core2-shim-0.3/`** and **`vendor/core2-shim-0.4/`**: New temporary shim crates (~10 lines of Rust each) that re-export `corez` under the `core2` name to work around the upstream yanking of all `core2` registry versions on 2026-04-14. See "Notes for review" and `vendor/core2-shim-0.3/README.md` for full context and removal criteria.
- **`supply-chain/config.toml`**: Drop the `@git:0268bbc...` qualifiers from policy entries and exemptions; remove the obsolete `audit-as-crates-io = true` policy stubs that existed only to support the git-pinned versions; prune stale `core2` registry exemptions (no longer needed since the shims provide first-party `core2`). New exemptions for `orchard 0.13.1`, `sapling-crypto 0.7.0`, and `corez 0.1.1` (added in earlier commits of this PR) are retained.
- **`supply-chain/imports.lock`** and **`supply-chain/audits.toml`**: Refreshed via `cargo vet regenerate imports` and `cargo vet fmt` (the latter to match cargo-vet 0.10.2's preferred string serialization, used by CI).
- **`list_unspent.rs`** (from prior commit): Add required `TransparentOutputFilter::All` parameter to `get_spendable_transparent_outputs` calls -- preserves previous behavior (returns all transparent outputs, not coinbase-only).
- **`config.rs`** (from prior commit): Adapt `ConfirmationsPolicy::new(...)` error handling to match updated return type while keeping Zallet's existing `Result<_, ()>` API.

### `core2` shim crates (TEMPORARY)

All `core2` versions on crates.io were yanked by their author on 2026-04-14, breaking fresh dependency resolution (and thus the `Latest build on macOS-latest` CI job, which deletes `Cargo.lock` before building). Two transitive paths still request `core2`:

- `equihash 0.2.2` requires `core2 ^0.3` — pulled in via `zcash_primitives 0.24.1` / `zcash_proofs 0.24.0` (zaino's pinned librustzcash) and `zebra-chain 2.0.0` (our direct dep).
- `zaino-state 0.1.2` requires `core2 ^0.4` directly.

`vendor/core2-shim-0.3/` and `vendor/core2-shim-0.4/` are minimal vendored crates (~10 lines of Rust each) that re-export `corez` (the Zcash ecosystem's clean-room replacement at https://github.com/zcash/corez) under the `core2` name. `[patch.crates-io]` entries direct the resolver to these shims for `^0.3` and `^0.4` consumers respectively, using the `package = ...` rename trick from [Cargo's overriding-dependencies docs](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#using-patch-with-multiple-versions) to disambiguate two patch entries for the same crate name. The shims are excluded from the workspace so that two members named `core2` can coexist.

The shims are TEMPORARY and **must be removed** once `cargo tree --workspace --all-features -i core2` reports no results. Removal requires bumping our `zebra-chain`/`zaino-*` deps to revisions that use `zcash_primitives >= 0.27` (which uses `equihash 0.3.0` -> `corez`). Bumping zaino alone is currently blocked because zaino's `dev` branch requires `zebra-chain >= 4.x`, incompatible with our pinned `zebra-chain 2.0.0`. See `vendor/core2-shim-0.3/README.md` for the full removal criteria.

### Other behavioral notes

- **`OvkPolicy::Sender` semantics changed upstream.** Per the `zcash_client_backend` 0.22.0 changelog, when `OvkPolicy::Sender` is used (the only `OvkPolicy` site in Zallet, in `z_send_many.rs`), wallet-internal change outputs are now decrypted via the wallet's internal IVK rather than the sender's OVK. For Zallet's SQLite-backed `z_sendmany`, change remains recoverable via wallet sync, so user-visible behavior should be equivalent -- but worth confirming this is the desired semantic for Zallet's `z_sendmany`.
- **`NoteFilter::ExceedsMinValue` semantics changed from `>=` to `>`.** The single Zallet call site (`zallet/src/components/json_rpc/methods/get_notes_count.rs:54`) uses `Zatoshis::ZERO`, which previously matched all notes and now excludes 0-value (uneconomic) notes. Small behavioral change to `z_getnotescount`. Flagged for awareness; left as-is since 0-value notes are uneconomic.
- **All commits are folded as additions on top of prior commits** rather than amended/squashed, because each commit must pass `cargo clippy --all-targets -- -D warnings` per `AGENTS.md`, and the rev bumps introduce required trait methods that prevent intermediate "dep bump only" commits from building.

## CI failures unrelated to this PR

- **`Test NU7 on Linux`**: pre-existing failure on `main` since 2026-04-07. With `--cfg zcash_unstable="nu7"`, registry `zebra-chain 2.0.0` does a non-exhaustive match on `NetworkUpgrade` missing the `Nu7` variant. `continue-on-error: true` so doesn't block the PR.

## Verification

```
cargo fmt --all -- --check                                # clean
cargo clippy --all-targets -- -D warnings                 # clean
cargo test --workspace --all-features                     # 64 passed; 0 failed
cargo +beta clippy --tests --all-features --all-targets   # one pre-existing warning (migrate_zcashd_wallet.rs:301), unrelated
cargo vet --locked                                         # Vetting Succeeded (256 fully audited, 91 partially audited, 409 exempted)
rm Cargo.lock && cargo +stable build --workspace --all-targets --all-features  # succeeds (verified locally; this is the Latest build on macOS-latest CI job)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,19 +152,6 @@ anyhow = "1.0"
 tonic = "0.14"
 
 [patch.crates-io]
-equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4" }
 
 zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,14 @@
 members = [
     "zallet",
 ]
+# `vendor/core2-shim-0.3` and `vendor/core2-shim-0.4` are referenced from the
+# `[patch.crates-io]` table below, but excluded from the workspace because
+# Cargo requires workspace member crate names to be unique and both shims must
+# be named `core2` to satisfy the resolver.
+exclude = [
+    "vendor/core2-shim-0.3",
+    "vendor/core2-shim-0.4",
+]
 resolver = "2"
 
 [workspace.package]
@@ -152,6 +160,16 @@ anyhow = "1.0"
 tonic = "0.14"
 
 [patch.crates-io]
+# TEMPORARY: All `core2` versions on crates.io were yanked by upstream on
+# 2026-04-14, breaking fresh dependency resolution because we still pull
+# `core2 ^0.3` (via `equihash 0.2.2`) and `core2 ^0.4` (via `zaino-state`)
+# transitively. Both entries below redirect those `core2` requests to local
+# shims that re-export `corez` (the Zcash ecosystem's clean-room replacement,
+# https://github.com/zcash/corez). See `vendor/core2-shim-0.3/README.md` for
+# context and removal criteria.
+core2 = { path = "vendor/core2-shim-0.3" }
+core2_v04 = { package = "core2", path = "vendor/core2-shim-0.4" }
+
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4" }
 
 zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,9 +110,9 @@ orchard = "0.13"
 sapling = { package = "sapling-crypto", version = "0.7" }
 transparent = { package = "zcash_transparent", version = "0.7" }
 zcash_encoding = "0.4"
-zcash_keys = { version = "0.12", features = ["transparent-inputs", "sapling", "orchard", "transparent-key-encoding"] }
+zcash_keys = { version = "0.13", features = ["transparent-inputs", "sapling", "orchard", "transparent-key-encoding"] }
 zcash_primitives = "0.27"
-zcash_proofs = "0.26"
+zcash_proofs = "0.27"
 zcash_script = "0.4.3"
 secp256k1 = { version = "0.29", features = ["recovery"] }
 
@@ -136,8 +136,8 @@ schemerz-rusqlite = "0.370.0"
 shardtree = "0.6"
 time = "0.3"
 uuid = "1"
-zcash_client_backend = "0.21"
-zcash_client_sqlite = "0.19"
+zcash_client_backend = "0.22"
+zcash_client_sqlite = "0.20"
 zcash_note_encryption = "0.4"
 zip32 = "0.2"
 bip32 = "0.2"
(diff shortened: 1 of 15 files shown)


zallet/src/components/json_rpc/methods/list_unspent.rs
```diff
@@ -192,7 +192,12 @@ pub(crate) fn call(
             .iter()
             .try_fold(vec![], |mut acc, (addr, _)| {
                 let mut outputs = wallet
-                    .get_spendable_transparent_outputs(addr, target_height, confirmations_policy)
+                    .get_spendable_transparent_outputs(
+                        addr,
+                        target_height,
+                        confirmations_policy,
+                        TransparentOutputFilter::All,
+                    )
```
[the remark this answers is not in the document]
schell: @copilot Yes, but what was the _previous_ behavior? The only difference with the new parameter is that if passed `TransparentOutputFilter::Coinbase`, it will only return coinbase transaction outputs. So here this change should result in the same behavior as previously.

ValarDragon (commented):
LGTM, its a valid dependency bump to: https://github.com/zcash/librustzcash/commits/main/?after=976efa76ca2195d693f373aeaa201a2c50b6e0ab+69

Assuming correct implementation upstream of get_spendable_transparent_outputs , this is correct. It will help remove excess updates in other PR's that touch dependency updates

Cargo.toml
```diff
@@ -152,18 +152,18 @@ anyhow = "1.0"
...
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "cb6c182f9ba0f5c29d8e43bb8aee4c3ae37d0dcb" }
```
nuttycom: This is still quite a ways behind `main`; we should consider zcash/librustzcash@976efa76ca2195d693f373aeaa201a2c50b6e0ab instead.

Cargo.toml
nuttycom: Since the releases last week, you'll need to update `orchard` and `sapling-crypto` to 0.13 and 0.7 respectively if we update to zcash/libruszcash@976efa76ca2195d693f373aeaa201a2c50b6e0ab

Cargo.toml
```diff
@@ -152,18 +160,15 @@ anyhow = "1.0"
...
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "eb70164b0ea72f87dba8ae54e5e0fd5f5be44029" }
+# TEMPORARY: All `core2` versions on crates.io were yanked by upstream on
+# 2026-04-14, breaking fresh dependency resolution because we still pull
+# `core2 ^0.3` (via `equihash 0.2.2`) and `core2 ^0.4` (via `zaino-state`)
+# transitively. Both entries below redirect those `core2` requests to local
+# shims that re-export `corez` (the Zcash ecosystem's clean-room replacement,
+# https://github.com/zcash/corez). See `vendor/core2-shim-0.3/README.md` for
+# context and removal criteria.
+core2 = { path = "vendor/core2-shim-0.3" }
```
nullcopy, line 170: I wonder if we should create a tracking issue to follow up on this

vendor/core2-shim-0.4/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "core2"
+version = "0.4.99"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+description = "Temporary local shim providing the core2 0.4.x API by re-exporting `corez`. See README.md for context and removal criteria."
```
nullcopy, line 6: Is shimming really the best way forward here? I'd almost rather crate patch to our own forks with `core2` replaced. This shim just feels like something that will get forgotten about

But that's just my opinion. Not a blocker

nullcopy (approved):
I don't love vendoring a few `core2` shims, but I accept that may be the easiest/quickest way out of the tricky situation `core2` put us in.

utACK 6025a0d343a48a2a4bb2e5adb086c19e2f2bb8e9

schell:
@nullcopy - I also don't love the shims. The `zebra` PR to bump deps merged today, though, so they might not be here long.

vendor/core2-shim-0.4/Cargo.toml
schell: It really depends on how long it takes `zebra` crates to release, and then for `zaino` crates to update. This at least gets us merging again.

ValarDragon (commented):
This PR LGTM! 

 Reviewed thoroughly:
- Shim crates
- The changes for updating

I think we should proceed with publishing this to unblock other work.

Other work has been blocked for 3+ weeks. We should get this in, and then do another update when upstream is ready. Getting Zallet miner readyness is great

PR #430, opened 2026-04-21
Add migration support for all zcashd key types

### LINKED ISSUE #424: bug: keys imported into zcashd via `importprivkey` do not get imported during `migrate-zcashd-wallet`
After `zallet migrate-zcashd-wallet --no-scan` on a zcashd `wallet.dat` containing transparent private keys added via `importprivkey`, the corresponding addresses do not appear in the `listaddresses` RPC output, even though the migration command exits successfully and the relevant code path *appears* to import them.

These failures can be observed by running the `zcash-integration-tests/qa/rpc-tests/zcashd_key_import.py` integration test (currently pending review: https://github.com/zcash/integration-tests/pull/82

## Reproduction
To manually reproduce the failures:
1. Build Zallet with `--features zcashd-import`
1. Use `qa/rpc-tests/test-wallet/wallet.dat` from the integration-tests branch above. It contains 9 phases (Sprout → NU6.1) each adding an imported transparent privkey via `zcashd`'s `importprivkey`.
1. Run:
   ```
   zallet -d=$DATADIR migrate-zcashd-wallet \
     --path wallet.dat \
     --no-scan \
     --buffer-wallet-transactions \
     --allow-warnings \
     --this-is-alpha-code-and-you-will-need-to-redo-the-migration-later
   ```
3. Start the wallet and call `listaddresses`.

## Expected
All 9 imported transparent privkey addresses appear in `listaddresses`.

## Actual
None of the 9 appear. HD-derived transparent / sapling / orchard addresses and imported sapling **spending** keys all do appear correctly. `listaddresses` returns 260 addresses total; the 9 below are not among them.

| Phase | Address |
|-------|---------|
| 0 | `tmGMqHGjnw3447yHsgTZRDsX2oaokkJLyrc` |
| 1 | `tmPqawQFBzFZS5wDtZT4VoHPSCHwW6JEeTW` |
| 2 | `tmUXaxJNVYjqonz1v2BV9uvFfo8eesejiFU` |
| 3 | `tmYnhoLZA7Ca9EdoDxxYXzXvddrwxZUxkwT` |
| 4 | `tmA3QDF3EnRMv7WFaUKoFwkj6szdW4PPrmK` |
| 5 | `tmBPvGztds5ub4xz6Tse1SYtV3HVMcVMrdn` |
| 6 | `tmASTcm3TpxZrP3mjJg1WMnF5QKgRSfYe4n` |
| 7 | `tmDFXtEMm6eS3kq5YRjSDymMxCYpfcBa8hU` |
| 8 | `tmACKQUJDeGj4bVd5Li15mx1ZYxYSxzq4Ex` |

## Additional context
What's puzzling, is that the migration code at `zallet/src/commands/migrate_zcashd_wallet.rs:649-669` already does what looks like the right thing for these keys:

A few possible hypotheses (un-investigated):
1. `zewif`'s `wallet.keys().keypairs()` iterator doesn't yield keys that
   originated from `importprivkey` (only HD-derived legacy transparent keys).
   The `info!` line above would say `Importing 0 legacy standalone transparent
   keys` if so — easy to confirm by re-running with `RUST_LOG=info`.
2. Keys are stored under a different account UUID than the one `listaddresses`
   walks for the `mnemonic_seed` source group.
3. The pubkey→address derivation in `import_standalone_transparent_pubkeys`
   produces a different P2PKH address than the original `tm…` from
   `importprivkey` (e.g. compressed vs. uncompressed pubkey mismatch).

### LINKED ISSUE #426: bug: imported P2PKH & P2SH addresses not migrated from zcashd
zcashd's `importaddress` allows importing P2SH and P2PKH addresses, but zcashd migration has no code path to migrate them into Zallet.

This can be observed by running the `zcash-integration-tests/qa/rpc-tests/zcashd_key_import.py` integration test (currently pending review: https://github.com/zcash/integration-tests/pull/82

Example: in the above integration test, the following P2SH and P2PKH address are expected (from the phase 4 manifest), but are not returned in a query to `listaddresses`:
* example of missing P2SH: `t27VLDjeyhVjpZrKP64vX9NszZ3ifUPpSCN`
* example of missing P2PKH: `tm9sWbYCFpn9awN9o1aAAJAVEixQN3ZteEb`

### LINKED ISSUE #427: bug: Sapling view-keys not migrated from zcashd
Migration handles standalone sapling **spending** keys via `encrypt_and_store_standalone_sapling_key` (line 600), but there is no equivalent path for standalone `ExtendedFullViewingKey`s

nullcopy:
Based on https://github.com/zcash/wallet/pull/428 for latest zcash crate releases. The first 8 commits of this PR will drop out once that merges. I recommend reviewing the last three commits in isolation.

This PR does two things to improve migration coverage
1) mark addresses as exposed during key import
  * use zcashd wallet.dat mined transaction data to know if an address was exposed
  * estimate wallet birthday height to use as conservative exposure, since actual exposure cannot be known
2) parse keys from the `watchs` and `cscript` tables from zcashd
  * pending https://github.com/zcash/zewif-zcashd/pull/7
  * enables parsing standalone key types (p2pkh & p2sh) from their tables in zcashd wallet.dat

Resolves: https://github.com/zcash/wallet/issues/426
Resolves: https://github.com/zcash/wallet/pull/423
Resolves: https://github.com/zcash/wallet/issues/424
Resolves: https://github.com/zcash/wallet/issues/427
Resolves: [COR-1189](https://linear.app/zodl/issue/COR-1189/bug-imported-p2pkh-and-p2sh-addresses-not-migrated-from-zcashd), [COR-1182](https://linear.app/zodl/issue/COR-1182/bug-zcashd-wallet-migration-transaction-buffer-does-not-update-key), [COR-1187](https://linear.app/zodl/issue/COR-1187/bug-keys-imported-into-zcashd-via-importprivkey-do-not-get-imported), [COR-1190](https://linear.app/zodl/issue/COR-1190/bug-sapling-view-keys-not-migrated-from-zcashd)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -173,7 +173,10 @@ core2_v04 = { package = "core2", path = "vendor/core2-shim-0.4" }
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4" }
 
 zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
-zewif-zcashd = { git = "https://github.com/zcash/zewif-zcashd.git", rev = "9c72d1805c269f88ea5caa4eb33c6d38013b9776" }
+# Temporarily pointed at https://github.com/zcash/zewif-zcashd/pull/7 (parses
+# `watchs` and `cscript` BDB records into `ZcashdWallet`). Revert to the
+# `zcash/zewif-zcashd` rev once the PR has merged.
+zewif-zcashd = { git = "https://github.com/nullcopy/zewif-zcashd.git", rev = "94737a9c9100ba44415d8931790c2f3cf41b9758" }
 
 zaino-common = { git = "https://github.com/zingolabs/zaino.git", rev = "15b81f110349b34090341b3ab8b7cd58c7b9aeef" }
 zaino-fetch = { git = "https://github.com/zingolabs/zaino.git", rev = "15b81f110349b34090341b3ab8b7cd58c7b9aeef" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,19 +152,6 @@ anyhow = "1.0"
 tonic = "0.14"
 
 [patch.crates-io]
-equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "0268bbcf8007bff6a7a94cb7c1e229b595a6489e" }
-
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4" }
 
 zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,14 @@
 members = [
     "zallet",
 ]
+# `vendor/core2-shim-0.3` and `vendor/core2-shim-0.4` are referenced from the
+# `[patch.crates-io]` table below, but excluded from the workspace because
+# Cargo requires workspace member crate names to be unique and both shims must
+# be named `core2` to satisfy the resolver.
+exclude = [
+    "vendor/core2-shim-0.3",
+    "vendor/core2-shim-0.4",
+]
 resolver = "2"
 
 [workspace.package]
@@ -152,6 +160,16 @@ anyhow = "1.0"
 tonic = "0.14"
 
 [patch.crates-io]
+# TEMPORARY: All `core2` versions on crates.io were yanked by upstream on
+# 2026-04-14, breaking fresh dependency resolution because we still pull
+# `core2 ^0.3` (via `equihash 0.2.2`) and `core2 ^0.4` (via `zaino-state`)
+# transitively. Both entries below redirect those `core2` requests to local
+# shims that re-export `corez` (the Zcash ecosystem's clean-room replacement,
+# https://github.com/zcash/corez). See `vendor/core2-shim-0.3/README.md` for
+# context and removal criteria.
+core2 = { path = "vendor/core2-shim-0.3" }
+core2_v04 = { package = "core2", path = "vendor/core2-shim-0.4" }
+
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4" }
 
 zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,9 +110,9 @@ orchard = "0.13"
 sapling = { package = "sapling-crypto", version = "0.7" }
 transparent = { package = "zcash_transparent", version = "0.7" }
 zcash_encoding = "0.4"
-zcash_keys = { version = "0.12", features = ["transparent-inputs", "sapling", "orchard", "transparent-key-encoding"] }
+zcash_keys = { version = "0.13", features = ["transparent-inputs", "sapling", "orchard", "transparent-key-encoding"] }
 zcash_primitives = "0.27"
-zcash_proofs = "0.26"
+zcash_proofs = "0.27"
 zcash_script = "0.4.3"
 secp256k1 = { version = "0.29", features = ["recovery"] }
 
@@ -136,8 +136,8 @@ schemerz-rusqlite = "0.370.0"
 shardtree = "0.6"
 time = "0.3"
 uuid = "1"
-zcash_client_backend = "0.21"
-zcash_client_sqlite = "0.19"
+zcash_client_backend = "0.22"
+zcash_client_sqlite = "0.20"
 zcash_note_encryption = "0.4"
 zip32 = "0.2"
 bip32 = "0.2"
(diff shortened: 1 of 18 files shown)


zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -615,6 +654,111 @@ impl MigrateZcashdWalletCmd {
...
+            );
+            for script in watchonly_scripts {
+                let script_addr =
+                    TransparentAddress::from_script_pubkey(&zcash_script::descriptor::sh(&script));
+                // `import_standalone_transparent_script` currently only accepts multisig
+                // redeem scripts; skip anything else so a single unsupported script does
+                // not abort the migration.
+                if db_data
+                    .import_standalone_transparent_script(target_account, script)
+                    .is_ok()
```
schell: Is it ok to swallow errors here? Should we log them?

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -223,11 +229,30 @@ impl MigrateZcashdWalletCmd {
...
+        // (entries in `cscripts()`). Address-only imports cannot be represented in the
+        // Zallet wallet schema and are silently skipped.
+        let watchonly_pubkeys = wallet
+            .watch_scripts()
+            .iter()
+            .filter_map(|w| match w.kind() {
+                WatchScriptKind::P2PK(pubkey) => PublicKey::from_slice(pubkey.as_slice()).ok(),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
```
schell, line 282: Claude pointed out that "both compressed and uncompressed pubkeys flow through here". But I don't know if that's actually an issue.

schell (commented):
I had a couple questions, but for the most part this area is not my wheelhouse, so 🧂 . 

I do know, though, that this can be rebased on `main` to get the underlying `core2` shims.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -223,11 +229,99 @@ impl MigrateZcashdWalletCmd {
...
+        // entries with a matching `cscripts()` record are imported via that path below.
+        let mut skipped_p2pkh_address_only = 0usize;
+        let mut skipped_p2sh_address_only = 0usize;
+        let mut skipped_nonstandard = 0usize;
+        let mut skipped_malformed_pubkeys = 0usize;
+        let watchonly_pubkeys = wallet
+            .watch_scripts()
+            .iter()
+            .filter_map(|w| match w.kind() {
+                WatchScriptKind::P2PK(pubkey) => match PublicKey::from_slice(pubkey.as_slice()) {
```
dianaravenborne, line 262: 💡 **Encoding loss on watch-only P2PK pubkeys.** `WatchScriptKind::P2PK(pubkey)` here is the raw bytes as stored in `zcashd`'s `watchs` record — these may be either 33-byte compressed *or* 65-byte uncompressed (`importpubkey` accepts both). `secp256k1::PublicKey::from_slice` happily parses both, but the resulting `PublicKey` doesn't remember which encoding it came from, and downstream we feed it into `TransparentAddress::from_pubkey`, which calls `pubkey.serialize()` (always compressed). For uncompressed entries the derived P2PKH `hash160` will not match the address `zcashd` actually had on-chain, so the corresponding row in `to_expose ∩ known` (a few hundred lines down) will silently miss.

Two options:
- Preserve the original byte representation and use that to compute the address (e.g. `Hash160::digest(raw_bytes)`), or
- Detect 65-byte entries and either reject them with a warn or import them under the uncompressed-derived P2PKH explicitly.

This is the concrete form of the issue Claude flagged in schell's earlier comment.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -223,11 +229,99 @@ impl MigrateZcashdWalletCmd {
...
+            warn!(
+                "Skipped {} watch-only entries with non-standard script kinds.",
+                skipped_nonstandard,
+            );
+        }
+        let mut skipped_unparseable_scripts = 0usize;
+        let watchonly_scripts = wallet
+            .cscripts()
+            .values()
+            .filter_map(|s| match Redeem::parse(&Code(s.as_ref().to_vec())) {
```
dianaravenborne, line 324: 📝 **Nit / future-proofing.** `Code(s.as_ref().to_vec())` copies the script bytes. If `Redeem::parse` can take a `&[u8]` or `Code(&[u8])` in a future `zcash_script` version, this hot-loop-ish allocation goes away. Not blocking — just leaving a marker.

dianaravenborne (commented):
## Hermes Agent Review

Reviewed the last three commits in isolation (per the PR description). Overall the change is well-structured and the warning hygiene around skipped entries is excellent. I have two correctness questions and one observability suggestion — none clearly blocking, hence `Comment` rather than `Request changes`.

### ❓ Open questions
1. **Compressed-vs-uncompressed pubkeys flowing through `TransparentAddress::from_pubkey`** — picking up the thread schell / Claude opened. `TransparentAddress::from_pubkey` in `zcash_transparent` calls `pubkey.serialize()`, which always emits the 33-byte compressed form, whereas `zcashd`'s `importpubkey` accepts uncompressed (65-byte) pubkeys and would have tracked the P2PKH of the uncompressed encoding. After we round-trip through `secp256k1::PublicKey::from_slice(...)` we lose the original encoding choice, so for any P2PK entry imported in uncompressed form the derived P2PKH won't match what `zcashd` actually had — and the `to_expose ∩ known` intersection silently drops it. See inline comment on `migrate_zcashd_wallet.rs` for the specific spots.
2. **Silent drops in `to_expose ∩ known`** — see inline. Worth at least an info-log of the count that fell out, otherwise migration bugs in the import path are invisible.

### ✅ Looks good
- Clean separation between `watchs` (P2PK / P2PKH / P2SH / Other) and `cscript` paths, with explicit warn-logs for every skip reason.
- Sapling EFVK 169-byte round-trip with a clear comment explaining why the `.expect()` is sound (stable ZIP 32 encoding).
- `get_account_for_ufvk` guard correctly dedupes when both `z_importkey` and `z_importviewingkey` were called for the same key.
- Explicit-exposure accumulator pattern is a nice fix for the `listaddresses` / `exposed_at_height IS NOT NULL` interaction.
- CHANGELOG updated, no-scan ↔ buffer-wallet-transactions conflict removal documented.

### 📦 Pre-merge
- `Cargo.toml` is pointed at `nullcopy/zewif-zcashd` — needs to revert to `zcash/zewif-zcashd` after [zcash/zewif-zcashd#10](https://github.com/zcash/zewif-zcashd/pull/10) merges (already noted in the PR body, flagging here for the merge checklist).
- `Test NU7 on Linux` CI job is red, but the error is `non-exhaustive patterns: zcash_primitives::consensus::NetworkUpgrade::Nu7 not covered` inside `zebra-chain`, which looks orthogonal to this PR. Worth confirming it's pre-existing / a flaky/main-branch issue before merge.

---
*Reviewed by Hermes Agent (acting on behalf of dianaravenborne).*

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -570,6 +678,40 @@ impl MigrateZcashdWalletCmd {
...
+                info!("Processed {} view-only Sapling keys", idx);
+            }
+            // `zewif-zcashd` parses Sapling types against an older `sapling-crypto`. The
+            // ZIP 32 EFVK encoding is 169 bytes and stable across versions, so round-trip
+            // through bytes to get the version this crate uses.
+            let mut bytes = [0u8; 169];
+            zewif_extfvk
+                .write(&mut bytes[..])
+                .expect("Sapling EFVK fits in 169 bytes");
+            let extfvk = sapling::zip32::ExtendedFullViewingKey::read(&bytes[..])
```
dianaravenborne, line 715: 👍 Nice. The 169-byte round-trip is the cleanest fix here, and the `.expect()` is genuinely sound because ZIP 32 EFVK encoding is fixed-width and version-stable. Worth keeping the comment exactly as written for future maintainers who'll wonder why this isn't fallible.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -570,6 +678,40 @@ impl MigrateZcashdWalletCmd {
...
+                .expect("Sapling EFVK encoding is stable across sapling-crypto versions");
+            let ufvk = UnifiedFullViewingKey::from_sapling_extended_full_viewing_key(extfvk)?;
+
+            if db_data.get_account_for_ufvk(&ufvk)?.is_none() {
+                db_data.import_account_ufvk(
+                    &format!("zcashd imported view-only sapling {}", idx),
```
dianaravenborne, line 721: 💡 **Suggestion: stable, user-facing name for view-only accounts.** Using the `BTreeMap`-iteration `idx` as the name (`zcashd imported view-only sapling 0`, `… 1`, …) means a user re-running migration on a clone of the same wallet.dat will get the same names — good — but the indices don't correspond to anything meaningful in the source wallet (no `z_importviewingkey` label is preserved, and the order isn't the import order). Would it be worth using e.g. a short fingerprint of the EFVK as a suffix instead of `idx`, so the account name is identity-stable across re-imports rather than position-stable? Not blocking.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -608,13 +756,154 @@ impl MigrateZcashdWalletCmd {
...
+                );
+            }
+
+            let birthday_height = wallet_birthday.height();
+            let to_mark: Vec<(TransparentAddress, BlockHeight)> = to_expose
+                .intersection(&known)
+                .map(|addr| (*addr, birthday_height))
+                .collect();
+            if !to_mark.is_empty() {
+                db_data.mark_transparent_addresses_exposed(&to_mark)?;
```
dianaravenborne, line 926: 💡 **Observability:** the `to_expose ∩ known` filter is doing real work — it'll drop any address we *thought* we imported but that didn't end up in `get_transparent_receivers` (e.g. the compressed/uncompressed mismatch flagged above, or a redeem-script that failed `import_standalone_transparent_script` after we'd already pushed it onto `explicit_exposures`). Currently those silent drops are invisible. Consider logging the difference:

```rust
let dropped = to_expose.difference(&known).count();
if dropped > 0 {
    warn!("{} addresses queued for exposure marking were not tracked by any account and were skipped", dropped);
}
```

Makes regressions in the import paths above easier to spot in user logs.

Cargo.toml
```diff
@@ -173,8 +173,11 @@ core2_v04 = { package = "core2", path = "vendor/core2-shim-0.4" }
 
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4" }
 
-zewif = { git = "https://github.com/zcash/zewif.git", rev = "f84f80612813ba00a0a8a9a5f060bd217fa981cc" }
-zewif-zcashd = { git = "https://github.com/zcash/zewif-zcashd.git", rev = "9c72d1805c269f88ea5caa4eb33c6d38013b9776" }
+zewif = { git = "https://github.com/zcash/zewif.git", rev = "798d7b3554ebada05bea63b67ee66f9c2558a21a" }
+# Temporarily pointed at https://github.com/zcash/zewif-zcashd/pull/10 (parses
+# `watchs`, `cscript`, and `sapextfvk` BDB records into `ZcashdWallet`).
+# Revert to the `zcash/zewif-zcashd` rev once the PR has merged.
+zewif-zcashd = { git = "https://github.com/nullcopy/zewif-zcashd.git", rev = "cb2efe7f49b97f10d62100db3adcacbae31c28b1" }
```
dianaravenborne: 📦 **Merge gate reminder:** this needs to flip back to `zcash/zewif-zcashd` (and `Cargo.lock` regenerated) once https://github.com/zcash/zewif-zcashd/pull/10 lands. The inline comment above is great — leaving this note so it shows up in the inline review thread for the merge checklist.

nullcopy:
All feedback addressed

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -570,6 +697,40 @@ impl MigrateZcashdWalletCmd {
             }
         }
 
+        // Import view-only Sapling keys added via `z_importviewingkey`. zcashd stores
+        // these as `sapextfvk` BDB records; each EFVK becomes its own view-only account,
```
dannywillems: As mentioned by Daira [here](https://github.com/zcash/zewif-zcashd/pull/10#discussion_r3324638826) and [here](https://github.com/zcash/zewif-zcashd/pull/10#discussion_r3324627914), it would be better to use `extended FVK` to not create new acronyms.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -570,6 +697,40 @@ impl MigrateZcashdWalletCmd {
...
+            if idx % 100 == 0 && idx > 0 {
+                info!("Processed {} view-only Sapling keys", idx);
+            }
+            // `zewif-zcashd` parses Sapling types against an older `sapling-crypto`. The
+            // ZIP 32 EFVK encoding is 169 bytes and stable across versions, so round-trip
+            // through bytes to get the version this crate uses.
+            let mut bytes = [0u8; 169];
+            zewif_extfvk
+                .write(&mut bytes[..])
+                .expect("Sapling EFVK fits in 169 bytes");
```
dannywillems: See https://github.com/zcash/wallet/pull/430/changes#r3372395175

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -570,6 +697,40 @@ impl MigrateZcashdWalletCmd {
...
+            let extfvk = sapling::zip32::ExtendedFullViewingKey::read(&bytes[..])
+                .expect("Sapling EFVK encoding is stable across sapling-crypto versions");
```
dannywillems: See https://github.com/zcash/wallet/pull/430/changes#r3372395175

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/7f9344a15992beecc1adb9a874c796b21cdb2634..b97045a74ca0ef491d6314e447914f1c258fc4b6) to address feedback and rebase onto main

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -223,11 +229,118 @@ impl MigrateZcashdWalletCmd {
...
+                // different address than `zcashd` had on-chain. Skip with a warn
+                // rather than silently migrating to the wrong address.
+                WatchScriptKind::P2PK(pubkey) if !pubkey.is_compressed() => {
+                    skipped_uncompressed_pubkeys += 1;
+                    None
+                }
+                WatchScriptKind::P2PK(pubkey) => match PublicKey::from_slice(pubkey.as_slice()) {
+                    Ok(pk) => Some(pk),
+                    Err(_) => {
+                        skipped_malformed_pubkeys += 1;
```
dannywillems, line 265: Nit: I'm wondering if we shouldn't raise an exception, or have a `strict` mode, when we detect that an address is corrupted. It might be coming from critical bugs or larger corrupted data.

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -605,13 +771,168 @@ impl MigrateZcashdWalletCmd {
...
+        }
+
+        // Collect transparent addresses that need to be explicitly marked as exposed at
+        // the wallet birthday:
+        //   * In no-scan mode, every address observed in the wallet's transaction set, since
+        //     the chain will not be scanned to detect exposure heights.
+        //   * In any mode, every imported watch-only address (from `importaddress` /
+        //     `importpubkey`), since these have no on-chain derivation history that a chain
+        //     scan would discover.
+        let mut to_expose: HashSet<TransparentAddress> = HashSet::new();
```
dannywillems: Why do you use a later hashset `to_expose`?
Can it be defined above, as `explicit_exposures` (which could be actually `to_expose`)?

dannywillems (approved):
utACK b97045a74ca0ef491d6314e447914f1c258fc4b6.

I think there is room for improvements in the logic, and we could reduce the code size. I left a few comments that are mostly nitpicks.
As `zallet` must be deployed soon, I am focusing on the actual feature being implemented. I don't want to block anything.

In the near future, I guess we will have actual e2e tests, see https://github.com/zcash/wallet/issues/409

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: In the case of migrating from old zcashd wallets, if a failure like this happens, I would still prefer to migrate _something_ and let the user know we may not have migrated everything, rather than fail to migrate anything at all. Either way the user can still open an issue or reach out to us about the migration errors, but at least they will not be blocked on us to get access to at least part of their funds.

I could be convinced otherwise, but I say we defer that to a follow up if you think it needs following up on.

zallet/src/commands/migrate_zcashd_wallet.rs
nullcopy: No good reason. Fixed.

nullcopy:
[Force-pushed](https://github.com/zcash/wallet/compare/b97045a74ca0ef491d6314e447914f1c258fc4b6..7688a633142d09a642d623be04ebfef69bf5e50d) to address feedback

dannywillems (approved):
utACK 7688a633142d09a642d623be04ebfef69bf5e50d.

Issue #405, opened 2026-03-26
test: Need integration tests for zcashd migration


nullcopy:
In addition to https://github.com/zcash/wallet/issues/263, which asks for manual verification of Sapling key import, we should improve coverage to both:
* automate migration testing + CI
* coverall Zcashd key types (not just Sapling)
* for a wallet.dat file exported from any zcashd version we wish to support migration from

IMO, the deliverable will be an integration test (hooked to the CI runner) which:
* imports a golden wallet.dat(s) from the supported zcashd version(s)
* a series of test queries to ensure all keys are correctly identified and associated with the expected accounts

TODO:
- [ ] Build integration test (WIP https://github.com/zcash/integration-tests/pull/82)
- [ ] Enable test in Zallet CI
- [ ] Enable test in integration suite (see https://github.com/zcash/integration-tests/pull/82/changes/1958ba914ced302b5e75d696dddda1715c4d7833)

nullcopy:
Once https://github.com/zcash/integration-tests/pull/82 merges, all that will remain is to add the new integration test to Zallet's CI. See https://github.com/zcash/integration-tests/pull/82/changes/c6a1053f413ccbda9b4b71e5eb8e45d54e5bed29

Issue #16, opened 2025-01-30
Write a test harness for checking equivalence of `zcashd` and Zallet output


str4d:
It would be nice to have tooling that can take a fully-synced `zcashd` wallet, and a Zallet wallet containing the same set of keys, and diff the output of the various JSON-RPC methods to locate differences. Then we can either address the discrepancies, or document them as intentional.

nuttycom:
This is satisfied by the https://github.com/zcash/integration-tests repository

PR #431, opened 2026-04-24
rpc: Implement z_importviewingkey for Sapling

### LINKED ISSUE #80: rpc: Implement `z_importviewingkey`
Only Sapling (`zcashd` only supported Sprout or Sapling).

mnm458:
## Summary                                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Implements `z_importviewingkey` RPC method for Sapling extended full viewing keys (issue #80).                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                 
- Imports a Sapling extfvk (`zxviews`/`zxviewtestsapling`) so the wallet can track transactions without holding spending authority. Accounts are stored with `AccountPurpose::ViewOnly`.                                                                                                                                                                                                                                       
  - Errors if the wallet already holds the spending key for the given viewing key (matching zcashd behavior).
  - Adds `fetch_account_birthday` to `json_rpc::utils` as a shared helper (also needed by #400 — whichever merges first provides it).                                                                                                                                                                                                                                                                                            
  - Includes unit tests for parameter validation, key decoding, address derivation, encoding roundtrips, and rejection of invalid/wrong-network/spending keys.
                                                                                                                                                                                                                                                                                                                                                         
  **Why implement this despite zero usage in the survey?** Watch-only wallets are a fundamental capability for exchanges, auditors, and operators who need to monitor balances without holding spending keys. As zcashd is deprecated, this becomes the only path for users to import existing viewing keys into the new stack. It also supports the separation-of-concerns model where spending keys stay on cold storage while 
  a hot node tracks balances.                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  ## Test plan                                              

  - [x] `cargo build` — clean                                                                                                                                                                                                                                                                                                                                                                                                    
  - [x] `cargo test -p zallet import_viewing_key` — 13/13 passing
  - [ ] Integration test with live wallet DB (blocked on #405 / #16)                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Close #80

### DIFF
--- a/zallet-core/src/components/json_rpc/methods.rs
+++ b/zallet-core/src/components/json_rpc/methods.rs
@@ -42,6 +42,8 @@ mod get_wallet_status;
 pub(crate) mod help;
 #[cfg(zallet_build = "wallet")]
 mod import_key;
+#[cfg(zallet_build = "wallet")]
+mod import_viewing_key;
 mod list_accounts;
 mod list_addresses;
 #[cfg(zallet_build = "wallet")]
@@ -458,6 +460,29 @@ pub(crate) trait WalletRpc {
         hex_data: &str,
         rescan: Option<bool>,
     ) -> z_import_address::Response;
+
+    /// Imports a Sapling viewing key into the wallet.
+    ///
+    /// Only Sapling extended full viewing keys are supported. The wallet will track
+    /// incoming and outgoing transactions for addresses derived from this key, but
+    /// will not have spending authority.
+    ///
+    /// # Arguments
+    ///
+    /// - `vkey` (string, required) The viewing key (see `z_exportviewingkey`).
+    /// - `rescan` (string, optional, default="whenkeyisnew") Whether to rescan the
+    ///   blockchain for transactions ("yes", "no", or "whenkeyisnew"). When rescan is
+    ///   enabled, the wallet's background sync engine will scan for historical
+    ///   transactions from the given start height.
+    /// - `startHeight` (numeric, optional, default=0) Block height from which to begin
+    ///   the rescan. Only used when rescan is "yes" or "whenkeyisnew" (for a new key).
+    #[method(name = "z_importviewingkey")]
+    async fn import_viewing_key(
+        &self,
+        vkey: &str,
+        rescan: Option<&str>,
+        start_height: Option<u64>,
+    ) -> import_viewing_key::Response;
 
     /// Returns the total value of funds stored in the node's wallet.
     ///
@@ -1043,6 +1068,22 @@ impl<C: Chain> WalletRpcServer for WalletRpcImpl<C> {
             account,
             hex_data,
             rescan,
+        )
+        .await
+    }
+
+    async fn import_viewing_key(
+        &self,
+        vkey: &str,
+        rescan: Option<&str>,
+        start_height: Option<u64>,
+    ) -> import_viewing_key::Response {
+        import_viewing_key::call(
+            self.wallet().await?.as_mut(),
+            self.chain().await?,
+            vkey,
+            rescan,
+            start_height,
         )
         .await
     }
--- a/zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
+++ b/zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
@@ -1,0 +1,346 @@
+use documented::Documented;
+use jsonrpsee::core::RpcResult;
+use schemars::JsonSchema;
+use serde::Serialize;
+use zcash_client_backend::data_api::{Account, AccountPurpose, WalletRead, WalletWrite};
+use zcash_keys::{
+    encoding::{decode_extended_full_viewing_key, encode_payment_address},
+    keys::UnifiedFullViewingKey,
+};
+use zcash_protocol::consensus::{BlockHeight, NetworkConstants};
+
+use crate::components::{
+    chain::Chain,
+    database::DbConnection,
+    json_rpc::{server::LegacyCode, utils::fetch_account_birthday},
+};
+
+/// Response to a `z_importviewingkey` RPC request.
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// Result of importing a viewing key.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct ResultType {
+    /// The type of the imported address (always "sapling").
+    address_type: String,
+
+    /// The Sapling payment address corresponding to the imported viewing key
+    /// (the default address).
+    address: String,
+}
+
+pub(super) const PARAM_VKEY_DESC: &str =
+    "The viewing key (only Sapling extended full viewing keys are supported).";
+pub(super) const PARAM_RESCAN_DESC: &str = "Whether to rescan the blockchain for transactions (\"yes\", \"no\", or \"whenkeyisnew\"; default is \"whenkeyisnew\"). When rescan is enabled, the wallet's background sync engine will scan for historical transactions from the given start height.";
+pub(super) const PARAM_START_HEIGHT_DESC: &str = "Block height from which to begin the rescan (default is 0). Only used when rescan is \"yes\" or \"whenkeyisnew\" (for a new key).";
+
+/// Parsed `rescan` parameter for key-import RPCs.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum RescanPolicy {
+    Yes,
+    No,
+    WhenKeyIsNew,
+}
+
+impl RescanPolicy {
+    /// Parses the `rescan` parameter string.
+    ///
+    /// Returns the parsed policy, or an RPC error if the value is invalid.
+    fn parse(rescan: Option<&str>) -> RpcResult<Self> {
+        match rescan {
+            None | Some("whenkeyisnew") => Ok(Self::WhenKeyIsNew),
+            Some("yes") => Ok(Self::Yes),
+            Some("no") => Ok(Self::No),
+            Some(_) => Err(LegacyCode::InvalidParameter.with_static(
+                "Invalid rescan value. Must be \"yes\", \"no\", or \"whenkeyisnew\".",
+            )),
+        }
+    }
+}
+
+/// Decodes a Sapling extended full viewing key and derives the default payment address.
+///
+/// Returns the decoded extended full viewing key and the encoded payment address string.
+fn decode_vkey_and_address(
+    hrp_fvk: &str,
+    hrp_payment_address: &str,
+    vkey: &str,
+) -> RpcResult<(sapling::zip32::ExtendedFullViewingKey, String)> {
+    let extfvk = decode_extended_full_viewing_key(hrp_fvk, vkey).map_err(|e| {
+        LegacyCode::InvalidAddressOrKey.with_message(format!("Invalid viewing key: {e}"))
+    })?;
+
+    let (_, payment_address) = extfvk.default_address();
+
+    let address = encode_payment_address(hrp_payment_address, &payment_address);
+
+    Ok((extfvk, address))
+}
+
+pub(crate) async fn call<C: Chain>(
+    wallet: &mut DbConnection,
+    chain: C,
+    vkey: &str,
+    rescan: Option<&str>,
+    start_height: Option<u64>,
+) -> Response {
+    let rescan = RescanPolicy::parse(rescan)?;
+
+    // Parse start_height if provided, keeping it as Option so we can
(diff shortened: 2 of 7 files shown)


zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,326 @@
...
+};
+
+/// Response to a `z_importviewingkey` RPC request.
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// Result of importing a viewing key.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct ResultType {
+    /// The type of the imported address (always "sapling").
+    address_type: String,
```
nullcopy, line 25:
```suggestion
    type: String,
```

Match zcashd naming

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,326 @@
...
+
+pub(super) const PARAM_VKEY_DESC: &str =
+    "The viewing key (only Sapling extended full viewing keys are supported).";
+pub(super) const PARAM_RESCAN_DESC: &str = "Whether to rescan the blockchain for transactions (\"yes\", \"no\", or \"whenkeyisnew\"; default is \"whenkeyisnew\"). When rescan is enabled, the wallet's background sync engine will scan for historical transactions from the given start height.";
+pub(super) const PARAM_START_HEIGHT_DESC: &str = "Block height from which to begin the rescan (default is 0). Only used when rescan is \"yes\" or \"whenkeyisnew\" (for a new key).";
+
+/// Validates the `rescan` parameter.
+///
+/// Returns the validated rescan value, or an RPC error if the value is invalid.
+fn validate_rescan(rescan: Option<&str>) -> RpcResult<&str> {
```
nullcopy: Please create an `enum` for the rescan value, so that we don't rely on stringly-typed data

zallet-core/src/components/json_rpc/utils.rs
```diff
@@ -44,6 +52,63 @@ pub(super) async fn ensure_wallet_is_unlocked(keystore: &KeyStore) -> RpcResult<
     }
 }
 
+/// Builds an [`AccountBirthday`] by fetching the treestate at the given height.
+///
+/// For height 0 (genesis), returns a birthday with an empty chain state since the
+/// commitment trees are empty. For non-zero heights, fetches the real treestate from
+/// the chain indexer so the sync engine can validate note commitment tree continuity.
+#[cfg(zallet_build = "wallet")]
+pub(super) async fn fetch_account_birthday(
```
nullcopy: This is a useful helper. There's a few places in the codebase that should use this helper now:
* get_new_account.rs:58-92
* recover_accounts.rs:85-125

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,326 @@
...
+                "yes" | "whenkeyisnew" => start_height,
+                "no" => chain_tip.unwrap_or(BlockHeight::from_u32(0)),
+                _ => unreachable!(),
+            };
+
+            let birthday = fetch_account_birthday(wallet, &chain, effective_height).await?;
+
+            wallet
+                .import_account_ufvk(
+                    &format!("Imported Sapling viewing key {}", &address[..16]),
```
nullcopy:
```suggestion
                    &format!("Imported Sapling viewing key {}", &address),
```

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,326 @@
...
+        u32::try_from(start_height.unwrap_or(0))
+            .map_err(|_| LegacyCode::InvalidParameter.with_static("Block height out of range."))?,
+    );
+
+    let chain_tip = wallet
+        .chain_height()
+        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?;
+
+    if let Some(tip) = chain_tip {
+        if start_height > tip {
```
nullcopy: This check should only apply if `rescan="yes"`

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,326 @@
...
+                return Err(LegacyCode::Wallet.with_message(format!(
+                    "The wallet already contains the private key for this viewing key (address: {})",
+                    address
+                )));
+            }
+            // ViewOnly — key already exists, return result.
+            //
+            // TODO: When rescan is "yes" and the key already exists, zcashd would force a
+            // rescan from start_height. We currently skip this because zcash_client_sqlite
+            // does not expose a way to reset scan ranges for an existing account.
```
nullcopy: This TODO is out of date. `zcash_client_backend` recently added `WalletWrite::rewind_to_chain_state` which does the rescan you need here. You'll need to update the crate patch in `Cargo.toml` for `zcash_client_backend` to point to latest main.

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,326 @@
...
+        .map_err(|e| LegacyCode::Wallet.with_message(e.to_string()))?;
+
+    // Check if the key is already known to the wallet.
+    let existing_account = wallet
+        .get_account_for_ufvk(&ufvk)
+        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?;
+    match existing_account {
+        Some(account) => {
+            if matches!(account.purpose(), AccountPurpose::Spending { .. }) {
+                return Err(LegacyCode::Wallet.with_message(format!(
```
nullcopy, line 125:
```suggestion
                    "The wallet already contains the private key for this viewing key (address: {address})",
```

zallet-core/src/components/json_rpc/methods.rs
```diff
@@ -425,6 +427,29 @@ pub(crate) trait WalletRpc {
...
 
+    /// Imports a Sapling viewing key into the wallet.
+    ///
+    /// Only Sapling extended full viewing keys are supported. The wallet will track
+    /// incoming and outgoing transactions for addresses derived from this key, but
+    /// will not have spending authority.
+    ///
+    /// # Arguments
+    ///
+    /// - `vkey` (string, required) The viewing key (see `z_exportviewingkey`).
```
nullcopy, line 472:
```suggestion
    /// - `vkey` (string, required) The viewing key.
```
`z_exportviewingkey` hasn't been implemented yet, so this reference doesn't make sense. IMO just drop the reference for now

nullcopy (changes requested):
Thanks for taking this on! I've left a few requests before this can merge.

Also, I think we need an integration test, before we merge this. I've created an issue (https://github.com/zcash/integration-tests/issues/99). It'd be helpful if you wanted to take that on too, but no obligation ofc 😅 

Reviewed as of dca4cbe6b240b41917c6f3dcf0e1906e7f1f991c

nuttycom:
@mnm458 #614 has merged; you had mentioned you wanted to rebase this to add integration tests?

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
nuttycom: It's annoying that zcashd had such lousy naming but yeah we should match it.

mnm458: I think `address_type` is the canonical field and `type` is deprecated, it's gated behind `fEnableAddrTypeField` and marked deprecated:
https://github.com/zcash/zcash/blob/master/src/wallet/rpcdump.cpp#L799

Should we keep `address_type` as-is, or still switch to `type`?

nuttycom: Ah, okay, I didn't realize that `type` was deprecated, definitely keep `address_type` then.
I probably deprecated that but had forgotten, thanks for catching that.

zallet-core/src/components/json_rpc/utils.rs
mnm458: Agreed this is a good refactor target, but neither call site is a drop-in substitution currently.

To make the helper work for both, it would need to accept an optional `recover_until` parameter and a `ChainView` instead of `Chain` (to avoid redundant snapshots).

I can take this up in a new seperate issue.

zallet-core/src/components/json_rpc/methods.rs
mnm458: `z_exportviewingkey` has been implemented now, please help resolve this comment

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
mnm458: zcashd validates startHeight unconditionally so kept the same behaviour for compatibility:

https://github.com/zcash/zcash/blob/558f686599586f55def3db86955d74d3be44605e/src/wallet/rpcdump.cpp#L858-L859

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
mnm458: Fixed

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
mnm458: Updated the TODO, `rewind_to_chain_state` is already available in the current crate patch and used in `z_import_address`. I've updated the comment to reference it as the mechanism for implementing rescan-on-existing-key.

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
mnm458: Fixed

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,334 @@
...
+            Some("no") => Ok(Self::No),
+            Some(_) => Err(LegacyCode::InvalidParameter
+                .with_static("Invalid rescan value. Must be \"yes\", \"no\", or \"whenkeyisnew\".")),
+        }
+    }
+}
+
+/// Decodes a Sapling extended full viewing key and derives the default payment address.
+///
+/// Returns the decoded viewing key and the encoded payment address string.
```
nullcopy:
```suggestion
/// Returns the decoded extended full viewing key and the encoded payment address string.
```

(nit) I know zcashd used the terminology you're using, but to avoid confusion, let's use "extended full viewing key" and `efvk` where possible

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
```diff
@@ -0,0 +1,334 @@
...
+            //
+            // TODO: When rescan is "yes" and the key already exists, zcashd would force a
+            // rescan from start_height. We could use `WalletWrite::rewind_to_chain_state`
+            // for this (see `z_import_address` for an example).
+        }
+        None => {
+            // new key
+            let effective_height = match rescan {
+                RescanPolicy::Yes | RescanPolicy::WhenKeyIsNew => start_height,
+                RescanPolicy::No => chain_tip.unwrap_or(BlockHeight::from_u32(0)),
```
nullcopy: IMO, if a `start_height` was supplied, we should use that even if `RescanPolicy::No`.

nullcopy (changes requested):
A few minor asks and some CI failures to figure out

zallet-core/src/components/json_rpc/methods/import_viewing_key.rs
mnm458: Resolved, please take a look when you have a chance: integration tests can be found here https://github.com/zcash/integration-tests/pull/181

nullcopy (changes requested):
Concept ACK d32a5b4a6e252b9b43ce0df382b524268603a47b modulo CI fails

mnm458:
Fixed formatting attributed to CI check fail

nullcopy (approved):
utACK 908627392fd748d6b50d8d01d717784ecdd0a0bb pending CI

PR #434, opened 2026-05-04
Gate incompatible alpha wallet databases

### LINKED ISSUE #264: If we make a breaking change in an alpha, detect when loading an earlier alpha's wallet and refuse to run
We will most likely be making breaking changes during the alpha window. We store each version at which we migrated the database, so we can use this to detect (prior to attempting to migrate the database) whether the last database migration ran with a version prior to the latest breaking change. If it is, we should stop and tell the user they need to start again with a fresh Zallet wallet.

czarcas7ic:
Closes #264.

Summary:
- Add an early wallet SQLite compatibility gate before `WalletMigrator` can mutate an existing database.
- Use the latest recorded Zallet version in `ext_zallet_db_version_metadata` to detect databases last opened by an incompatible alpha.
- Reject existing databases whose latest recorded Zallet version is older than `0.1.0-alpha.4`, while preserving the existing network mismatch error path.
- Bump Zallet to `0.1.0-alpha.4` so accepted databases record the compatible version after opening.

Validation:
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test -p zallet components::database::tests --all-features`
- `env -u RUST_LOG cargo test --workspace --all-features`

### DIFF
--- a/zallet/Cargo.toml
+++ b/zallet/Cargo.toml
@@ -126,6 +126,7 @@ schemerz-rusqlite.workspace = true
 secp256k1.workspace = true
 secrecy.workspace = true
 serde.workspace = true
+semver.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
 shadow-rs.workspace = true
--- a/zallet/src/components/database.rs
+++ b/zallet/src/components/database.rs
@@ -3,6 +3,7 @@ use std::fmt;
 use abscissa_core::tracing::info;
 use rusqlite::{OptionalExtension, named_params};
 use schemerz_rusqlite::RusqliteMigration;
+use semver::Version;
 use tokio::fs;
 
 use zcash_client_sqlite::wallet::init::{WalletMigrationError, WalletMigrator};
@@ -26,6 +27,10 @@ mod ext;
 mod tests;
 
 pub(crate) type DbHandle = deadpool::managed::Object<connection::WalletManager>;
+
+// Old databases are accepted only if the last recorded Zallet version is at
+// least the oldest release whose wallet database this binary accepts.
+const MIN_COMPATIBLE_ZALLET_VERSION: &str = "0.1.0-alpha.3";
 
 /// Returns the full list of migrations defined in Zallet, to be applied alongside the
 /// migrations internal to `zcash_client_sqlite`.
@@ -66,30 +71,12 @@ impl Database {
         let handle = database.handle().await?;
 
         if db_exists {
-            // Verify that the database matches the configured network type before we make
-            // any changes (including migrations, some of which make use of the network
-            // params), to avoid leaving the database in an inconsistent state. We can
-            // assume the presence of this table, as it's added by the initial migrations.
-            handle.with_raw(|conn, _| {
-                let wallet_network_type = conn
-                    .query_row(
-                        "SELECT network_type FROM ext_zallet_db_wallet_metadata",
-                        [],
-                        |row| row.get::<_, crate::network::kind::Sql>("network_type"),
-                    )
-                    .map_err(|e| ErrorKind::Init.context(e))?;
-
-                if wallet_network_type.0 == config.consensus.network {
-                    Ok(())
-                } else {
-                    Err(ErrorKind::Init.context(fl!(
-                        "err-init-config-db-mismatch",
-                        db_network_type = crate::network::kind::type_to_str(&wallet_network_type.0),
-                        config_network_type =
-                            crate::network::kind::type_to_str(&config.consensus.network),
-                    )))
-                }
-            })?;
+            // Verify that the database matches the configured network type and was
+            // last opened by a compatible alpha before we make any changes (including
+            // migrations, some of which make use of the network params), to avoid
+            // leaving the database in an inconsistent state. We can assume the network
+            // metadata table is present, as it's added by the initial migrations.
+            handle.with_raw(|conn, _| verify_existing_database(conn, config))?;
 
             info!("Applying latest database migrations");
         } else {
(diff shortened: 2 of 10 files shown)


nullcopy (changes requested):
When I read https://github.com/zcash/wallet/issues/264, I interpret the request to mean to compare the earliest stored Zallet version (in `ext_zallet_db_version_metadata`) against the current running version, and refuse to start if the versions differ by a breaking change (new alpha tag). Am I missing a reason not to just compare version numbers?

The biggest thing that makes me uncomfortable about this approach, is that we would now need to remember to change the compatibility epoch on releases, and we would need to be clear about when/why to change it.

zallet/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zallet"
-version = "0.1.0-alpha.3"
+version = "0.1.0-alpha.4"
```
czarcas7ic: This is not marking alpha.4 as released. The changelog still says `0.1.0-alpha.4 - PLANNED`. The bump is needed because the DB compatibility gate rejects versions older than alpha.4, and successful opens record `crate::build::PKG_VERSION`. If this stayed alpha.3, this binary could record alpha.3 and then reject the same DB on the next open.

czarcas7ic:
@nullcopy That makes sense. I removed the compatibility epoch and switched this to use the existing `ext_zallet_db_version_metadata` table.

One small difference from your read is that I’m comparing the latest recorded Zallet version rather than the earliest one. The issue says:
> the last database migration ran with a version prior to the latest breaking change.

and using the earliest version would reject a DB that was originally created by alpha.3 but later successfully opened by alpha.4.

I also bumped the package version to alpha.4 because a successful open records `crate::build::PKG_VERSION`. If this binary kept reporting alpha.3, it could record alpha.3 and then reject that same DB on the next open.

nullcopy:
Hmm, this still requires us to manually set `MIN_COMPATIBLE_ZALLET_VERSION` to specify which versions are no longer compatible. Instead, we should be inferring incompatibility from the version numbers themselves.

Essentially: get the latest version from `ext_zallet_db_version_metadata` and the current running version of Zallet. Compare the version tag (e.g. the `-alpha.X`) part of the version strings. If either tag is an alpha AND the tags are not identical, refuse to startup. This check should only allow startup iff:
* neither tag is alpha
* both tags are alpha and alpha versions are identical

> One small difference from your read is that I’m comparing the latest recorded Zallet version rather than the earliest one

That seems reasonable to me :+1:

czarcas7ic:
Hey @nullcopy, I think I understand the suggestion. Wouldn’t that make every alpha to alpha upgrade fail this DB compatibility check and show the incompatible DB message, even if that alpha did not actually introduce a DB break?

My read of the issue was that we needed an explicit cutoff for alpha DB compatibility, not that every alpha to alpha upgrade should automatically fail. In this case the cutoff is alpha.4 because of the Zaino DB migration change. If another alpha introduces a DB compatibility break later, we would bump the cutoff again.

If the intended policy is that every alpha tag is a DB compatibility boundary, I can switch this to compare the alpha tags directly.

czarcas7ic:
`Test NU7 on Linux` is failing on main. 

I manually ran `Trigger integration tests` and it passes, seems to fail due to fork not having required secrets.

zallet/src/components/database.rs
```diff
@@ -181,3 +168,70 @@ impl Database {
             .map_err(|e| ErrorKind::Generic.context(e).into())
     }
 }
+
+fn verify_existing_database(
+    conn: &rusqlite::Connection,
+    config: &ZalletConfig,
+) -> Result<(), Error> {
+    verify_wallet_network_type(conn, config)?;
+    verify_alpha_db_compatibility(conn)
```
nullcopy: (nit) better to check alpha db compatibility first, in case the `ext_zallet_db_wallet_metadata` table or its `network_type` column used by `verify_wallet_network_type()` are what become incompatible later

zallet/Cargo.toml
nullcopy: IMO, we should just leave this as-is for now. In the future and set `MIN_COMPATIBLE_ZALLET_VERSION` to `0.1.0-alpha.3` (or just empty string for now?)

In the future, when we release `alpha.4` we will need to update `MIN_COMPATIBLE_ZALLET_VERSION` accordingly.

nullcopy: I won't leave a bunch of comments, but same goes for the versions in the other `.toml`s

zallet/src/components/database.rs
```diff
@@ -181,3 +168,70 @@ impl Database {
...
+                config_network_type = crate::network::kind::type_to_str(&config.consensus.network),
+            ))
+            .into())
+    }
+}
+
+fn verify_alpha_db_compatibility(conn: &rusqlite::Connection) -> Result<(), Error> {
+    let latest_version = latest_recorded_zallet_version(conn)?;
+    let latest_version =
+        Version::parse(&latest_version).map_err(|_| incompatible_alpha_database_error())?;
```
nullcopy: I'd prefer not to blindly treat any parse error as an incompatible version error.

nullcopy (changes requested):
Ok @czarcas7ic, sorry for the misunderstanding! Agreed, this approach makes sense 👍

Concept ACK, but I have a few small changes

nullcopy:
> I manually ran Trigger integration tests and it passes, seems to fail due to fork not having required secrets.

Ah right... for now we have to use local branches. If you don't have perms to create a local branch, then I can create one for you once we've worked through everything else.

czarcas7ic:
@nullcopy addressed all comments, with the understanding that `MIN_COMPATIBLE_ZALLET_VERSION` should be bumped in tandem with the tag for alpha.4

nullcopy (approved):
utACK b017484eceb145db75d83ec5936a3200b4d612a6

nullcopy:
Superseded by https://github.com/zcash/wallet/pull/435

PR #435, opened 2026-05-06
Gate incompatible alpha wallet databases

### LINKED ISSUE #264: If we make a breaking change in an alpha, detect when loading an earlier alpha's wallet and refuse to run

nullcopy:
Copy of https://github.com/zcash/wallet/pull/434 on local branch to allow integration tests

Original description below:

---
Closes #264.

Summary:
- Add an early wallet SQLite compatibility gate before `WalletMigrator` can mutate an existing database.
- Use the latest recorded Zallet version in `ext_zallet_db_version_metadata` to detect databases last opened by an incompatible alpha.
- Reject existing databases whose latest recorded Zallet version is older than `0.1.0-alpha.4`, while preserving the existing network mismatch error path.
- Bump Zallet to `0.1.0-alpha.4` so accepted databases record the compatible version after opening.

Validation:
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test -p zallet components::database::tests --all-features`
- `env -u RUST_LOG cargo test --workspace --all-features`

### DIFF
--- a/zallet/Cargo.toml
+++ b/zallet/Cargo.toml
@@ -126,6 +126,7 @@ schemerz-rusqlite.workspace = true
 secp256k1.workspace = true
 secrecy.workspace = true
 serde.workspace = true
+semver.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
 shadow-rs.workspace = true
--- a/zallet/src/components/database.rs
+++ b/zallet/src/components/database.rs
@@ -3,6 +3,7 @@ use std::fmt;
 use abscissa_core::tracing::info;
 use rusqlite::{OptionalExtension, named_params};
 use schemerz_rusqlite::RusqliteMigration;
+use semver::Version;
 use tokio::fs;
 
 use zcash_client_sqlite::wallet::init::{WalletMigrationError, WalletMigrator};
@@ -26,6 +27,10 @@ mod ext;
 mod tests;
 
 pub(crate) type DbHandle = deadpool::managed::Object<connection::WalletManager>;
+
+// Old databases are accepted only if the last recorded Zallet version is at
+// least the oldest release whose wallet database this binary accepts.
+const MIN_COMPATIBLE_ZALLET_VERSION: &str = "0.1.0-alpha.3";
 
 /// Returns the full list of migrations defined in Zallet, to be applied alongside the
 /// migrations internal to `zcash_client_sqlite`.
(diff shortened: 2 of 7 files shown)


czarcas7ic:
ACKing that this PR matches the hash from my PR, but was opened locally so that the CI job that requires repo secret access can run 👍

czarcas7ic (approved):
Can approve now

zallet/src/components/database.rs
```diff
@@ -181,3 +168,80 @@ impl Database {
             .map_err(|e| ErrorKind::Generic.context(e).into())
     }
 }
+
+fn verify_existing_database(
+    conn: &rusqlite::Connection,
+    config: &ZalletConfig,
+) -> Result<(), Error> {
+    verify_alpha_db_compatibility(conn)?;
+    verify_wallet_network_type(conn, config)
```
pacu, line 177:
```suggestion
    verify_wallet_network_type(conn, config)
    verify_alpha_db_compatibility(conn)?;
```

I think that a network type mismatch is greater than an alpha db mismatch so it should be checked first

zallet/src/components/database/tests.rs
```diff
@@ -80,3 +81,201 @@ fn verify_schema() {
...
+        err.to_string().contains("fresh Zallet wallet"),
+        "unexpected error: {err}",
+    );
+
+    let conn = Connection::open(config.wallet_db_path()).unwrap();
+    assert_eq!(count_recorded_versions(&conn, crate::build::PKG_VERSION), 0);
+}
+
+#[test]
+fn network_mismatch_still_reports_network_error() {
```
pacu, line 179: Add a test that exercises that network mismatch should be checked first than Alpha version

zallet/Cargo.toml
pacu, line 3: [non-blocking] this should probably change to alpha.4

zallet/src/components/database.rs
```diff
@@ -27,6 +28,10 @@ mod tests;
 
 pub(crate) type DbHandle = deadpool::managed::Object<connection::WalletManager>;
 
+// Old databases are accepted only if the last recorded Zallet version is at
+// least the oldest release whose wallet database this binary accepts.
+const MIN_COMPATIBLE_ZALLET_VERSION: &str = "0.1.0-alpha.3";
```
pacu, line 33: [non-blocking] this should probably change to alpha.4

pacu (changes requested):
Tested the flow and it worked. If the comments are addressed this should be good to go. 

Test:
1. checkout alpha.3 tag
2. create a new wallet
3. checkout this branch
4. update versions to alpha.3
5. start the wallet

Expected: start fails because of version mismatch. 
Result: expectation met

zallet/Cargo.toml
czarcas7ic: From the following comment https://github.com/zcash/wallet/pull/434#discussion_r3190094909 we agreed to keep the version as is until it was time for tag, and that as part of the release process (just this once) we also bump `MIN_COMPATIBLE_ZALLET_VERSION`

zallet/src/components/database.rs
czarcas7ic: This was not done for the same reason here https://github.com/zcash/wallet/pull/435#discussion_r3205298526

zallet/src/components/database.rs
nullcopy: I actually asked for this ordering, because the network type check depends on information in the database. If the database is incompatible, we can't expect the network type check to be sane. Although I agree with you semantically, practically I think the db compatibility check ~~has to~~ should come first.

zallet/src/components/database/tests.rs
czarcas7ic: https://github.com/zcash/wallet/pull/435/commits/b447fd681310d2c480fcee0135bdfe7a171a8750

Added test here, but currently tests the existing ordering (version then network)

pacu (approved):

PR #438, opened 2026-05-10
rpc: Add cookie file authentication for the JSON-RPC interface

### LINKED ISSUE #305: rpc: Add cookie file authorization method to the JSON-RPC interface
Deferred from #210 because there's enough annoying things to figure out about the config side that I don't want to block that on it.

mnm458:
Generate an RPC cookie on startup and write it to {datadir}/.cookie. The server loads that credential into the auth layer alongside any configured users, and removes the cookie file on clean shutdown.

When no RPC password is configured, `zallet rpc` now reads the cookie file so local RPC works without extra setup.

Closes #305

### DIFF
--- a/zallet/src/commands/rpc_cli.rs
+++ b/zallet/src/commands/rpc_cli.rs
@@ -4,11 +4,17 @@ use std::fmt;
 use std::time::Duration;
 
 use abscissa_core::Runnable;
+use base64ct::{Base64, Encoding};
+use hyper::header::{AUTHORIZATION, HeaderMap, HeaderValue};
 use jsonrpsee::core::{client::ClientT, params::ArrayParams};
 use jsonrpsee_http_client::HttpClientBuilder;
 use secrecy::{ExposeSecret, SecretString};
+use tracing::warn;
 
-use crate::{cli::RpcCliCmd, commands::AsyncRunnable, error::Error, prelude::*};
+use crate::{
+    cli::RpcCliCmd, commands::AsyncRunnable, components::json_rpc::server::cookie, error::Error,
+    fl, prelude::*,
+};
 
 const DEFAULT_HTTP_CLIENT_TIMEOUT: u64 = 900;
 
@@ -43,31 +49,51 @@ impl AsyncRunnable for RpcCliCmd {
             None => DEFAULT_HTTP_CLIENT_TIMEOUT,
         });
 
-        // Find a password we can use. If none are configured, we assume none is needed.
-        let auth_prefix = config
+        // Find credentials: prefer configured password, fall back to cookie file.
+        let credentials = config
             .rpc
             .auth
             .iter()
             .find_map(|auth| {
                 auth.password
                     .as_ref()
-                    .map(|pw| SecretString::new(format!("{}:{}@", auth.user, pw.expose_secret())))
+                    .map(|pw| SecretString::new(format!("{}:{}", auth.user, pw.expose_secret())))
             })
-            .unwrap_or_else(|| SecretString::new(String::new()));
+            .or_else(|| {
+                // Fall back to cookie-based auth.
+                cookie::read_cookie(config.datadir())
+                    .map_err(|err| {
+                        warn!("{}", fl!("rpc-cookie-read-failed", error = err.to_string()));
+                    })
+                    .ok()
+                    .map(SecretString::new)
+            });
+
+        // Build auth header if credentials are available.
+        let mut headers = HeaderMap::new();
+        if let Some(creds) = &credentials {
+            let encoded = Base64::encode_string(creds.expose_secret().as_bytes());
+            let mut value = HeaderValue::from_str(&format!("Basic {encoded}"))
+                .map_err(|_| RpcCliError::FailedToConnect)?;
+            value.set_sensitive(true);
+            headers.insert(AUTHORIZATION, value);
+        }
 
         // Connect to the Zallet wallet.
         let client = match config.rpc.bind.as_slice() {
             &[] => Err(RpcCliError::WalletHasNoRpcServer),
             &[bind] => HttpClientBuilder::default()
                 .request_timeout(timeout)
-                .build(format!("http://{}{bind}", auth_prefix.expose_secret()))
+                .set_headers(headers.clone())
+                .build(format!("http://{bind}"))
                 .map_err(|_| RpcCliError::FailedToConnect),
             addrs => addrs
                 .iter()
                 .find_map(|bind| {
                     HttpClientBuilder::default()
                         .request_timeout(timeout)
-                        .build(format!("http://{}{bind}", auth_prefix.expose_secret()))
+                        .set_headers(headers.clone())
+                        .build(format!("http://{bind}"))
                         .ok()
                 })
                 .ok_or(RpcCliError::FailedToConnect),
--- a/zallet/src/components/json_rpc/server/cookie.rs
+++ b/zallet/src/components/json_rpc/server/cookie.rs
@@ -1,0 +1,196 @@
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use base64ct::{Base64, Encoding};
+use rand::{Rng, rngs::OsRng};
+use tracing::{info, warn};
+
+use super::authorization::PasswordHash;
+use crate::error::{Error, ErrorKind};
+use crate::fl;
+
+/// Username for cookie-based auth, matching zcashd convention.
+pub(crate) const COOKIE_USER: &str = "__cookie__";
+
+/// Default cookie filename within the data directory.
+const COOKIE_FILENAME: &str = ".cookie";
+
+/// Guard that deletes the cookie file when dropped.
+///
+/// This ensures the cookie is cleaned up on normal shutdown or task cancellation.
+pub(crate) struct CookieGuard {
+    path: PathBuf,
+}
+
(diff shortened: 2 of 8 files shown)


zallet/src/commands/rpc_cli.rs
```diff
@@ -43,31 +49,51 @@ impl AsyncRunnable for RpcCliCmd {
...
                     .as_ref()
-                    .map(|pw| SecretString::new(format!("{}:{}@", auth.user, pw.expose_secret())))
+                    .map(|pw| SecretString::new(format!("{}:{}", auth.user, pw.expose_secret())))
             })
-            .unwrap_or_else(|| SecretString::new(String::new()));
+            .or_else(|| {
+                // Fall back to cookie-based auth.
+                cookie::read_cookie(config.datadir())
+                    .map_err(|err| {
+                        warn!("Failed to read cookie file: {}", err);
```
nullcopy: When generating logs/error strings/etc, please use the `fl!` macro to be consistent with our i18n. Also see the `wfl!` and `wlnfl!` helper macros at the top of this file, where applicable.

zallet/src/components/json_rpc/server/cookie.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[cfg(feature = "rpc-cli")]
+pub(crate) fn read_cookie(datadir: &Path) -> Result<String, Error> {
+    let path = datadir.join(COOKIE_FILENAME);
+    let cookie = fs::read_to_string(&path)
+        .map_err(|e| ErrorKind::Init.context(e))?
+        .trim()
+        .to_string();
+    if !cookie.starts_with("__cookie__:") {
```
nullcopy: Please use `COOKIE_USER` here

zallet/src/components/json_rpc/server/cookie.rs
```diff
@@ -0,0 +1,168 @@
...
+        let _guard = generate_cookie(dir.path()).unwrap();
+        assert!(dir.path().join(COOKIE_FILENAME).exists());
+    }
+
+    #[test]
+    fn generate_file_contents_valid() {
+        let dir = TempDir::new().unwrap();
+        let _guard = generate_cookie(dir.path()).unwrap();
+        let contents = fs::read_to_string(dir.path().join(COOKIE_FILENAME)).unwrap();
+        assert!(contents.starts_with("__cookie__:"));
```
nullcopy: Please use `COOKIE_USER` here

zallet/src/components/json_rpc/server/cookie.rs
```diff
@@ -0,0 +1,168 @@
...
+        assert!(contents.len() > "__cookie__:".len());
```
nullcopy: Please use `COOKIE_USER` here

zallet/src/components/json_rpc/server/cookie.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+pub(crate) fn generate_cookie(
+    datadir: &Path,
+) -> Result<(String, PasswordHash, CookieGuard), Error> {
+    let password: [u8; 32] = OsRng.r#gen();
+    let password = Base64::encode_string(&password);
+    let cookie = format!("{COOKIE_USER}:{password}");
+
+    let cookie_path = datadir.join(COOKIE_FILENAME);
```
nullcopy: (nit) would be nice to wrap this in a `cookie_path()` getter, similar to our other path getters (e.g. `wallet_db_path()`)

zallet/src/components/json_rpc/server/cookie.rs
```diff
@@ -0,0 +1,168 @@
...
+        .trim()
+        .to_string();
+    if !cookie.starts_with("__cookie__:") {
+        return Err(ErrorKind::Init.context("Invalid cookie file format").into());
+    }
+    Ok(cookie)
+}
+
+#[cfg(test)]
+mod tests {
```
nullcopy, line 112: There's a few test gaps to close:
* no test for the invalid-cookie-format branch
* no test for the "configured user named `__cookie__` -> skip generation" branch

zallet/src/commands/rpc_cli.rs
```diff
@@ -43,31 +49,51 @@ impl AsyncRunnable for RpcCliCmd {
...
             .find_map(|auth| {
                 auth.password
                     .as_ref()
-                    .map(|pw| SecretString::new(format!("{}:{}@", auth.user, pw.expose_secret())))
+                    .map(|pw| SecretString::new(format!("{}:{}", auth.user, pw.expose_secret())))
             })
-            .unwrap_or_else(|| SecretString::new(String::new()));
+            .or_else(|| {
+                // Fall back to cookie-based auth.
+                cookie::read_cookie(config.datadir())
```
nullcopy, line 64: Please also add a note about this (in a separate commit) in `book/src/cli/rpc.md`

zallet/src/components/json_rpc/server/cookie.rs
```diff
@@ -0,0 +1,168 @@
...
+) -> Result<(String, PasswordHash, CookieGuard), Error> {
+    let password: [u8; 32] = OsRng.r#gen();
+    let password = Base64::encode_string(&password);
+    let cookie = format!("{COOKIE_USER}:{password}");
+
+    let cookie_path = datadir.join(COOKIE_FILENAME);
+    let tmp_path = datadir.join(format!("{COOKIE_FILENAME}.tmp"));
+
+    // Write to temp file first for atomic creation.
+    fs::write(&tmp_path, cookie.as_bytes()).map_err(|e| ErrorKind::Init.context(e))?;
```
nullcopy: This file is briefly world-readable. Instead of creating it and then restricting its perms, we should create the file restricted from the start. e.g:
```rust
     use std::os::unix::fs::OpenOptionsExt;
     let mut f = fs::OpenOptions::new()
         .write(true).create_new(true).mode(0o600)
         .open(&tmp_path)?;
     f.write_all(cookie.as_bytes())?;
```

zallet/src/commands/rpc_cli.rs
```diff
@@ -43,31 +49,51 @@ impl AsyncRunnable for RpcCliCmd {
...
+            headers.insert(AUTHORIZATION, value);
+        }
 
         // Connect to the Zallet wallet.
         let client = match config.rpc.bind.as_slice() {
             &[] => Err(RpcCliError::WalletHasNoRpcServer),
             &[bind] => HttpClientBuilder::default()
                 .request_timeout(timeout)
-                .build(format!("http://{}{bind}", auth_prefix.expose_secret()))
+                .set_headers(headers.clone())
```
nullcopy, line 87: This is a good fix, but it warrants a `change` entry in the CHANGELOG

nullcopy (changes requested):
Thanks for taking this on! Overall this looks good, but I've left some requests we need to address before merge.

Reviewed as of cb697db20aa49f1c774f4756c54d2ab86f78174d

zallet/src/commands/rpc_cli.rs
mnm458: Done. Left the Drop impl warning as a raw string since fl! allocates and Drop can run during unwinding.

zallet/src/components/json_rpc/server/cookie.rs
mnm458: Fixed across production code and tests.

zallet/src/components/json_rpc/server/cookie.rs
mnm458: Done

zallet/src/components/json_rpc/server/cookie.rs
mnm458: Added:                                                                                                                                                                                                                                            

- `read_invalid_format_errors` covers the format validation branch in `read_cookie`
- `configured_cookie_user_works_without_generated_cookie` exercises `AuthorizationLayer::new` with a configured `__cookie__` user and `cookie: None`
- `generated_cookie_hash_validates_against_file` end-to-end proof that the hash returned by `generate_cookie` validates against the password written to disk

zallet/src/commands/rpc_cli.rs
mnm458: Done

zallet/src/components/json_rpc/server/cookie.rs
mnm458: Switched to `OpenOptions::new().create_new(true).mode(0o600)` so the file is born restricted. Went with per-file mode rather than process-wide umask (the Bitcoin Core approach) since it's self-contained and doesn't require coordinating a global umask at startup. Added a `remove_file` before `create_new` to handle leftover tmp files from a previous crash since `O_EXCL` would otherwise fail. Non-unix path falls back to `fs::write`, consistent with how Bitcoin Core handles Windows.

zallet/src/commands/rpc_cli.rs
mnm458: Done

nullcopy (approved):
Looks good to me! Thank you 🙏 
utACK 50803d60089e967012a3f81f338d1d54f0f8b4b5

mnm458:
Rebased on latest main, merge conflict resolved.

nullcopy:
Subsumed by https://github.com/zcash/wallet/pull/502

PR #452, opened 2026-06-02
Makefile: delete duplicate .PHONY for build and import

dannywillems:
.PHONY for the targets `build` and `import` are already appended later.

Duplications don't hurt, but they are useless.


nullcopy (approved):
utACK 85795dd2a0b61911e80d38700d9f48a5a95fa403

PR #455, opened 2026-06-04
sync: Treat an absent shielded tree state as an empty note commitment tree

### LINKED ISSUE #454: wallet sync task crashes with "Missing Orchard tree state" when a shielded pool is active but its tree is still empty
### Summary
`fetch_chain_state` (`zallet/src/components/sync/steps.rs`) returns a `"Missing Orchard tree state"` error — which exits the wallet sync task — whenever a shielded pool is active at the requested height but the indexer reports no tree state for it. This happens on any chain where a pool is active while its note commitment tree is still empty, most reliably on **regtest**, where NU5 is active from height 1 but the Orchard tree has no commitments yet.

### Steps to reproduce
1. Run zallet against a regtest node + Zaino (NU5 active from height 1).
2. Start wallet sync over the early regtest chain (Orchard tree empty).
3. Sync task exits with:
   ```
   IndexerError::InvalidData { message: "Missing Orchard tree state" }
   ```

(Reproduced in the integration-tests suite: the `wallet.py` and `wallet_orchard_init.py` regtest scenarios.)

### Expected behaviour
An empty Orchard (or Sapling) tree is a valid state; sync should proceed treating it as an empty note commitment tree, not crash.

### Root cause
`fetch_chain_state` assumes that if a pool is active, the indexer always returns `Some(tree_state)`:

```rust
let final_orchard_tree = if params.is_nu_active(NetworkUpgrade::Nu5, height.0.into()) {
    read_frontier_v0(orchard.ok_or_else(|| IndexerError::InvalidData {
        message: "Missing Orchard tree state".into(),
    })?.as_slice()) ...
} else {
    Frontier::empty()
};
```

But an empty incremental-Merkle tree has no frontier, so the node reports `None` for it: `zebra_rpc::z_get_treestate` builds the field via `tree.map(|t| t.to_rpc_bytes())`, and `zebra-state`'s `read::orchard_tree()` returns `None` when no tree is stored at that height. Zaino faithfully relays this `None`. So `None` ⟺ "empty tree" — it is **not** an error signal (genuine fetch failures surface via the `?` on `z_get_treestate`). Treating `None` as fatal is therefore incorrect.

This is a zallet-side issue; Zaino/Zebra are behaving correctly (there is no empty frontier to return).

### Scope
The **Sapling** path has the identical latent bug (`sapling.ok_or_else(|| "Missing Sapling tree state")?`).

### Proposed fix
Treat an absent tree state the same as an inactive pool — an empty frontier — for both pools. Branch (no PR yet): `fix/empty-shielded-tree-state`.

oxarbitrage:
Closes #454

## Problem
`fetch_chain_state` errored with `"Missing Sapling/Orchard tree state"` — exiting the wallet sync task — whenever a shielded pool was active at the requested height but the indexer returned no tree state for it. This crashes sync on any chain where a pool is active while its note commitment tree is still empty, most reliably on **regtest** (NU5 active from height 1, Orchard tree empty).

## Root cause
An absent tree state is not an error: an empty incremental-Merkle tree has no frontier to serialize, so the node reports `None` for it (`zebra_rpc::z_get_treestate` maps the tree via `tree.map(|t| t.to_rpc_bytes())`, and `zebra-state`'s `read::orchard_tree()` returns `None` when no tree is stored at that height). Zaino relays this `None` faithfully. Genuine fetch failures surface separately via the `?` on `z_get_treestate`. So `None` ⟺ "empty tree", and treating it as fatal was incorrect — this is a zallet-side bug; Zaino/Zebra are behaving correctly.

## Fix
Treat "pool active but tree state absent" the same as "pool not yet active": an empty frontier. Both the Sapling and Orchard paths had this bug; both are fixed symmetrically.

## Testing
`fetch_chain_state` requires a live indexer connection and has no unit coverage. Validated end-to-end against the integration-tests suite — the `wallet.py` and `wallet_orchard_init.py` regtest scenarios sync to completion with this change (previously they hit `"Missing Orchard tree state"`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ be considered breaking changes.
   standalone imported transparent keys (e.g. from a `zcashd` migration).
 - No longer crashes in regtest mode when a Sapling or NU5 activation height is
   not defined.
+- The wallet sync task no longer exits with a "Missing Sapling/Orchard tree
+  state" error when a shielded pool is active but its note commitment tree is
+  still empty (e.g. on regtest, where NU5 is active from height 1). An absent
+  tree state is now correctly treated as an empty tree.
 - Zallet now refuses to open wallet databases from incompatible earlier alpha
   releases instead of attempting to migrate them.
 - `z_sendmany` no longer drop standalone transparent signing keys when the same
--- a/zallet/src/components/sync/steps.rs
+++ b/zallet/src/components/sync/steps.rs
@@ -309,36 +309,25 @@ pub(super) async fn fetch_chain_state(
         .await?
         .into_parts();
 
-    let final_sapling_tree = if params.is_nu_active(NetworkUpgrade::Sapling, height.0.into()) {
-        read_frontier_v0(
-            sapling
-                .ok_or_else(|| IndexerError::InvalidData {
-                    message: "Missing Sapling tree state".into(),
-                })?
-                .as_slice(),
-        )
-        .map_err(|e| IndexerError::InvalidData {
-            message: format!("{e}"),
-        })?
-    } else {
-        // Sapling is not yet active; the Sapling tree is empty.
-        Frontier::empty()
+    // An absent tree state (`None`) means the pool's note commitment tree is empty, not an
+    // error (fetch errors surface via `?` above). Treat "active but empty" like "not yet
+    // active": an empty frontier.
+    let final_sapling_tree = match sapling {
+        Some(sapling) if params.is_nu_active(NetworkUpgrade::Sapling, height.0.into()) => {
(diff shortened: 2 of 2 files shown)


oxarbitrage:
FYI: this is no longer blocking integration-tests. After zcash/integration-tests#104 aligned regtest NU5 activation to height 1, the wallet suite syncs to completion on `main` without this change, so CI no longer hits "Missing Orchard tree state."

That said, I think this is still worth fixing: an absent tree state (`None`) means the pool's note-commitment tree is empty, not an error, and the current code treats it as fatal. zcash/integration-tests#104 made the crash stop reproducing in our harness, but the underlying assumption (None ⟺ empty) holds on any chain, so it's worth landing as a small robustness fix (symmetric across Sapling and Orchard).

Rebased onto current `main`.

PR #465, opened 2026-06-10
fix(rpc): bound async operation queue to prevent memory DoS

ouicate:
## Summary

- Confirms and fixes GHSA-3wr9-v982-59fj: `WalletRpcImpl::async_ops` was an unbounded `Vec<AsyncOperation>` with no automatic pruning.
- Prunes finished operations (`success`, `failed`, `cancelled`) before enqueueing new async RPCs.
- Rejects new async RPCs when 64 in-flight operations are already queued, matching Bitcoin Core's `DEFAULT_HTTP_WORKQUEUE` used by `zcashd`'s RPC layer.
- Checks the limit **before** spawning the background Tokio task, so rejected calls do not leak work.

## Vulnerability analysis

An authenticated JSON-RPC caller (HTTP basic auth) could call `z_sendmany` or `z_shieldcoinbase` repeatedly without calling `z_getoperationresult`. Each call appended an `AsyncOperation` retaining full RPC parameters (`ContextInfo.params` as `JsonValue`) until explicitly removed. There was no maximum size, expiration, or background cleanup — enabling linear memory growth until OOM.

### DIFF
--- a/zallet-core/src/components/json_rpc/asyncop.rs
+++ b/zallet-core/src/components/json_rpc/asyncop.rs
@@ -15,6 +15,12 @@ use uuid::Uuid;
 
 use super::server::LegacyCode;
 
+/// Maximum number of in-flight async operations.
+///
+/// Matches Bitcoin Core's `DEFAULT_HTTP_WORKQUEUE` (64), which bounds the depth of
+/// the RPC work queue in `zcashd`.
+pub(super) const MAX_ASYNC_OPERATIONS: usize = 64;
+
 /// An async operation ID.
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Deserialize, Serialize, Documented, JsonSchema)]
 #[serde(try_from = "String")]
@@ -71,6 +77,10 @@ impl OperationState {
             "success" => Some(Self::Success),
             _ => None,
         }
+    }
+
+    pub(super) fn is_finished(self) -> bool {
+        matches!(self, Self::Success | Self::Failed | Self::Cancelled)
     }
 }
 
--- a/zallet-core/src/components/json_rpc/methods.rs
+++ b/zallet-core/src/components/json_rpc/methods.rs
@@ -11,7 +11,10 @@ use crate::components::{
 
 #[cfg(zallet_build = "wallet")]
 use {
-    super::asyncop::{AsyncOperation, ContextInfo, OperationId},
+    super::{
+        asyncop::{AsyncOperation, ContextInfo, MAX_ASYNC_OPERATIONS, OperationId},
+        server::LegacyCode,
+    },
     crate::components::{
         json_rpc::payments::AmountParameter, keystore::KeyStore, sync::WalletDecryptorHandle,
     },
@@ -779,16 +782,23 @@ impl<C: Chain> WalletRpcImpl<C> {
         self.general.chain().await
     }
 
-    async fn start_async<F, T>(&self, (context, f): (Option<ContextInfo>, F)) -> OperationId
+    async fn start_async<F, T>(
+        &self,
+        (context, f): (Option<ContextInfo>, F),
+    ) -> RpcResult<OperationId>
     where
         F: Future<Output = RpcResult<T>> + Send + 'static,
         T: Serialize + Send + 'static,
     {
         let mut async_ops = self.async_ops.write().await;
+        get_operation::prune_finished(&mut async_ops).await;
+        if async_ops.len() >= MAX_ASYNC_OPERATIONS {
+            return Err(LegacyCode::Misc.with_static("Too many async operations in progress"));
+        }
         let op = AsyncOperation::new(context, f).await;
         let op_id = op.operation_id().clone();
         async_ops.push(op);
-        op_id
+        Ok(op_id)
     }
 }
 
@@ -1082,21 +1092,20 @@ impl<C: Chain> WalletRpcServer for WalletRpcImpl<C> {
         fee: Option<JsonValue>,
         privacy_policy: Option<String>,
     ) -> z_send_many::Response {
-        Ok(self
-            .start_async(
-                z_send_many::call(
-                    self.wallet().await?,
-                    self.keystore.clone(),
-                    self.chain().await?,
-                    fromaddress,
-                    amounts,
-                    minconf,
-                    fee,
-                    privacy_policy,
-                )
-                .await?,
+        self.start_async(
+            z_send_many::call(
+                self.wallet().await?,
+                self.keystore.clone(),
+                self.chain().await?,
+                fromaddress,
+                amounts,
+                minconf,
+                fee,
+                privacy_policy,
             )
-            .await)
+            .await?,
+        )
+        .await
     }
 
     async fn z_shieldcoinbase(
@@ -1120,7 +1129,7 @@ impl<C: Chain> WalletRpcServer for WalletRpcImpl<C> {
             privacy_policy,
         )
         .await?;
-        let opid = self.start_async((context, fut)).await;
+        let opid = self.start_async((context, fut)).await?;
         Ok(z_shieldcoinbase::ShieldCoinbaseResult::new(preflight, opid))
     }
 }
--- a/zallet-core/src/components/json_rpc/methods/get_operation.rs
+++ b/zallet-core/src/components/json_rpc/methods/get_operation.rs
@@ -5,9 +5,7 @@ use jsonrpsee::core::RpcResult;
 use schemars::JsonSchema;
 use serde::Serialize;
 
-use crate::components::json_rpc::asyncop::{
-    AsyncOperation, OperationId, OperationState, OperationStatus,
-};
+use crate::components::json_rpc::asyncop::{AsyncOperation, OperationId, OperationStatus};
 
 /// Response to a `z_getoperationstatus` or `z_getoperationresult` RPC request.
 pub(crate) type Response = RpcResult<ResultType>;
@@ -19,6 +17,18 @@ pub(crate) struct ResultType(Vec<OperationStatus>);
 
 pub(super) const PARAM_OPERATIONID_DESC: &str = "A list of operation ids we are interested in.";
 pub(super) const PARAM_OPERATIONID_REQUIRED: bool = false;
+
+/// Removes finished operations from the queue.
+pub(super) async fn prune_finished(async_ops: &mut Vec<AsyncOperation>) {
+    let mut pending = Vec::with_capacity(async_ops.len());
+    for op in async_ops.drain(..) {
+        if op.state().await.is_finished() {
+            continue;
+        }
+        pending.push(op);
+    }
+    *async_ops = pending;
+}
 
 pub(crate) async fn status(
     async_ops: &[AsyncOperation],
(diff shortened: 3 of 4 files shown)


ouicate:
Please review @nullcopy

zallet-core/src/components/json_rpc/methods/get_operation.rs
```diff
@@ -20,6 +18,18 @@ pub(crate) struct ResultType(Vec<OperationStatus>);
 pub(super) const PARAM_OPERATIONID_DESC: &str = "A list of operation ids we are interested in.";
 pub(super) const PARAM_OPERATIONID_REQUIRED: bool = false;
 
+/// Removes finished operations from the queue.
+pub(super) async fn prune_finished(async_ops: &mut Vec<AsyncOperation>) {
```
nuttycom, line 22: **Result loss:** this prunes *all* finished operations whenever any new async RPC starts, whether or not the caller has collected their results. Previously (and in zcashd) a finished operation persisted until `z_getoperationresult` explicitly consumed it.

Concrete failure: a client calls `z_sendmany` (op A), it completes with the txid in its result; the client then calls `z_sendmany` again (op B). This prune drops A before B is pushed, so a subsequent `z_getoperationresult([A])` / `z_getoperationstatus([A])` returns an empty array and the txid of an already-broadcast transaction is permanently unretrievable over RPC. Any "submit N sends, then collect all opids" workflow loses every result but the last, and status polling can no longer distinguish "never existed" from "finished and pruned".

Bounding memory shouldn't require discarding uncollected results — see the review summary for an eviction policy that preserves the retention contract.

zallet-core/src/components/json_rpc/methods.rs
```diff
@@ -779,16 +782,23 @@ impl<C: Chain> WalletRpcImpl<C> {
...
+        &self,
+        (context, f): (Option<ContextInfo>, F),
+    ) -> RpcResult<OperationId>
     where
         F: Future<Output = RpcResult<T>> + Send + 'static,
         T: Serialize + Send + 'static,
     {
         let mut async_ops = self.async_ops.write().await;
+        get_operation::prune_finished(&mut async_ops).await;
+        if async_ops.len() >= MAX_ASYNC_OPERATIONS {
```
nuttycom, line 795: **Permanent liveness wedge:** there is no execution timeout, age-based eviction, or cancellation path for in-flight operations (`OperationState::Cancelled` is unreachable — no RPC sets it). If 64 operations get stuck in `Ready`/`Executing` (e.g. futures blocked on a resource that never resolves), `prune_finished` never reclaims them and every subsequent async RPC is rejected for the lifetime of the process. The memory-DoS fix becomes a permanent liveness-DoS with no recovery path short of a restart.

zallet-core/src/components/json_rpc/asyncop.rs
```diff
@@ -15,6 +15,12 @@ use uuid::Uuid;
 
 use super::server::LegacyCode;
 
+/// Maximum number of in-flight async operations.
+///
+/// Matches Bitcoin Core's `DEFAULT_HTTP_WORKQUEUE` (64), which bounds the depth of
+/// the RPC work queue in `zcashd`.
+pub(super) const MAX_ASYNC_OPERATIONS: usize = 64;
```
nuttycom, line 22: **The cited justification is factually wrong on both counts:**

1. zcashd's `DEFAULT_HTTP_WORKQUEUE` is **16**, not 64 ([zcash/zcash src/httpserver.h](https://github.com/zcash/zcash/blob/master/src/httpserver.h)). 64 is Bitcoin Core's value only since 29.0 (bitcoin/bitcoin#31215, 2025), which zcashd never took.
2. `DEFAULT_HTTP_WORKQUEUE` bounds the transient queue of pending HTTP requests awaiting worker threads — drained in milliseconds — not a table of long-lived operations that can each run for minutes. zcashd's actual async operation table (`asyncrpcqueue.cpp`) has no size limit at all; entries are removed only when `z_getoperationresult` collects them.

So 64 is essentially arbitrary here. Given ops can be long-lived, consider whether 64 concurrent in-flight sends is enough for batch workloads, and consider making it configurable — `BuilderLimitsSection` in config.rs is the existing precedent for exactly this category of anti-memory-exhaustion knob.

zallet-core/src/components/json_rpc/methods.rs
```diff
@@ -1082,21 +1092,20 @@ impl<C: Chain> WalletRpcServer for WalletRpcImpl<C> {
...
-                    self.keystore.clone(),
-                    self.chain().await?,
-                    fromaddress,
-                    amounts,
-                    minconf,
-                    fee,
-                    privacy_policy,
-                )
-                .await?,
+        self.start_async(
```
nuttycom, line 1095: **The cap bounds memory but not work:** `z_send_many::call(...)` runs `propose_transfer` (full input selection over the wallet DB) and `keystore.decrypt_seed` eagerly, before `start_async` ever checks the cap. So when the queue is full, each rejected flood request has already paid the full proposal cost plus a keystore decryption — the PR description's "rejected calls do not leak work" only covers the tokio spawn. A cheap queue-fullness pre-check before invoking `call()` (accepting the small TOCTOU window) would make rejection actually cheap.

zallet-core/src/components/json_rpc/methods.rs
```diff
@@ -779,16 +782,23 @@ impl<C: Chain> WalletRpcImpl<C> {
...
-    async fn start_async<F, T>(&self, (context, f): (Option<ContextInfo>, F)) -> OperationId
+    async fn start_async<F, T>(
+        &self,
+        (context, f): (Option<ContextInfo>, F),
+    ) -> RpcResult<OperationId>
     where
         F: Future<Output = RpcResult<T>> + Send + 'static,
         T: Serialize + Send + 'static,
     {
         let mut async_ops = self.async_ops.write().await;
```
nuttycom, line 793: **Lock held across awaits:** the exclusive `async_ops` write lock is now held across `prune_finished`, which awaits `op.state()` sequentially for up to 64 ops (each acquiring that op's `data` RwLock, contending with the executing task's own state-transition writes). While parked at any of those yields, `z_getoperationstatus`, `z_listoperationids`, `z_getoperationresult`, and all other submissions block on this lock — a single submission can stall all status polling, and rejected flood requests still force the full scan under the write lock.

Cheaper shape: under a *read* lock, snapshot each op's `Arc` data handle and compute the finished-id set outside the global lock; then take the write lock briefly for a synchronous `retain` + capacity check + push. That also avoids the unconditional `Vec` reallocation in the common nothing-to-prune case.

zallet-core/src/components/json_rpc/methods/get_operation.rs
```diff
@@ -20,6 +18,18 @@ pub(crate) struct ResultType(Vec<OperationStatus>);
...
+    let mut pending = Vec::with_capacity(async_ops.len());
```
nuttycom, line 23: **Queue invariants owned by nobody:** this duplicates the eviction semantics already in `result()` below (gather finished via `is_finished`, then remove) with a divergent idiom, and it lives in the `z_getoperationstatus`/`z_getoperationresult` method module even though it's queue maintenance called only from `start_async`. The capacity/retention rules are now enforced ad hoc across three sites on a raw `RwLock<Vec<AsyncOperation>>`, so a future method that touches `async_ops` can silently regress the bound. Consider a dedicated queue type (in `asyncop.rs`) that owns capacity, pruning policy, and result retention as invariants.

zallet-core/src/components/json_rpc/methods/get_operation.rs
```diff
@@ -67,3 +74,47 @@ fn filtered(
...
+        let mut ops = vec![finished, pending];
+        prune_finished(&mut ops).await;
+
+        assert_eq!(ops.len(), 1);
+        assert_eq!(ops[0].operation_id(), &pending_id);
+    }
+
+    #[test]
+    fn max_async_operations_matches_bitcoin_rpc_workqueue() {
+        assert_eq!(MAX_ASYNC_OPERATIONS, 64);
```
nuttycom, line 118: This test is tautological — it restates the constant's definition, so it can't catch a real regression; it just has to be edited in lockstep whenever the cap is retuned (and the name asserts a relationship to Bitcoin's workqueue that the code doesn't actually have — see the comment on the constant). Suggest deleting it; the `prune_finished` test above is the one carrying weight.

zallet-core/src/components/json_rpc/methods.rs
```diff
@@ -779,16 +782,23 @@ impl<C: Chain> WalletRpcImpl<C> {
...
+        get_operation::prune_finished(&mut async_ops).await;
+        if async_ops.len() >= MAX_ASYNC_OPERATIONS {
+            return Err(LegacyCode::Misc.with_static("Too many async operations in progress"));
```
nuttycom, line 796: Minor: `LegacyCode::Misc` (-1) makes queue-full indistinguishable from any generic failure, so clients can't implement backoff-and-retry for what is a transient, retriable condition. A more specific code from the existing `LegacyCode` enum would let callers tell them apart.

nuttycom (commented):
Bounding the queue is the right goal, but as written this trades the memory-DoS for a worse defect: it silently destroys the results of completed operations before the caller can collect them, which for a wallet means losing the txid of a broadcast transaction. Summary of the issues detailed in the inline comments:

1. **Result loss (most severe):** `prune_finished` runs on every new async RPC and deletes *all* finished operations, collected or not. zcashd retains a finished operation until `z_getoperationresult` explicitly collects it (`asyncrpcqueue.cpp` removes entries only via `popOperationForId()`). A batch script that submits N sends and then polls loses every result except the last, and `z_getoperationstatus` polling becomes non-idempotent.
2. **No timeout or eviction for in-flight ops:** an operation that never finishes occupies a slot forever; there is no cancellation RPC (`OperationState::Cancelled` is unreachable), so 64 wedged operations permanently reject all future async RPCs.
3. **The 64 constant's justification is factually wrong:** zcashd's `DEFAULT_HTTP_WORKQUEUE` is 16, not 64 (64 is Bitcoin Core ≥29.0, via bitcoin/bitcoin#31215, which zcashd never took), and that constant bounds the transient HTTP request queue, not the async operation table — which in zcashd is unbounded. The limit is also hardcoded rather than joining the existing configurable-limits surface (cf. `BuilderLimitsSection`).
4. **The cap bounds memory but not work:** `propose_transfer` and `keystore.decrypt_seed` run eagerly in `call()` before `start_async` checks the cap, so each rejected flood request still pays full input selection plus a keystore decryption.
5. **Lock-hold regression:** the global `async_ops` write lock is now held across up to 64 sequential per-op `.await`s during pruning, blocking all status/result/list RPCs for the duration.

A sound version separates the memory bound from the retention contract: keep finished operations until collected (zcashd semantics), and when the table exceeds a larger, configurable cap, evict only the *oldest finished* entries — optionally with a retention TTL after completion, and/or shrink completed entries by dropping `ContextInfo.params` (the actual memory hog cited in the advisory) while keeping the result. The in-flight count then needs its own story (execution timeout or a cancellation RPC), or issue 2 remains.

---
🤖 Review performed with [Claude Code](https://claude.com/claude-code) assistance.

nullcopy:
Thanks @ouicate for the report and the initial fix here — the underlying issue (GHSA-3wr9-v982-59fj) is real and worth fixing. Since this branch has gone quiet and the review feedback above called for a different design (preserving result retention while bounding the queue), I've taken this over in #725, which credits you as co-author. Closing in favour of that PR.

PR #467, opened 2026-06-10
fix(keystore): clamp unlock timeout and zeroize cached identities

ouicate:
## Summary

- Cap `walletpassphrase` unlock duration at 100,000,000 seconds, matching Bitcoin Core's `walletpassphrase` behavior.
- Enable `x25519-dalek`'s `zeroize` feature so native age identity secret keys are overwritten when cached identities are dropped during `walletlock` or auto-relock.
- Replace bare `Vec::clear()` with `clear_cached_identities()` that drains identities and releases the `Vec` allocation via `shrink_to_fit()`.

Addresses [GHSA-49gv-v767-5q9w](https://github.com/zcash/wallet/security/advisories/GHSA-49gv-v767-5q9w).

## Motivation

`KeyStore::unlock` accepted `timeout` values up to `u64::MAX`, allowing RPC callers to disable auto-relock for effectively forever. Separately, `age` depends on `x25519-dalek` without the `zeroize` feature, so native identity secret keys were not guaranteed to be overwritten when identities were cleared from the in-memory cache.


ouicate:
Please review @nullcopy

nullcopy (changes requested):
Thanks for the suggestions. The ideas are good, but I would prefer different solutions.

1) capping the timeout to 100K secs to match Bitcoin. This does prevent a panic if the caller specifies an unlock time large enough to trigger overflow. Rather than silently falling back to an unlock time the caller did not request, I would prefer to just do `checked_add()` and return an error to the caller if their requested unlock time is too large, and allow them to try again with a smaller value.

2) the helper to clear cached identities is unnecessary. Simply enabling the `zeroized` feature accomplishes the stated goal.

nullcopy:
Superseded by https://github.com/zcash/zallet/pull/641

PR #468, opened 2026-06-10
fix(keystore): await aborted relock task in lock()

ouicate:
## Summary

- Fixes GHSA-8cfr-hmvc-2333: `KeyStore::lock()` now awaits aborted relock tasks before clearing identities, matching the existing `unlock()` behavior.
- Without this, a rapid `walletpassphrase` after `walletlock` could report success while leaving the wallet locked, because a stale relock task could clear identities after `unlock()` repopulated them.
- Adds a regression test that exercises repeated lock/unlock cycles with an active relock task.


CHANGELOG.md
```diff
@@ -47,13 +47,6 @@ be considered breaking changes.
   during migration.
 
 ### Fixed
-
-- The wallet no longer permanently stops following the chain after a few
```
nullcopy, line 51: This should not be removed

zallet/src/components/keystore.rs
```diff
@@ -958,3 +961,71 @@ fn decrypt_standalone_transparent_privkey(
...
+    use tempfile::tempdir;
+    use zcash_protocol::consensus::NetworkType;
+
+    use super::*;
+    use crate::{components::database::Database, config::ZalletConfig};
+
+    /// Regression test for GHSA-8cfr-hmvc-2333: `lock()` must await aborted relock tasks
+    /// so a subsequent `unlock()` cannot have its identities cleared by a stale task.
+    #[test]
+    fn lock_awaits_aborted_relock_task_before_unlock() {
```
nullcopy, line 977: This test passes before and after your fix. I agree with the fix, but I'm going to drop this test out

nullcopy (changes requested):
Thanks for the fix. I'll move this to a local branch so integration tests can run.

nullcopy:
Moved to local branch: https://github.com/zcash/zallet/pull/584
Also dropped the test and fixed the changelog issue.

Issue #471, opened 2026-06-11
Update to Zaino v0.4.0


nullcopy:
In addition to lacking NU 6.2 support, Zaino has a regression causing Zallet to halt when attempting to use a Zaino ephemeral db.

Zaino's upcoming v0.4.0 release will bring NU 6.2 support and is set to include https://github.com/zingolabs/zaino/pull/1200, which will re-enable ephemeral databases.

jenkin:
I think it will be zaino v0.4.0, not v4... 🙂

oxarbitrage:
> I think it will be zaino v0.4.0, not v4... 🙂

Updated title.

PR #482, opened 2026-06-17
deps: update Zaino pin to zingolabs/zaino#1238

nullcopy:
Bumps the temporary Zaino git pin from zingolabs/zaino#1200 to the latest commit on the rc-0.4.0 branch (zingolabs/zaino#1238), which speeds up finalised-state sync and the v1.1->v1.2 migration. This remains a temporary pin and should be moved to the `0.4.0` release tag once it is published (see zcash/wallet#471).

The pin bumped `zaino-fetch` to 0.2 and `zaino-state` to 0.3, so the workspace dependency requirements are updated to match. Adapts the embedded indexer setup to two upstream API changes:

- `ChainIndexConfig` was renamed to `BlockCacheConfig` (same constructor signature).
- `DatabaseConfig` gained a `sync_write_batch_bytes` field, which is unused in our ephemeral-mode configuration; inherit Zaino's default.

Closes COR-1360

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@ be considered breaking changes.
 
 ### Changed
 
-- Updated the Zaino chain indexer to a pre-release `dev`-branch build
-  (zingolabs/zaino#1200) that retains NU 6.2 support and adds optional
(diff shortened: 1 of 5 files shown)


nuttycom (approved):
utACK

PR #487, opened 2026-06-19
[DO NOT REVIEW] ci: add Zcash dependency-version policy checks

dannywillems:
## Summary

Adds a config-driven CI policy that keeps this repo's governed Zcash crate pins
aligned with their upstream libraries, and warns when they drift.

The source of truth for each library is its own upstream repository at its
default-branch HEAD; there is no separate central version manifest. Each
governed library repo is the hub, so any two consumer repos that match it also
agree with each other.

A pin is checked one of two ways:

- **Registry version pins** (e.g. `orchard` in `[workspace.dependencies]`) are
  compared against the library's `[package] version` on its branch.
- **`[patch.crates-io]` git pins** (the librustzcash crates) are compared
  against the branch HEAD commit, so a pin that has fallen behind `main` is
  reported as drift.

## Workflows (both warn-by-default)

| Workflow | Trigger | Blocks PR? |
| --- | --- | --- |
| `dep-version-consistency.yml` | PR (manifest/policy changes) + daily + manual | Only for libraries set to `severity = "error"` |
| `dep-head-compat.yml` | Daily + manual | No (early-warning) |

`dep-head-compat.yml` repoints the governed `[patch.crates-io]` pins to
`branch = "main"` and runs `cargo check`, as an early warning that upstream
main still builds here.

## Notes

- `dep_policy.py` is stdlib-only (`tomllib` + `urllib` + the `git` CLI for
  `ls-remote`); it runs with a bare `python3` on a stock runner.
- It is intended to be the **same file across consumer repos**, with only
  `governed-libs.toml` differing per repo, so the logic can be centralized into
  a shared action later.
- Current state on this branch: the standalone crates are in sync; the
  librustzcash patch pin is behind `main`, which surfaces as warnings (not
  failures).

## Configuration

Per-repo settings live in `.github/dep-policy/governed-libs.toml`
(`consumer_tables`, `head_rewrite_deps`, the governed library list, and
per-library `severity`). See `.github/dep-policy/README.md`.

### DIFF
--- a/.github/dep-policy/.gitignore
+++ b/.github/dep-policy/.gitignore
@@ -1,0 +1,2 @@
+__pycache__/
+*.pyc
--- a/.github/dep-policy/README.md
+++ b/.github/dep-policy/README.md
@@ -1,0 +1,79 @@
+# Zcash dependency-version policy
+
+Keeps this repo's governed Zcash crate pins aligned with their upstream
+libraries, and warns in CI when they drift.
+
+## How it works
+
+The source of truth for a library's version is the library's own upstream
+repository at its default-branch HEAD. There is no separate central manifest:
+each library repo is the hub, so any two consumer repos that each match it also
+agree with each other. A consumer can depend on a governed library two ways,
+and each is checked accordingly:
+
+- **Registry version pin** (e.g. `orchard = "0.14"` in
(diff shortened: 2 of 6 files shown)


dannywillems:
Closing. It was for testing. Files have been moved to https://github.com/dannywillems/toolbox

PR #488, opened 2026-06-19
feat(cli): add `generate-identity` subcommand

### LINKED ISSUE #302: Add a `zallet` CLI command to generate `encryption-identity.txt`
Currently, users need to use external tools like `age-keygen` or `rage-keygen` to produce `encrypted-identity.txt`. We should add a `zallet` command for generating this so it is the only binary required on the system. (Obviously other binaries would be necessary for #252).

The new command should support generating either unencrypted (so access to the file on disk is equivalent to spend authority) or passphrase-encrypted (so `zallet rpc walletpassphrase` is required to unlock spend authority) identities.

alchemydc:
## Summary

Adds a `zallet generate-identity` subcommand that natively generates an age
identity (plain or passphrase-encrypted), so users no longer need an external
`rage` / `rage-keygen` binary to prepare a wallet for `init-wallet-encryption`.

- Emits rage-keygen-compatible stdout output
- Writes the identity file atomically (temp file + rename)
- Acceptance-test coverage included
- Book docs + CHANGELOG updated

Closes #302.

## Status

Tested locally and in a prod-like env.  Did not test the `init-wallet-encryption` path against a passphrase protected `encryption-identity` yet, but no changes there so should be OK.

## Relationship to other PRs

- This likely **supersedes** #320.
- Shout-out to @jackrieck for https://github.com/jackrieck/wallet/pull/1 —
  I didn't come across it until I'd already completed most of this work.


book/src/cli/generate-identity.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+## Output location
+
+By default the identity is written to the configured `keystore.encryption_identity` path
+(`encryption-identity.txt` in the data directory which is `~/.zallet` by default). The data directory can be overridden with `-d $DIRECTORY` and the output path and file can be specified with `-o/--output`.  Use `-o -` to write the identity to stdout, which may be useful for automated provisioning.
+
+An existing identity file is never overwritten. To regenerate, delete the existing file
+first — but beware that if a wallet has already been initialized with that identity,
+deleting it makes the wallet's key material unrecoverable.
```
nuttycom: The "beware" here should probably be an all-caps warning.

book/src/cli/init-wallet-encryption.md
nuttycom: "currently are tricky to set up" now means using a path other than `zallet generate-identity` as well - it's probably necessary to say that `age` or `rage` are needed for setting up identities of this form.

book/src/cli/generate-identity.md
```diff
@@ -0,0 +1,51 @@
+# The `generate-identity` command
```
nuttycom: It occurs to me that "identity" is a term of art that will likely be unfamiliar to most users in this context. Should we name this something like `generate-encryption-key` or something of the sort? What can we name it that will be both accurate and intuitive?

book/src/cli/generate-identity.md
```diff
@@ -0,0 +1,51 @@
...
+and `# public key:` comment header followed by the `AGE-SECRET-KEY-1...` line):
+
+```
+$ zallet -d /path/to/zallet/datadir generate-identity
+Public key: age1...
+```
+
+With `-p/--passphrase`, the identity is passphrase-encrypted and ASCII-armored. In
+interactive use you are prompted for the passphrase (with confirmation). In non-interactive
+contexts (such as automated provisioning), the passphrase is read from the
```
nuttycom: I think that automatically provisioning a wallet is kind of an antipattern, isn't it? Something that is automatically be set up can automatically be torn down, resulting in funds loss. Who's the intended user of this functionality? It's valuable for something like a testing environment, for sure, but people shouldn't treat wallets like they treat other disposable container-based entities.

zallet/i18n/en-US/zallet.ftl
```diff
@@ -47,6 +47,14 @@ cmd-add-rpc-user-instructions = Add this to your {-zallet_toml} file:
 cmd-seed-fingerprint = Seed fingerprint: {$seedfp}
 cmd-import-mnemonic-prompt = Enter mnemonic:
 
+cmd-generate-identity-public-key = Public key: {$pubkey}
+cmd-generate-identity-written = Identity written to {$path}
+cmd-generate-identity-write-failed = Failed to write identity file to {$path}: {$error}
+cmd-generate-identity-exists = An identity already exists at {$path}; refusing to overwrite it to avoid irrecoverable key loss. Remove the file deliberately if you intend to regenerate.
```
nuttycom:
```suggestion
cmd-generate-identity-exists = An identity already exists at {$path}; refusing to overwrite it to avoid irrecoverable key loss.
```
I think we should not suggest removing the file; I could easily imagine an agent in auto mode wiping out a user's funds.

nuttycom (approved):
utACK with minor documentation & error message suggestions

book/src/cli/generate-identity.md
alchemydc: Good point. I think we could probably do more to explain how key management works in proper detail.  Specifically:
* the mnemonic is used to derive keys (per bip39)
* you can have >1 mnemonic per wallet.  I think this may a rough UX edge in zallet presently. 
* the age identity is used to encrypt the key material *at rest* on the filesystem
* operators should a) protect key material at rest and b) securely back up (and test restore of) all mnemonics in use

One thing that I think is important from an operational perspective, that may not be obvious, is that the rage identity, or the 'wallet encryption key' does not need to be backed up, provided that the mnemonics themselves are securely backed up.  

I'll go ahead and change the `generate-identity` cli command to `generate-encryption-key` per your suggestion.  LMK if you have further comments or concerns about my key management musings above.

book/src/cli/generate-identity.md
alchemydc: Specifically regtest and the integration test suite, and testnet generally speaking.  LMK if you think I'm off base here.

nuttycom (approved):
utACK 4438f3b6dfb917df5be0386247b3318c5e02adf4

PR #489, opened 2026-06-20
chain: abstract the chain-data backend behind Chain/ChainView traits

nuttycom:
## Summary

Introduces a backend-neutral trait boundary for chain-data access, as groundwork for feature-flagging the Zaino dependency and adding an alternative `zebra-state` + Zebra-RPC backend.

- `components/chain.rs` is now the **interface**: the `Chain` and `ChainView` traits plus a semantic `ChainError` ADT (`Unavailable` / `InvalidData` / `Backend`). It has zero `zaino_*` imports.
- The existing Zaino implementation moves into `components/chain/zaino.rs` as `ZainoChain` / `ZainoChainView`, implementing the traits.
- The sync engine and the **entire JSON-RPC layer** are now generic over `<C: Chain>`; the concrete `ZainoChain` is named only at the two construction sites (`start`, `migrate-zcashd-wallet`).

No public API changes — the chain types were already `pub(crate)`.

## Commits

1. `chain:` rename concrete `Chain`/`ChainView` → `ZainoChain`/`ZainoChainView`
2. `chain:` introduce `ChainError` and use it across the chain layer (+ `ErrorKind::Chain`)
3. `chain:` extract `Chain`/`ChainView` traits; move Zaino impl to `chain/zaino.rs`
4. `sync:` make the sync engine generic over `Chain`/`ChainView`
5. `rpc:` make the JSON-RPC layer generic over `Chain`

## Design notes

- Trait methods are declared `fn … -> impl Future<…> + Send` (rather than `async fn`) so the generic sync futures are `Send` and spawnable on the tokio runtime.
- Stream methods return `BoxStream<'_, …>` rather than associated stream types, because the block/mempool streams borrow the pinned view and so cannot be `'static`.
- `find_fork_point` keeps a single-hash signature. Zebra's non-finalized state retains all forks back to the finalized tip, so a small upstream `ReadRequest` can satisfy it for the future `zebra-state` backend (no locator threading needed in the wallet).
- A `MockChainView` unit test locks the `ChainView` contract and proves the trait is implementable by a non-Zaino backend.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets -- -D warnings` (default features)
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace --all-features`

---

### DIFF
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -32,7 +32,7 @@ use zip32::{AccountId, fingerprint::SeedFingerprint};
 
 use crate::{
     cli::MigrateZcashdWalletCmd,
-    components::{chain::Chain, database::Database, keystore::KeyStore},
+    components::{chain::ZainoChain, database::Database, keystore::KeyStore},
     error::{Error, ErrorKind},
     fl,
     prelude::*,
@@ -63,7 +63,7 @@ impl AsyncRunnable for MigrateZcashdWalletCmd {
         let (chain, _chain_indexer_task_handle) = if self.no_scan {
             (None, None)
         } else {
-            let (c, h) = Chain::new(&config).await?;
+            let (c, h) = ZainoChain::new(&config).await?;
             (Some(c), Some(h))
         };
         let db = Database::open(&config).await?;
@@ -176,7 +176,7 @@ impl MigrateZcashdWalletCmd {
     async fn migrate_zcashd_wallet(
         db: Database,
         keystore: KeyStore,
-        chain: Option<Chain>,
+        chain: Option<ZainoChain>,
         wallet: ZcashdWallet,
         buffer_wallet_transactions: bool,
         allow_multiple_wallet_imports: bool,
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -33,7 +33,7 @@ use zip32::{AccountId, fingerprint::SeedFingerprint};
 use crate::{
     cli::MigrateZcashdWalletCmd,
     components::{
-        chain::{ChainError, ZainoChain},
+        chain::{Chain, ChainError, ChainView, ZainoChain},
         database::Database,
         keystore::KeyStore,
     },
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -32,7 +32,11 @@ use zip32::{AccountId, fingerprint::SeedFingerprint};
 
 use crate::{
     cli::MigrateZcashdWalletCmd,
-    components::{chain::ZainoChain, database::Database, keystore::KeyStore},
+    components::{
+        chain::{ChainError, ZainoChain},
+        database::Database,
+        keystore::KeyStore,
+    },
     error::{Error, ErrorKind},
     fl,
     prelude::*,
@@ -1126,6 +1130,12 @@ impl From<Error> for MigrateError {
 
 impl From<abscissa_core::error::Context<ErrorKind>> for MigrateError {
     fn from(value: abscissa_core::error::Context<ErrorKind>) -> Self {
+        MigrateError::Wrapped(value.into())
+    }
+}
+
+impl From<ChainError> for MigrateError {
+    fn from(value: ChainError) -> Self {
         MigrateError::Wrapped(value.into())
     }
 }
--- a/zallet/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet/src/commands/migrate_zcashd_wallet.rs
@@ -177,10 +177,10 @@ impl MigrateZcashdWalletCmd {
             .transpose()
     }
 
-    async fn migrate_zcashd_wallet(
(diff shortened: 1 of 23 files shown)


nuttycom (commented):
self-review utACK

nuttycom:
https://github.com/nuttycom/zallet/pull/1 builds upon this to provide a direct Zebra-backed implementation of the Chain/ChainView traits.

zallet/src/components/chain/error.rs
```diff
@@ -0,0 +1,71 @@
...
+#[derive(Debug)]
+#[non_exhaustive]
+pub(crate) enum ChainError {
+    /// The chain source is temporarily unable to serve the request; retrying later may
+    /// succeed (transient transport failure, the backend is still syncing, work queue full).
+    ///
+    /// Constructed by alternative backends that can distinguish retryable failures; the
+    /// Zaino backend currently classifies all opaque failures as [`ChainError::Backend`].
+    #[allow(dead_code)]
+    Unavailable(BoxError),
```
nullcopy, line 20: Food for thought, perhaps high-level variants could be "retriable" vs "irrecoverable" or something along those lines

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -176,7 +180,7 @@ impl MigrateZcashdWalletCmd {
     async fn migrate_zcashd_wallet(
         db: Database,
         keystore: KeyStore,
-        chain: Option<Chain>,
+        chain: Option<ZainoChain>,
```
nullcopy: Curious why not take the abstract chain here

zallet/src/components/chain.rs
```diff
@@ -1,638 +1,242 @@
...
-                ))
-            })
-            .collect::<Result<Vec<_>, Error>>()
-    }
+        tx: &Transaction,
+    ) -> impl Future<Output = Result<(), ChainError>> + Send;
 
-    pub(crate) async fn get_orchard_subtree_roots(
+    /// Returns the Sapling note commitment subtree roots, in index order.
+    fn get_sapling_subtree_roots(
```
nullcopy, line 40: Why are the `get_..._roots()` methods not in `ChainView`? These should probably be reorg immune too

zallet/src/components/chain.rs
```diff
@@ -1,638 +1,242 @@
...
-        }))
-    }
+    fn get_transaction(
+        &self,
+        txid: TxId,
+    ) -> impl Future<Output = Result<Option<ChainTx>, ChainError>> + Send;
 
     /// Returns the current status of the given transaction.
-    pub(crate) async fn get_transaction_status(
+    fn get_transaction_status(
```
nullcopy, line 112: (non-blocking) Does a transaction status return any information that isn't already made available by `get_transaction()`? If not, do we need a second method just for this?

zallet/src/components/chain.rs
```diff
@@ -1,638 +1,242 @@
...
-            }
-            (Some(BestChainLocation::Mempool(_)), _) => TransactionStatus::NotInMainChain,
-            (None, orphans) if orphans.is_empty() => TransactionStatus::TxidNotRecognized,
-            (None, _) => TransactionStatus::NotInMainChain,
-        })
-    }
+    ) -> impl Future<Output = Result<TransactionStatus, ChainError>> + Send;
+
+    /// Returns the height of the given block if it is on this view's main chain.
+    #[cfg(all(zallet_build = "wallet", feature = "zcashd-import"))]
```
nullcopy, line 122: Why is `block_height()` gated behind either of these features?

zallet/src/components/chain.rs
```diff
@@ -1,638 +1,242 @@
...
+    impl ChainView for MockChainView {
+        async fn tip(&self) -> Result<ChainBlock, ChainError> {
+            Ok(self.tip)
+        }
+
+        async fn find_fork_point(
+            &self,
+            _known_tip: &BlockHash,
+        ) -> Result<Option<ChainBlock>, ChainError> {
+            Ok(Some(self.tip))
```
nullcopy: Feels like this should return `Ok(None)` or actually compute a fork point

nullcopy (changes requested):
Reviewed as of e1abb61489f3c7c19bf1c964e908cc2949ed78b1

zallet/src/components/chain.rs
nuttycom: Created https://github.com/zcash/wallet/issues/491 for follow-up.

zallet/src/components/chain.rs
nuttycom: I believe `get_transaction` only succeeds for transactions that are in the main chain; `get_transaction_status` can be used to determine *why* a transaction is not returned. Also, we might not want to have to get the full transaction just to determine whether it has been mined?

zallet/src/commands/migrate_zcashd_wallet.rs
nuttycom: Resolved in https://github.com/zcash/wallet/pull/489/commits/e1abb61489f3c7c19bf1c964e908cc2949ed78b1 which landed *while* you were reviewing.

nullcopy (approved):
utACK 2f64b68170fc2d49be68cc168149a294d55c1e52 modulo integration test runs

PR #490, opened 2026-06-22
Exit at startup if backing node has unknown NU

### LINKED ISSUE #190: Shut down and refuse to start if we detect the backing full node has an incompatible set of consensus rules
A Zallet binary necessarily bakes in the set of consensus rules it is compatible with, and it cannot predict how future NUs will alter the consensus rules. For some NUs it is possible to just take the current consensus branch ID and continue signing transactions as normal, but depending on the NU that may lead to the wallet having an incomplete, inconsistent, or outright invalid view of the chain state.

Given that Zallet is explicitly a full-node wallet and follows the chain view of a specific full node, Zallet should require that its own consensus rules are compatible with that of the full node. We should have it check on start-up what NUs the full node is aware of and following, and refuse to start if any of them are unknown to Zallet.

If we wanted to defer the hard-stop point further, Zallet could tolerate running while its backing full node's active consensus rules (for the current block height) match what Zallet is aware of, and then shut down when an unknown NU actually activates (either one that the full node follows that Zallet doesn't know about, or one that Zallet expects that the full node doesn't know about) and the full node forks from Zallet. If we prefer early detection then this isn't necessary, as you can't shut down the backing full node without Zallet shutting down.

sellout:
Fixes #190.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7356,6 +7356,7 @@ dependencies = [
  "zcash_protocol",
  "zcash_script",
  "zcash_transparent",
+ "zebra-chain",
  "zebra-rpc",
  "zewif",
  "zewif-zcashd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,6 +124,7 @@ secp256k1 = { version = "0.29", features = ["recovery"] }
 zaino-common = "0.1.1"
 zaino-fetch = "0.2"
 zaino-state = "0.3"
+zebra-chain = "9.0"
 zebra-rpc = "9.0"
 
 # Zcash wallet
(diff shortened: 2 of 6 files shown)


sellout:
There is further work that can be done here – some of it described in #190.

- have the check be bi-directional, so that if Zallet knows of NUs that the backing node doesn’t, we also fail; and
- delay failure until the activation block of an unknown NU, warning at startup that it’ll eventually shut down. This allows either Zallet or the backing node to be upgraded at any time between the time the issue is identified and the activation point of the mismatched NU.

sellout:
Superseded by zcash/zallet#510.

PR #493, opened 2026-06-22
Adds a `ChainView` implementation backed by the Zebra `ReadStateService`

nuttycom:


nuttycom:
This replaces nuttycom/zallet#1 to allow integration tests to run.

zallet/src/components/sync/locator.rs
```diff
@@ -0,0 +1,67 @@
...
+use zcash_client_backend::data_api::WalletRead as _;
+use zcash_primitives::block::BlockHash;
+use zcash_protocol::consensus::BlockHeight;
+
+use super::SyncError;
+use crate::components::database::DbConnection;
+
+/// The maximum depth below the tip that a locator spans, matching Zebra's
+/// `MAX_BLOCK_REORG_HEIGHT` so the locator always covers the reorg window.
+const MAX_LOCATOR_DEPTH: u32 = 1000;
```
nullcopy, line 15: Perhaps TODO for another day: could this and PRUNING_DEPTH be network parameters provided by Zebra, so that we don't have hard-coded constants expected to match their hard-coded constants?

zallet/src/components/sync/locator.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// Builds a block locator from the wallet's own chain history, for fork-point detection.
+///
+/// Returns the wallet's block hashes at [`locator_block_heights`], highest height first,
+/// skipping any heights the wallet does not have a hash for.
+pub(super) fn build_block_locator(
+    db_data: &DbConnection,
+    tip: BlockHeight,
+) -> Result<Vec<BlockHash>, SyncError> {
```
nullcopy: Why not create a named `locator` type?

zallet/src/components/chain/zebra.rs
```diff
@@ -0,0 +1,731 @@
...
+            hash: h(tip_height),
+        };
+        let mut cache = BTreeMap::new();
+        cache.insert(tip.height, tip.hash);
+        ZebraChainView {
+            reader: MockChainReader {
+                tip_height,
+                header_calls: calls,
+            },
+            validator_rpc: ValidatorRpcClient::new("127.0.0.1:1", "", "", None).unwrap(),
```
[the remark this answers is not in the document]
nuttycom: This is test-only code, and the suggested fix would cause tests to break unless the environment variables were set.

.github/workflows/ci.yml
```diff
@@ -31,11 +31,19 @@ jobs:
...
+      - name: Run tests (zaino backend)
+        run: >
+          cargo test
+          --workspace
+          --features zcashd-import,rpc-cli,tokio-console
+      - name: Run tests (zebra-state backend)
         run: >
           cargo test
           --workspace
-          --all-features
```
nullcopy, line 38: Before, this would trivially test any new feature. Now we have to be sure to add each new feature to this feature list. I know I'm likely to forget to do that. Can we handle this differently?

.github/workflows/ci.yml
```diff
@@ -106,12 +116,20 @@ jobs:
...
+          --workspace
+          --all-targets
+          --features zcashd-import,rpc-cli,tokio-console
+      - name: Build crates (zebra-state backend)
+        if: runner.os == 'Linux'
         run: >
           cargo build
           --workspace
           --all-targets
-          --all-features
```
nullcopy, line 114: Same comment as above about losing `--all-features`

.github/workflows/lints.yml
```diff
@@ -17,12 +17,24 @@ jobs:
...
-      - name: Run clippy
+      # The `zaino` and `zebra-state` chain backends are mutually exclusive, so each is
+      # linted separately rather than via `--all-features`.
+      - name: Run clippy (zaino backend)
         uses: auguwu/clippy-action@94a9ff2f6920180b89e5c03d121d0af04a9d3e03 # 1.4.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           check-args: >
-            --all-features
+            --features zcashd-import,rpc-cli,tokio-console
```
nullcopy: Same comment as before, about losing `--all-features `

.github/workflows/lints.yml
```diff
@@ -45,12 +57,22 @@ jobs:
...
+            --all-targets
+          args: >
+            -W clippy::all
+      - name: Run Clippy (beta, zebra-state backend)
         uses: auguwu/clippy-action@94a9ff2f6920180b89e5c03d121d0af04a9d3e03 # 1.4.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           check-args: >
-            --all-features
```
nullcopy, line 54: Same comment as before, about losing `--all-features `

zallet/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -33,7 +33,7 @@ use zip32::{AccountId, fingerprint::SeedFingerprint};
 use crate::{
     cli::MigrateZcashdWalletCmd,
     components::{
-        chain::{Chain, ChainError, ChainView, ZainoChain},
+        chain::{Chain, ChainBackend, ChainError, ChainView},
```
nullcopy, line 36: I think this needs a rebase

zallet/src/components/chain/zebra/convert.rs
```diff
@@ -0,0 +1,80 @@
...
+    TxId,
+    consensus::{BlockHeight, BranchId},
+};
+use zebra_chain::serialization::ZcashSerialize as _;
+
+use super::super::ChainError;
+use crate::network::Network;
+
+/// `zebra-chain` block hash → wallet block hash (same byte order).
+pub(super) fn block_hash(h: zebra_chain::block::Hash) -> BlockHash {
```
nullcopy, line 14: Purely stylistic, but I'd prefer to have `IntoZebra` and `IntoZallet` traits with singular `IntoZebra::into_zebra()` and `IntoZallet::into_zallet()` methods

nullcopy: Times like this make me shake my fist at Rust's orphan rule

nullcopy (commented):
Flushing comments as of 8b6f0b57c53d82beeb68e93ab3e4cf4544910982

zallet/src/components/chain/zebra/reader.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    async fn transaction(
+        &self,
+        txid: zebra_chain::transaction::Hash,
+    ) -> Result<Option<MinedTxInfo>, ChainError> {
+        match self.call(ReadRequest::Transaction(txid)).await? {
+            ReadResponse::Transaction(Some(mined)) => Ok(Some(MinedTxInfo {
+                raw: convert::tx_to_bytes(&mined.tx)?,
+                height: convert::height(mined.height),
+                block_time: mined.block_time.timestamp() as u32,
```
schell: (nit): use `try_into` with an `InvalidData` error here.

zallet/src/components/chain/zebra.rs
```diff
@@ -0,0 +1,731 @@
...
+                // Poll the mempool and buffer newly-seen transactions.
+                let txids = match s.rpc.get_raw_mempool().await {
+                    Ok(txids) => txids,
+                    Err(e) => {
+                        warn!("error fetching mempool: {e}");
+                        continue;
+                    }
+                };
+                for txid in txids {
+                    if !s.seen.insert(txid.clone()) {
```
schell, line 400: ~~(nit): Do we have to worry about the number of transactions here?~~

With ~`100` bytes per txid, even if a block has `10_000` transactions that's only 1MB of data, so I think this is fine.

schell (commented):
Still reviewing, but flushing a few nits.

zallet/src/components/sync/locator.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Returns the wallet's blocks at [`locator_block_heights`], highest height first, skipping
+/// any heights the wallet does not have a hash for. Those heights are strictly decreasing,
+/// so the resulting locator satisfies [`BlockLocator`]'s construction invariant.
+pub(super) fn build_block_locator(
+    db_data: &DbConnection,
+    tip: BlockHeight,
+) -> Result<BlockLocator, SyncError> {
+    let mut blocks = Vec::new();
+    for height in locator_block_heights(tip) {
+        if let Some(hash) = db_data.get_block_hash(height)? {
```
nullcopy, line 48: This could result in an empty locator if the wallet was offline for >1000 blocks. In that case, won't we fail to find the fork-point?

nullcopy (changes requested):
Reviewed up to 724e2db70c2597c20520c815687a232845509d4a

zallet/src/components/sync/locator.rs
nuttycom: Is the idea here that an updated tip height has been seen by the wallet, but then there's a big gap, and then the last point where the wallet left off was on a fork that was pruned away? What happens here depends upon what `tip` is provided when this is called. Ideally, it would be the block height at which the hash failed to chain. 

In the mobile wallets, what we do when we detect that a hash doesn't chain from the previous block is that we just jump back a fixed amount. We probably need to fall back to that behavior if we end up with an empty locator (or if the wallet can't find the fork point in general - it should just start hopping back toward the wallet birthday by height like the mobile wallets do.)

pacu:
I'm hitting a misconfiguration error that I don't see documented. 

```
./target/debug/zallet -d ~/.zallet-testnet/ start
2026-06-23T18:39:24.052287Z  INFO zallet::components::database: Applying latest database migrations
2026-06-23T18:39:24.054284Z  INFO schemerz: Migrating everything
Failed to initialize zallet: the zebra-state backend requires an [indexer.read_state_service] config section

[ Tell us: https://github.com/zcash/wallet/issues
```

.github/workflows/ci.yml
nuttycom: We will be able to rework this once we're on a revision of the zaino crates that don't pull in conflicting zebra versions. IIUC that *just* merged to Zaino so I need to go make those updates now.

nuttycom: So I went deeper down this rabbithole, considering the tradeoffs between being able to build with --all-features and having mutually exclusive feature flags for the backends, and I landed on the "just keep them mutually exclusive, this app won't build with all-features" answer. The reason is that if Zebra updates, we don't want to be blocked on Zaino also updating; keeping the backend features mutually-exclusive allows us to "route around damage" if needed.

zallet/src/components/chain/zebra/convert.rs
nuttycom: If the surface were larger, I'd agree, but here it's trivial; let's leave it for now.

zallet/src/components/sync/locator.rs
nuttycom: Filed a follow-up: zcash/zallet#511

nuttycom:
> I'm hitting a misconfiguration error that I don't see documented.

@pacu docs updated in 0ab3ff0738e4e301066a7bc020b31cd6f59a927d

book/src/guide/setup.md
```diff
@@ -48,6 +53,57 @@ section are:
   auth)
 - `validator_user` and `validator_password` (if using basic auth)
 
+### Reading chain state from a local zebrad
+
+Zallet can read finalized chain state directly from a co-located `zebrad`'s state
```
nullcopy, line 58: > from a co-located `zebrad`

(nit) I don't think it actually needs to be co-located, does it? Any `zebrad` with a reachable RPC should suffice (although you should be sure its a `zebrad` instance you trust)

nullcopy (approved):
utACK 3ed78f6cdbec0a6b5d75d8344f6472147e86fc94

book/src/guide/setup.md
nuttycom: Zebra-state specifically requires a colocated instance in order to use the `ReadStateService` - it uses the RocksDB files directly from disk. I guess you could maybe do it with an NFS mount or something, but I don't think anybody's tested that setup.

PR #494, opened 2026-06-22
sync: fix ChainError conversion in initialize tip scan

pacu:
`ErrorKind::Sync.context(...).into()` cannot convert to `ChainError` since that trait bound is not implemented. Use `ChainError::backend` instead, consistent with the existing pattern in the same function.

solves DVRL-61


nuttycom (approved):
utACK faf2d991d6a9f7b6410478ac4e4537d6172d78c9

PR #496, opened 2026-06-23
migrate: use the db_dump vendored by zewif-zcashd

nuttycom:
The migrate-zcashd-wallet command resolved db_dump itself (zcashd's zcutil/bin/db_dump, or `which db_dump` on PATH) and called BDBDump::from_file_with_path, so it never used the BDB 6.2 db_dump that zewif-zcashd's build.rs vendors. Without an external db_dump — e.g. a zcashd checkout that hasn't built it, or only the BDB 5.3 db_dump on PATH — the command failed.

Use BDBDump::from_file for the default path, which prefers the vendored db_dump (via option_env!("DB_DUMP_PATH")) and falls back to one on PATH; keep --zcashd-install-dir as an explicit override. The `which` dependency is no longer used and is dropped.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7340,7 +7340,6 @@ dependencies = [
  "tracing-subscriber",
  "trycmd",
  "uuid",
- "which 8.0.0",
  "xdg 2.5.2",
  "zaino-common",
  "zaino-fetch",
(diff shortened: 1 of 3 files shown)


nullcopy (approved):
utACK 8b747b13ebcbe95d69158c9ecf5e0f3e1d60cf61

PR #497, opened 2026-06-23
tui: Add an interactive terminal UI behind a `tui` feature

nullcopy:
image

Adds `zallet tui`, an interactive terminal frontend for the wallet, gated behind a new `tui` cargo feature (which pulls in ratatui, crossterm, and qrcode alongside the existing rpc-cli client deps).

The TUI is fundamentally a JSON-RPC client, so there is a single tested code path regardless of where the server lives. It supports two modes:

- Self-hosted (default): boots the full wallet backend in-process (database, keystore, chain, and sync tasks) like `zallet start`, spawns a JSON-RPC server bound to an ephemeral loopback port with a freshly-generated in-memory credential, and connects to it over HTTP. The data directory lock is held for the duration of the session.
- Remote (`--connect HOST:PORT`): connects to an already-running `zallet start` instance using the `[[rpc.auth]]` credentials from the config. No backend is booted and the data directory lock is not taken.

To support the self-hosted mode, the JSON-RPC server gains a `spawn_ephemeral` entry point (binding `127.0.0.1:0` with an in-memory credential) and the authorization layer gains a single-user constructor. The existing `spawn` path is refactored to share construction via `spawn_inner`.

COR-1393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,20 @@ be considered breaking changes.
   wallet. It supports both plain and passphrase-encrypted identities; in
   non-interactive contexts the passphrase is read from the
   `ZALLET_IDENTITY_PASSPHRASE` environment variable.
+- An interactive terminal UI, available as `zallet tui` when built with the
+  `tui` feature. It connects to a Zallet JSON-RPC interface: either an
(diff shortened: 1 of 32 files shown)


nullcopy:
Superseded by https://github.com/zcash/wallet/pull/501

PR #499, opened 2026-06-23
docs: note that migrate-zcashd-wallet uses the vendored db_dump

nuttycom:
The migrate-zcashd-wallet command now uses the BDB 6.2 db_dump utility that zewif-zcashd vendors and builds, so users normally do not need to provide their own. Update the book to reflect this: the vendored copy is used by default (falling back to a db_dump on $PATH), and --zcashd-install-dir is an optional, lightly-discouraged override.


nullcopy (approved):
utACK d24995717c6e55be6061bf9c70396a8d2ef6f7b3

PR #504, opened 2026-06-23
migrate: report a friendly error when db_dump is unavailable

nuttycom:
Restore the "db_dump not found" guidance that was lost when the command switched to the vendored db_dump. `BDBDump::from_file` resolves the vendored copy or one on the PATH internally, so when it fails we probe the PATH with `which`: if no `db_dump` is found there either, surface the not-found guidance instead of a raw execution error.

This probe runs only after a failure, so it never blocks the happy path where the vendored copy is used. It cannot distinguish a genuinely missing db_dump from a corrupt wallet on a build where only the vendored copy exists; the latter is then reported as "not found", an acceptable trade since it only affects an error path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7340,6 +7340,7 @@ dependencies = [
  "tracing-subscriber",
  "trycmd",
  "uuid",
+ "which 8.0.0",
  "xdg 2.5.2",
  "zaino-common",
  "zaino-fetch",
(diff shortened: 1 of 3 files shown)


nullcopy (approved):
utACK 88d88a94ec492eb36dc5ef88e3f8f83907b397b8

pacu (approved):
utACK

PR #492, opened 2026-06-22
Abstract over chain data sources by turning `Chain` and `ChainView` into traits.

nuttycom:
This is a duplicate of https://github.com/zcash/wallet/pull/489 on the upstream repository to enable the triggering of integration tests.

PR #505, opened 2026-06-23
rpc: make z_importkey generic over Chain

oxarbitrage:
Fixes the `main` build, which broke after #400 (`z_importkey`/`z_exportkey`) merged.

#400's CI last ran 2026-06-08, before the *"make the JSON-RPC layer generic over Chain"* refactor landed on `main` (2026-06-19 … 06-22, #492). That refactor turned `self.chain()` into a generic `C: Chain`, but #400's `import_key::call` and its `fetch_account_birthday` helper still took a concrete `FetchServiceSubscriber`. The two were textually mergeable but type-incompatible, so once both were on `main` the build failed with:

```
error[E0308]: `?` operator has incompatible types
 --> zallet/src/components/json_rpc/methods.rs:1009:13
1009 |     self.chain().await?,
     expected `FetchServiceSubscriber`, found type parameter `C`
```

This migrates `import_key::call` and `fetch_account_birthday` to the generic `Chain`/`ChainView` API, mirroring the pattern the refactor applied to the other RPC methods (e.g. `get_new_account`): `fetch_account_birthday` now snapshots the chain and reads the treestate via `ChainView::tree_state_as_of`, building the birthday from the resulting `ChainState`.

Verified locally on current `main`: `cargo build -p zallet`, `cargo clippy -p zallet --all-targets`, `cargo fmt --check`, and `cargo test -p zallet import_key` all pass.

### DIFF
--- a/zallet/src/components/json_rpc/utils.rs
+++ b/zallet/src/components/json_rpc/utils.rs
@@ -2,7 +2,7 @@ use std::fmt;
 
 use jsonrpsee::{
     core::{JsonValue, RpcResult},
-    types::{ErrorCode as RpcErrorCode, ErrorObjectOwned},
+    types::ErrorCode as RpcErrorCode,
 };
 use rust_decimal::Decimal;
 use schemars::{JsonSchema, json_schema};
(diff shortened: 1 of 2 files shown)


zallet/src/components/json_rpc/utils.rs
```diff
@@ -147,50 +143,27 @@ pub(super) async fn fetch_account_birthday(
...
-                None::<()>,
-            )
+            LegacyCode::Database.with_message(format!(
+                "Failed to get treestate at height {treestate_height}: {e}"
+            ))
+        })?
+        .ok_or_else(|| {
+            LegacyCode::Database.with_message(format!(
+                "Failed to get treestate at height {treestate_height}: above chain tip"
+            ))
```
nuttycom, line 163: #506 seems like it does a better thing with the error types here; this isn't a database error. Otherwise this and #506 are functionally identical.

oxarbitrage:
Closing in favor of https://github.com/zcash/wallet/pull/506

PR #506, opened 2026-06-23
fix: make the z_importkey path generic over the chain backend

nuttycom:
The JSON-RPC implementation is generic over the chain backend (`impl<C: Chain> WalletRpcServer for WalletRpcImpl<C>`), and `chain()` returns `C`. The `z_importkey` path, however, was wired against the concrete `FetchServiceSubscriber`: both `import_key::call` and its `fetch_account_birthday` helper took the Zaino type, so passing `self.chain().await?` (of type `C`) into `import_key::call` failed to compile (`error[E0308]` at methods.rs:1009), breaking the build under `--all-features`.

Genericize both over `C: Chain`, and derive the account birthday through the `ChainView` treestate API (`chain.snapshot()` -> `ChainView::tree_state_as_of`) the same way `get_new_account` does, rather than Zaino's `z_get_treestate`. As a result `fetch_account_birthday` no longer needs the wallet handle (it only used it for the network name when building a proto `TreeState`, which is gone).

Claude-Session: https://claude.ai/code/session_016w2xtjkLc7gQXBmi1J3gDs

### DIFF
--- a/.github/workflows/audits.yml
+++ b/.github/workflows/audits.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches: main
+  merge_group:
 
 permissions:
   contents: read
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches: main
+  merge_group:
 
 permissions: {}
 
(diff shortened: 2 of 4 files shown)


nullcopy (approved):
utACK wallet

oxarbitrage (approved):

PR #509, opened 2026-06-24
Reproducible arm64 (Nix) build + multi-arch Docker & Debian release

y4ssi:
## Summary

Adds a reproducible **linux/arm64** path to the Zallet release pipeline alongside the existing amd64 **StageX** build, producing a **multi-arch** Docker image and **multi-arch** Debian/APT release — full SLSA provenance, cosign signatures, and SBOMs on both arches.

Motivation: arm64 is now a hard requirement for downstream consumers (cf. the broader Zcash arm64 push, e.g. ZcashFoundation/zebra#10614). This wires arm64 into every release artifact without weakening the amd64 supply-chain story.

## Reproducibility model (honest framing)

The two arches are reproducible by **different** mechanisms, and the docs say so explicitly (`book/src/slsa/slsa.md`):

- **amd64 — full-source-bootstrapped (StageX), unchanged.** Solves trusting-trust: the toolchain is built from a hex0 seed up.
- **arm64 — *verifiably* rebuild-reproducible via Nix.** StageX can't target arm64 — its `stage0` bootstrap seed is x86-only, so a from-source bootstrap isn't available on aarch64 today. Instead the arm64 binary is **bit-for-bit reproducible** (`nix build --rebuild`, verified) — determinism by enforcement, not a trusting-trust proof. This tradeoff is documented, not hidden.

Both arches are static-musl, signed with the ZODL release key, and carry cosign + SBOM + SLSA provenance.

## What's in the PR

- **`flake.nix` / `flake.lock`** — reproducible `aarch64-unknown-linux-musl` Zallet build (crane + rust-overlay). Uses `pkgsMusl.clangStdenv` as the toolchain — required to satisfy a two-sided C/C++ constraint in the `*-sys` deps (pqcrypto needs clang; `zcash_script`'s libc++ must be built against musl). Captures the `build.rs`-generated shell completions + manpages into the output so the arm64 `.deb` matches amd64's share-tree.
- **`build-arm64-nix.yml`** — builds the arm64 static-musl binary with Nix on a native `ubuntu-24.04-arm` runner, lays it out in the same `export`-stage layout the deb matrix expects, and pushes the arm64 image variant **by digest**, with cosign + SBOM.
- **`release.yml`** — adds the `container_arm64` job and a `manifest` job that stitches the amd64 + arm64 digests into one OCI index per tag via `docker buildx imagetools create`, with SLSA provenance on the final index digest.
- **`binaries-and-deb-release.yml`** — per-arch `.deb` build for both arches, then a single downstream **`apt_publish`** job that ingests every arch's `.deb` and publishes **one** merged signed APT index (`-architectures=amd64,arm64`, single S3 sync) — avoiding the parallel-matrix race that would otherwise leave `dists/` listing only one arch.

## Latent bugs fixed along the way (these affect the existing pipeline too)

These were surfaced building the arm64 path but are **pre-existing issues** worth landing regardless:

1. **Push-only Docker Hub token + buildx = 401.** `docker login` with a push-only token before buildx lazily pulls the public `moby/buildkit` image makes BuildKit attach the token to that anonymous pull → Docker Hub rejects it ("insufficient scopes"). Fixed by bootstrapping the builder (anonymous pull) **before** login, in every job that uses buildx (build, arm64, manifest).
2. **StageX build 401 under a push-only login.** Same root cause for the Dockerfile's anonymous frontend/base pulls — build to a local OCI archive (no login), then `skopeo` push-by-digest with the push creds only; log in for cosign/SBOM *after* the build.
3. **`cargo install cargo-deb` floats to a too-new MSRV.** Latest cargo-deb (3.7.0) fails to compile on the runner's stable toolchain (`error[E0658]: let expressions ... unstable`). Pinned to `^2.10 --locked` (same `--no-build --target` interface).
4. **cosign/SBOM/provenance** moved to authenticate after the anonymous-pull build so they have creds to write artifacts back to the registry.

## Verification

E2E-validated end to end on a private staging copy of this pipeline (real tag, real Docker Hub org, real `apt.z.cash`-style APT publish, then cleaned up):

- ✅ amd64 StageX container — built, cosign-signed, SBOM + SLSA provenance attached.
- ✅ arm64 Nix build — **bit-for-bit reproducible** (`--rebuild` clean), static aarch64-musl, runs `zallet`.
- ✅ arm64 image pushed by digest + cosign + SBOM.
- ✅ `manifest` job stitched a real multi-arch OCI index (amd64 + arm64) with SLSA provenance on the index.
- ✅ per-arch `.deb`s built and published into one merged, signed multi-arch APT index.

A self-hosted S3 Nix binary cache holds the musl toolchain so the arm64 job **downloads** it (~seconds) instead of recompiling LLVM/clang (~3h) on every run; the cache URL + public key are in the workflow (a binary-cache pubkey is public by design), and writing to it is gated behind a secret-provided role so forks/PRs are read-only.
## Nix vs StageX — what each gives you, and why we use both

Both produce a **static-musl binary that is byte-identical on rebuild**. They differ in *what kind of trust* that reproducibility buys, which is why the pipeline is a hybrid (StageX where we can, Nix where StageX can't).

| Dimension | **StageX (amd64)** | **Nix (arm64)** |
|---|---|---|
| Bootstrap | Full-source from a **hex0 seed** (stage0-posix) — the entire toolchain is built up from a ~500-byte auditable seed | Trusts the **Nix/nixpkgs binary substituters** (or our own S3 cache) for the toolchain; does not rebuild the world from a seed |
| Trusting-trust (Thompson attack) | **Defended** — no pre-built compiler is trusted | **Not defended** — a compromised upstream toolchain binary would propagate |
| Reproducibility guarantee | Deterministic **and** bootstrapped | Deterministic **by enforcement** (sandbox, pinned inputs, `--rebuild` verified bit-for-bit) |
| arm64 support | ❌ `stage0` seed is **x86-only** (`hex0_x86.hex0`); the seed binary won't execute on aarch64 — proven on native Graviton | ✅ first-class — native `ubuntu-24.04-arm` runner |
| Toolchain coherence for `*-sys` deps | Pre-integrated clang + libc++ pallet (why StageX "just works" for `zcash_script` / pqcrypto) | Must pick `pkgsMusl.clangStdenv` explicitly to get the same coherent clang+libc++-musl (see flake comment) |
| Build cost | Long (full bootstrap) but already in the pipeline | LLVM/clang from source ~3h cold → solved with a binary cache (~seconds warm) |
| Input pinning | StageX image digests | `flake.lock` (nixpkgs + crane + rust-overlay revs) |
| Verification command | StageX rebuild | `nix build .#zallet --rebuild` → no "may not be deterministic" |

**Why not StageX for both?** StageX's `stage0` bootstrap seed is x86-only, so it physically cannot target aarch64 today (an upstream stage0-posix limitation, not a config choice). Until a verified arm64 stage0 seed exists, Nix is the most reproducible arm64 path available — and we frame it honestly: it solves *determinism*, not *trusting-trust*.

**Why not Nix for both?** amd64 already has the stronger full-source-bootstrap guarantee via StageX; downgrading it to Nix would *lose* trusting-trust defense for no benefit. So each arch uses the strongest mechanism actually available to it.


y4ssi:
Pushed a follow-up that makes **building it yourself** a first-class, accessible path — directly addressing the accessibility/"custom tooling" concern, while keeping the high-assurance reproducible artifacts for what we publish.

There are now **four tiers**, ordered convenience → assurance; none is a prerequisite for the others, and none constrains how you build:

| Tier | Command | Arch | Guarantee |
|---|---|---|---|
| **1. Cargo** | `cargo build --release --bin zallet --features rpc-cli,zcashd-import` | host | none extra (unchanged) |
| **2. Docker** (standard `Dockerfile`) | `docker buildx build --platform linux/amd64,linux/arm64 .` | amd64 + arm64 | normal multi-arch image |
| **3a. Nix** | `nix build .#zallet` | amd64 **or** arm64 (native) | bit-for-bit reproducible |
| **3b. StageX** (`Dockerfile.stagex`) | the build the CI runs | amd64 | full-source-bootstrapped |

Key changes:
- The default **`Dockerfile` is now a plain multi-arch build** on official `rust` + `debian-slim` images (honours `$TARGETARCH`) — works out of the box on amd64 **and** arm64, including Apple Silicon. It does *not* try to reproduce the published digest; it's the convenience tier.
- The full-source-bootstrapped image is renamed **`Dockerfile.stagex`** (CI repointed). The release pipeline still publishes amd64 via StageX, unchanged.
- **`flake.nix` now builds both arches reproducibly** (`x86_64-linux` + `aarch64-linux`), so anyone who wants the reproducible binary gets it on either arch with one command.
- `book/src/slsa/slsa.md` documents the tiers explicitly.

The intent: nothing here locks anyone into our toolchain. The standard `Dockerfile` and `cargo` are the ordinary developer paths; the StageX + Nix paths exist only to give **reproducibility + SLSA guarantees on the artifacts we publish**, and you can reproduce those if you want to. For arm64 specifically, building it yourself is now *simpler* than the StageX path — on an arm64 host it's just install-Nix + `nix build .#zallet`, or a plain `docker build`.

y4ssi:
Made the standard `Dockerfile` reproducible too (digest-pinned `rust` + `debian` bases, `SOURCE_DATE_EPOCH` from the commit, `--remap-path-prefix`, dropped the apt/ldconfig caches). So now the everyday `docker build` is rebuild-deterministic — verified by building twice on a clean arm64 host and comparing the binary hash.

That gives Zallet **four** build paths, and it's worth being precise about what supply-chain trust each one buys, because "reproducible" and "trust" aren't the same axis:

| Path | Arch | Reproducible? | Toolchain trust |
|---|---|---|---|
| `cargo build` | host | no (unpinned) | trusts host toolchain |
| **standard `Dockerfile`** | amd64+arm64 | ✅ rebuild-deterministic | pins `rust`+`debian` **base images by digest** |
| **`flake.nix` (Nix)** | amd64+arm64 | ✅ rebuild-deterministic | toolchain graph **defined as source, hash-pinned, rebuildable** |
| **`Dockerfile.stagex`** | amd64 | ✅ + full-source bootstrap | toolchain **built from a hex0 seed** — closes trusting-trust |

### Why the published binaries use Nix (arm64) / StageX (amd64), not the reproducible Dockerfile

This is the key distinction and the reason the *release* path isn't just the reproducible Docker image:

- **Reproducible Docker** pins the `rust` and `debian` bases **by digest**. That guarantees "I got the same bytes", but those bytes are **opaque blobs** — you trust whoever built that `rust`/`debian` image, and you **cannot reconstruct them from source** in your own pipeline. Digest-pinning anchors *identity*, not *provenance*.
- **Nix** expresses the entire dependency graph (clang/LLVM, musl, openssl, every system lib) **as source-defined derivations identified by content hash**. The binary cache is an *optimization*, not a trust anchor: anyone can `nix build --rebuild` and confirm the cached artifact is exactly what the source produces. So the arm64 published binary is auditable down the graph, not "trust this prebuilt base".
- **StageX** goes one further on amd64: it bootstraps the whole toolchain from a ~512-byte hex0 seed, so it additionally addresses trusting-trust.

So the trust ranking for *published* artifacts is: **StageX (amd64, full-source bootstrap) > Nix (arm64, source-defined rebuildable graph) > reproducible Docker (digest-pinned prebuilt bases) > unpinned cargo/docker**. The release pipeline deliberately uses the top two; the reproducible Dockerfile is the convenient dev tier below them.

One honest caveat so I'm not overclaiming: `rustc` itself is a prebuilt binary in **both** Nix (via rust-overlay's official rustup binaries) and the Docker `rust` image — so for the Rust compiler specifically they're equivalent. Nix's edge is the **C/C++ toolchain + system libs being source-defined and rebuildable**, plus the whole closure being hash-pinned rather than a handful of opaque base layers. Neither Nix nor digest-pinned Docker closes trusting-trust; only the StageX (amd64) path does.

This is also why I'd keep Nix (or StageX) for **publishing binaries**, not just for the image: the `.deb`s and the standalone binaries inherit the same source-defined/bootstrapped toolchain, so a user installing via `dpkg`/apt gets the same provenance as someone pulling the image — the trust doesn't depend on which artifact they consume.

y4ssi:
One more practical distinction worth adding to the table above — the **binary type**, which is a separate reason the published artifacts use Nix/StageX rather than the reproducible Dockerfile:

| | reproducible `Dockerfile` | Nix (arm64) | StageX (amd64) |
|---|---|---|---|
| Binary extractable without running the image? | ✅ (`--target export`) | ✅ | ✅ |
| Reproducible (rebuild-deterministic)? | ✅ | ✅ | ✅ |
| **Binary type** | **glibc-dynamic** (needs a libc at runtime) | **musl-static** | **musl-static** |
| Toolchain trust | prebuilt `rust`+`debian` bases, digest-pinned (opaque blobs) | source-defined, hash-pinned, rebuildable graph | bootstrapped from a hex0 seed |

All three give you an extractable, reproducible binary — the reproducible `Dockerfile` even exposes a `scratch` `export` stage, so `docker buildx build --target export --output type=local,dest=./out .` drops the binary straight out (that's exactly how I rebuild-verified it). So for a developer who just wants to build it themselves, the Dockerfile is fine.

But for the **published** artifact the difference matters on two axes: (1) toolchain trust — the Docker binary is reproducible *against opaque Debian/rust base layers* (if an upstream base were compromised, the build would reproduce the compromise perfectly; "reproducible" only guarantees determinism, not provenance), whereas Nix reproduces against an auditable source graph and StageX from a 512-byte seed; and (2) the Docker binary is **glibc-dynamic** (links the Debian libc, needs a matching libc at runtime), while Nix/StageX produce a **musl-static** binary that runs on any distro with no libc dependency — which is what you want to ship in a `.deb`/release. So: extractable everywhere, but the release path picks Nix/StageX for the stronger toolchain provenance *and* the portable static binary.

zcash-harry (approved):
- Built zallet 0.1.0-alpha.4 (aarch64-linux-musl) inside a Linux arm64 Docker container
- Build succeeded ✅

nuttycom (approved):
lightly-reviewed utACK

PR #510, opened 2026-06-24
Exit before consensus divergence with backing node

### LINKED ISSUE #190: Shut down and refuse to start if we detect the backing full node has an incompatible set of consensus rules

sellout:
Fixes #190.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/zallet/src/components/chain/zaino.rs
+++ b/zallet/src/components/chain/zaino.rs
@@ -37,11 +37,13 @@ use zcash_protocol::{
 };
 #[cfg(not(feature = "spend-index"))]
 use zebra_rpc::client::{GetAddressBalanceRequest, GetAddressTxIdsRequest};
+use zebra_rpc::methods::{ConsensusBranchIdHex, NetworkUpgradeInfo, NetworkUpgradeStatus};
 
 use crate::{
     components::TaskHandle,
     config::ZalletConfig,
     error::{Error, ErrorKind},
+    fl,
     network::Network,
 };
 
@@ -249,6 +251,32 @@ impl ZainoChain {
         });
 
         Ok((chain, task))
+    }
+
+    /// Refuses to continue if the backing full node follows consensus rules that
+    /// this build of Zallet does not recognize.
+    pub(crate) async fn check_consensus_compatibility(&self) -> Result<(), Error> {
+        let info = self
+            .fetcher
+            .get_blockchain_info()
+            .await
+            .map_err(|e| ErrorKind::Init.context(e))?;
+
+        let unknown = detect_unknown_upgrades(&info.upgrades);
+        if unknown.is_empty() {
+            info!("Backing full node consensus rules are compatible with this Zallet build");
+            return Ok(());
+        }
+
+        let upgrades = unknown
+            .iter()
+            .map(UnknownUpgrade::describe)
+            .collect::<Vec<_>>()
+            .join(", ");
+        error!("Backing full node follows unrecognized consensus rules: {upgrades}");
+        Err(ErrorKind::Init
+            .context(fl!("err-init-incompatible-consensus", upgrades = upgrades))
+            .into())
     }
 }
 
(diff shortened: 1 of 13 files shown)


zallet/src/components/chain/zaino.rs
```diff
@@ -203,6 +205,32 @@ impl ZainoChain {
...
+
+        let unknown = detect_unknown_upgrades(&info.upgrades);
+        if unknown.is_empty() {
+            info!("Backing full node consensus rules are compatible with this Zallet build");
+            return Ok(());
+        }
+
+        let upgrades = unknown
+            .iter()
+            .map(UnknownUpgrade::describe)
```
nullcopy: I wonder if we should only block if any `UnknownUpgrade` has activated yet. Not-yet-activated unknown upgrades should not block startup, IMO

zallet/src/components/chain/zaino.rs
```diff
@@ -579,3 +607,123 @@ impl ChainBlock {
...
+
+/// A network upgrade that the backing full node follows but that this build of
+/// Zallet does not recognize.
+struct UnknownUpgrade {
+    /// The consensus branch ID reported by the full node.
+    branch_id: u32,
+    /// The full node’s name for the upgrade, used for diagnostics only.
+    name: String,
+    /// The activation height, if the upgrade is pending rather than already active.
+    pending_at: Option<u32>,
```
nullcopy: I think even for activated NUs, we should confirm that they were activated at the height which we expect. So maybe this should become a non-optional `activation_height` instead, and we also bubble up the `NetworkUpgradeStatus`

zallet/src/components/chain/zaino.rs
```diff
@@ -607,3 +635,123 @@ impl ChainBlock {
...
+    }
+}
+
+/// Identifies the network upgrades reported by the backing full node whose
+/// consensus branch IDs this build of Zallet cannot interpret.
+///
+/// Upgrades with [`NetworkUpgradeStatus::Disabled`] are ignored: they have no
+/// activation height on the current network, and so will never take effect.
+fn detect_unknown_upgrades<'a>(
+    upgrades: impl IntoIterator<Item = (&'a ConsensusBranchIdHex, &'a NetworkUpgradeInfo)>,
```
dannywillems: Nit: maybe extract the output in a structure? It is probably going to be used in other places. I think the config uses it?

nullcopy (dismissed):
In addition to the comments below, I think this same check needs to be implemented for the `ZebraChain` implementation which got added recently

Reviewed as of e099c3d6d119a3970738419d8cf3808c5ae1cec8

zallet/src/components/chain/zaino.rs
```diff
@@ -607,3 +635,123 @@ impl ChainBlock {
...
+        .into_iter()
+        .filter_map(|(branch, info)| {
+            let branch_id = branch.inner();
+
+            // If we can map the branch ID onto a set of consensus rules we were
+            // compiled with, then we understand this upgrade.
+            if consensus::BranchId::try_from(branch_id).is_ok() {
+                None
+            } else {
+                // `NetworkUpgradeInfo` is `Copy`, and `into_parts` takes `self`.
```
dannywillems: Unnecessary comment.

zallet/src/components/chain/zaino.rs
sellout: Ah, this was in a comment that I forgot to bring over from the original PR:

> There is further work that can be done here – some of it described in https://github.com/zcash/zallet/issues/190.
>
> -  have the check be bi-directional, so that if Zallet knows of NUs that the backing node doesn’t, we also fail; and
> - delay failure until the activation block of an unknown NU, warning at startup that it’ll eventually shut down. This allows either Zallet or the backing node to be upgraded at any time between the time the issue is identified and the activation point of the mismatched NU.

When I mentioned it in standup, @nuttycom suggested just doing the simplest thing for now. E.g., if not-yet-activated upgrades don’t stop startup, we have to make sure they fail when we get to that height.

sellout: Ok, I was spending more time dealing with merge conflicts than anything else, so I implemented all of this rather than doing it later.

So now it exits immediately if there’s an incompatibility between active NUs and otherwise warns at startup, then exits when it reaches the point at which they could diverge.

zallet/src/components/chain/zaino.rs
sellout: I added this as well.

zallet/src/components/chain/zaino.rs
sellout: Removed.

sellout:
Implemented for `ZebraChain`, too.

nuttycom (approved):
utACK 9af8a1710b5b7efd1c24944a7f6419ca02f5269a

Issue #511, opened 2026-06-24
sync: fall back to hopping back toward the wallet birthday when the fork point can't be found


nuttycom:
Follow-up from the PR #493 review (thread on `zallet/src/components/sync/locator.rs`).

## Problem

`resolve_fork_point` (`zallet/src/components/sync.rs`) handles `ChainView::find_fork_point` returning `None` in two ways:

- **empty locator** → assume no reorg, return `prev_tip`
- **non-empty locator, but no entry on the backend's best chain** → **hard error** ("Could not determine the reorg point: the wallet's previous chain tip … is not known to the chain indexer")

The hard-error branch is reachable when the wallet's divergence from the backend's best chain is deeper than the locator window (`MAX_LOCATOR_DEPTH` = 1000, matching Zebra's `MAX_BLOCK_REORG_HEIGHT`) — e.g. the wallet was offline across a deep reorg, or its last-synced blocks were on a side chain that has since been pruned. In that situation steady-state sync errors on every iteration and makes no forward progress.

## Desired behavior

Mirror the mobile wallets: when the fork point cannot be located, fall back to **hopping back toward the wallet birthday by height** instead of erroring, re-attempting from progressively lower heights until a chaining point is found (or the birthday is reached).

Open design questions:
- Hop-back step size — fixed amount, or exponential like the locator heights?
- The starting point for the hop-back. Per the review discussion, ideally the height at which the hash failed to chain.
- Termination at the wallet birthday.
- This is backend-neutral sync-engine behavior (applies to both the `zaino` and `zebra-state` backends).

🤖 Filed with Claude Code assistance.

PR #512, opened 2026-06-24
sync: step back along the wallet's view to recover from deep forks

nuttycom:
On top of `zebra-chain-view`. Addresses #511.

## Problem

When the wallet's recorded tip is on a fork below the backend's non-finalized window — e.g. it was offline while the chain advanced past its last-synced position — `find_fork_point` cannot locate the divergence (the wallet's locator is entirely off the best chain, and the non-finalized fork-point search can't see the wallet's much-lower blocks). Today that surfaces as a hard error and the wallet is stuck.

## Change

The reorg-window `find_fork_point` query stays the fast path for ordinary reorgs. When it can't resolve the divergence, the sync engine falls back to the mobile-wallet behaviour: walk the wallet's own view of the chain back 10 blocks at a time, comparing each of the wallet's recorded block hashes against the backend's best chain (`get_block_header(height)`), and resume scanning from the first block that matches. If the walk reaches the wallet birthday without rejoining the best chain, halt with a new `SyncError::WalletDivergedBelowBirthday` rather than rescanning blindly.

## Tests

- Unit tests for the step arithmetic (`rewind_step`: step size, birthday floor, halt).
- A mock-`ChainView` test for the step-back walk: it finds the resume point, and halts at the birthday when the wallet never rejoins the best chain.

Both backends compile; the sync unit tests are green.

### DIFF
--- a/zallet/src/components/sync.rs
+++ b/zallet/src/components/sync.rs
@@ -58,6 +58,7 @@ use zcash_client_backend::{
     sync::decryptor,
 };
 use zcash_client_sqlite::AccountUuid;
+use zcash_primitives::block::BlockHash;
 #[cfg(not(feature = "spend-index"))]
 use zcash_protocol::TxId;
 use zcash_protocol::consensus::BlockHeight;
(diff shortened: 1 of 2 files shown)


nullcopy (approved):
utACK 24ba3e19d2866e4abd7ed6707cdaca402ed77977

nullcopy (approved):
utACK 3788a0a10a26623c99f1ea9b3cbabe3d8505952d

Issue #507, opened 2026-06-23
Use a backend transparent-address index for TransactionsInvolvingAddress when available (full-block scan as fallback)


nuttycom:
## Summary

zallet currently does **all** transparent-output detection by scanning full blocks: during sync, `zcash_client_backend`'s scanner walks the transparent outputs of every transaction in every block to find any paying the wallet's transparent addresses. The sync engine explicitly ignores `TransactionDataRequest::TransactionsInvolvingAddress`:

```rust
// Ignore these, we do all transparent detection through full blocks.
TransactionDataRequest::TransactionsInvolvingAddress(_) => (),
```

This is correct, but it examines every transaction's transparent outputs across the whole scanned range — expensive, and the reason chain-wide `OP_RETURN` outputs get walked at all.

## Desired behavior

zallet should service `TransactionsInvolvingAddress` through the chain-backend abstraction, with **two alternative paths** depending on backend capability:

1. **Indexed path** — when the backend exposes a transparent-address index (transactions/outputs involving a given t-address), answer the query directly and skip the full-block walk.
2. **Fallback path** — when no index is available, fall back to the current full-block transparent scan.

The `Chain`/`ChainView` abstraction should surface this as a per-backend capability, so the sync engine takes the cheaper path when it exists and the correct fallback otherwise.

## Follow-ups this spawns

Providing the index is a **backend (Zaino) responsibility**, so this is a parent that will spawn more specific requirements, e.g.:
- a Zaino-side requirement to expose a transparent-address index;
- a zallet-side abstraction for the capability and the two paths above.

_Related: the noisy "unsupported script kind 'nulldata'" warnings seen during full-block scanning are being addressed separately by treating `OP_RETURN` as a recognized data output in `zcash_client_backend`._

---
🤖 Filed with Claude Code assistance.

PR #517, opened 2026-06-25
sync: recover missed transparent spends via per-outpoint spend index

nuttycom:
On top of `zebra-chain-view`. Addresses #507.

## Problem

Transparent spends can be missed by chain scanning. `zcash_client_sqlite` records a transparent spend only when the spending input's outpoint is already present in `transparent_received_outputs` at the time the spending block is processed. Because scanning is priority-based (out of order), a spend block processed *before* the block that created the output is never matched, and there is no retroactive re-scan. The wallet's designed recovery mechanism — `TransactionDataRequest::TransactionsInvolvingAddress` — was previously ignored by the sync engine, so these spends were never recovered and the affected outputs were reported as unspent.

Observed live: a testnet wallet (imported from `zcashd`) over-reported its transparent balance by 15 ZEC, with 6 spent-on-chain outputs shown as unspent at an imported address.

## Change

Implements the `zebra-state` side of per-outpoint spend recovery, using the new `spend-index` feature in zcash/librustzcash#2463. With `spend-index` enabled, `transaction_data_requests` emits `TransactionDataRequest::GetSpendingTx(outpoint)` for each unspent transparent output, and the sync engine services it via a new `ChainView::outpoint_spend_status`:

- Spentness is read **authoritatively** from the node's UTXO set (`ReadRequest::UnspentBestChainUtxo`), independent of the lazily-built spend index.
- The spending transaction is resolved via `ReadRequest::SpendingTransactionId` (Zebra's `indexer` feature) and ingested with `decrypt_and_store_transaction`.
- A spent output whose spender is not yet indexed — the index is backfilled lazily with no completion signal, see ZcashFoundation/zebra#10806 — is left queued and retried, never treated as unspent.

When the output is confirmed unspent through the snapshot tip, `notify_output_verified_unspent` advances the per-outpoint watermark so the request is not re-issued.

The `zebra-state` feature now enables `zebra-state/indexer` and the new zallet `spend-index` feature (→ `zcash_client_sqlite/spend-index`).

The `zaino` backend (with `spend-index` off) continues to build with no regression; its address-index-based recovery (the light-server `getaddresstxids`/`getaddressutxos` fallback — the other half of #507) is a follow-up.

## Tests / verification

- `zebra-state`: builds + 128 unit tests pass; clippy clean.
- `zaino`: builds (no regression).
- The librustzcash side (zcash/librustzcash#2463) adds a `spend_index_queries_are_valid_sql` test exercising both new SQL statements.
- Live end-to-end recovery (the 15 ZEC on the affected testnet wallet) is still to be verified against a running `indexer`-enabled zebrad.

## Notes

- Depends on zcash/librustzcash#2463. `[patch.crates-io]` temporarily pins the librustzcash crates to rev `22db69ca` until that PR merges and is released.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,19 +158,20 @@ anyhow = "1.0"
 tonic = "0.14"
 
 [patch.crates-io]
-# TEMPORARY: Pin librustzcash crates to upstream main
-equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
+# TEMPORARY: pin librustzcash crates to the `spend-index` PR (zcash/librustzcash#2463)
+# until it merges and is released.
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
+zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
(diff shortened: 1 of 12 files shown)


dannywillems (approved):
utACK 84a4d7cc70fe2a743d0d2231e1d90cb7d39c7cd9. Only uses the librustzcash new routines.

nuttycom:
Draft until I have tested the fix.

zallet/src/components/chain/zaino.rs
```diff
@@ -536,6 +542,51 @@ impl ChainView for ZainoChainView {
         })
     }
 
+    #[cfg(not(feature = "spend-index"))]
+    async fn get_address_unspent_outpoints(
```
nullcopy, line 546: I don't understand why `ZainoChain` loses functionality when we enable `spend-index`

zallet/src/components/chain/zaino.rs
```diff
@@ -536,6 +542,51 @@ impl ChainView for ZainoChainView {
...
+            .get_address_utxos(GetAddressBalanceRequest::new(vec![addr_str]))
+            .await
+            .map_err(ChainError::backend)?;
+        Ok(utxos
+            .into_iter()
+            .map(|utxo| (TxId::from_bytes(utxo.txid().0), utxo.output_index().index()))
+            .collect())
+    }
+
+    #[cfg(not(feature = "spend-index"))]
```
nullcopy, line 562: Same as above

Cargo.toml
```diff
@@ -158,19 +158,20 @@ anyhow = "1.0"
...
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "aeabc1a9160bdf5511e0d22e709fbc1d307f6853" }
+# TEMPORARY: pin librustzcash crates to the `spend-index` PR (zcash/librustzcash#2463)
+# until it merges and is released.
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "4cdba554f5545050da784fc9e1eb073ae958d819" }
```
nullcopy, line 163: (nit) librustzcash is now at `4fa0503e6cf5001ddcb4add6f116f1bb64f3954b`

zallet/src/components/sync.rs
```diff
@@ -531,6 +542,88 @@ async fn recover_history<C: Chain>(
...
+        .into_iter()
+        .collect();
+    let our_outputs = db_data.get_spendable_transparent_outputs(
+        &address,
+        TargetHeight::from(view_tip + 1),
+        ConfirmationsPolicy::MIN,
+        TransparentOutputFilter::All,
+    )?;
+    let any_spent = our_outputs.iter().any(|output| {
+        let outpoint = output.outpoint();
```
nullcopy, line 603: (nit) I would handle this in a `.map(|o| o.outpoint())` iterator adapter

nuttycom:
Self-ACK; I have verified that this correctly restores a zcashd wallet *and* discovers a transparent address (from ZODL reuse of the seed) that zcashd didn't know about. I have not tested the Zaino implementation here though.

nullcopy (commented):
flushing comments

Cargo.toml
nuttycom: We can't update beyond the current pin because that brings in a breaking change to the `zcash_primitives` API that neither Zebra nor Zaino can build against.

nuttycom: The comment is now wrong though.

zallet/src/components/chain/zaino.rs
nuttycom: `spend-index` isn't a per-backend capability toggle — it's enabled only by the `zebra-state` backend feature, which is mutually exclusive with `zaino`. So `ZainoChain` is only ever compiled with `spend-index` *off*, and these `get_address_*` methods are exactly the spend-search path used under the Zaino backend.

The feature gates the `ChainView` *capability* — per-outpoint `outpoint_spend_status` when the backend exposes a spend index, vs. address-based discovery otherwise — rather than removing anything from `ZainoChain`. (`zaino` + `spend-index` together wouldn't even compile, since `ZainoChain` doesn't implement `outpoint_spend_status`; that's the intended invariant.) Once Zaino exposes a per-outpoint spend/nullifier index, `ZainoChain` can implement the per-outpoint path under `spend-index` too.

Happy to add a clarifying comment near these gates if that'd help future readers.

*Drafted with [Claude Code](https://claude.com/claude-code) assistance.*

Cargo.toml
nuttycom: https://github.com/zcash/zallet/pull/522

zallet/src/components/sync.rs
nuttycom: I think that the existing code here is fine.

PR #521, opened 2026-06-25
Make the sync history-recovery batch size configurable

nuttycom:
Adds a `[sync] recover_batch_size` configuration option controlling the maximum number of blocks the history-recovery task downloads and trial-decrypts per batch.

## Motivation

The batch size was previously hardcoded to 1000. Larger batches improve scanning throughput on well-resourced machines, at the cost of higher peak memory usage: each block in a batch is held in memory while it is downloaded and trial-decrypted, and mainnet blocks can currently be up to ~2 MiB, so a batch of `N` blocks needs on the order of `N × 2 MiB` of memory. Making it configurable lets operators trade memory for sync speed based on their hardware.

## Changes

- New `[sync]` config section with `recover_batch_size` (`Option<NonZeroU32>`; a value of `0` is rejected at parse time). The default is **1000**, preserving the previous behaviour.
- The field is `#[serde(default)]`, so configuration files written before this section existed continue to parse without a `[sync]` section.
- `recover_history` now uses the configured value instead of the hardcoded constant.
- Documented in the book (`start.md` → "Tuning history recovery") and in the generated example config.
- Unit tests covering the default, an override, and rejection of `0`; regenerated the `example-config` and `migrate-zcash-conf` golden snapshots.

### DIFF
--- a/book/src/cli/start.md
+++ b/book/src/cli/start.md
@@ -8,3 +8,24 @@ begin listening for JSON-RPC connections.
 
 You can shut down a running Zallet wallet with Ctrl+C if `zallet` is in the foreground,
 or (on Unix systems) by sending it the signal `SIGINT` or `SIGTERM`.
+
+## Tuning history recovery
+
+When Zallet syncs a wallet for the first time (or recovers history for newly imported
+keys), it downloads and trial-decrypts historical blocks in batches. The maximum number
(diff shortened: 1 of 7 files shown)


nuttycom (commented):
self-utACK

nullcopy (approved):
utACK 312cfbf2015316c22ff9473764cc309c059fc2f7

PR #525, opened 2026-06-26
ci: lint main Dockerfile with hadolint and pin apt versions

### LINKED ISSUE #527: lint main Dockerfile with hadolint and pin apt versions

dannywillems:
The hadolint job only linted Dockerfile.stagex, leaving the main Dockerfile unchecked. Run hadolint over both via a matrix so the required-lints gate covers each one.

Pin the apt package versions in Dockerfile to the bookworm snapshot the digest-pinned base images resolve to, fixing the two DL3008 warnings without suppressing the rule. Bump these pins together with the base digests when a Debian point update supersedes a version.

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -106,8 +106,14 @@ jobs:
         run: cargo fmt --all -- --check
 
   hadolint:
-    name: Hadolint (Dockerfile)
+    name: Hadolint (${{ matrix.dockerfile }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        dockerfile:
+          - Dockerfile
+          - Dockerfile.stagex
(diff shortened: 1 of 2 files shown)


y4ssi (approved):
LGTM. Extending hadolint to the main Dockerfile via the matrix is a good catch — it was previously unlinted. Pinning the apt versions (rather than suppressing DL3008) is the correct fix and keeps the package set consistent with the digest-pinned bases for reproducibility.

Verified all CI is green, including both Hadolint matrix jobs and the Docker builds (build_and_run + Latest build on ubuntu-latest), which confirms the pinned versions resolve and install cleanly from the bookworm snapshot.

Nit (non-blocking): the inline comments already note to bump these pins with the base digests — worth keeping that discipline so a Debian point update doesn't silently fail the build.

PR #530, opened 2026-06-29
Payment: use methods on steps

dannywillems:
Old patch, but still useful. Small refacto.

Depends on https://github.com/zcash/librustzcash/pull/2497.

Would also require an update on the zebra and zaino sides.


nuttycom (approved):
utACK

PR #534, opened 2026-06-30
docs(book): document zaino backend selection and --no-default-features requirement

### LINKED ISSUE #533: Building with --features zaino fails due to default zebra-state feature conflict
## Problem

Running `cargo build --features "rpc-cli zaino"` (or any variant that adds `zaino` to the default feature set) fails with three compiler errors:

```
error[E0428]: the name `ChainBackend` is defined multiple times
  --> zallet/src/components/chain.rs:48:1

error: the `zaino` and `zebra-state` chain backends are mutually exclusive; enable exactly one (e.g. `--no-default-features --features zaino`)
  --> zallet/src/lib.rs:20:1

error[E0046]: not all trait items implemented, missing: `outpoint_spend_status`
  --> zallet/src/components/chain/zaino.rs:333:1
```

## Root cause

`zebra-state` is listed in `[features] default` in `zallet/Cargo.toml`. Cargo's `--features` flag *adds* to the active feature set; it does not replace defaults. So `--features "rpc-cli zaino"` resolves to `rpc-cli + zaino + zebra-state`, activating both mutually exclusive backends simultaneously.

The three errors all cascade from this: the `ChainBackend` type alias is defined twice (once per backend), the `compile_error!` guard fires, and the `spend-index` feature (enabled by `zebra-state`) requires `outpoint_spend_status` on `ZainoChainView`.

## Correct invocation

The build works correctly with:

```
cargo build --no-default-features --features "rpc-cli zaino"
```

CI already uses `--no-default-features` for both backends, so this was never caught there.

## Fix

Document the `--no-default-features` requirement in the Zallet book:

- Add a "Choosing a chain backend" section to the installation guide with a comparison table and the correct `cargo install` commands for each backend.
- Add a cross-reference in the setup guide's backend section pointing to the installation page.
- Explicitly call out that `--features zaino` without `--no-default-features` activates both backends and produces a compile error.

pacu:
## Summary

- Add a **"Choosing a chain backend"** section to the installation guide with a comparison table (zebra-state vs zaino: default, platform support, requirements), prose explaining when to use each, and correct `cargo install` commands using `--no-default-features`.
- Explicitly warn that adding `--features zaino` without `--no-default-features` activates both mutually exclusive backends simultaneously and produces a compile error.
- Add a cross-reference in `setup.md`'s backend section pointing to the new installation content.

Closes #533

## Test plan

- [ ] Verify table and prose in `guide/installation/README.md` are accurate against `zallet/Cargo.toml` feature definitions.
- [ ] Confirm the `cargo install` commands in the new section match the commands CI uses (`--no-default-features --features zaino,rpc-cli`).
- [ ] Build the book locally with `mdbook build` and check the rendered output.

### DIFF
--- a/book/src/guide/installation/README.md
+++ b/book/src/guide/installation/README.md
@@ -61,3 +61,35 @@ directly from the main branch of its code repository:
 ```
 cargo install --locked --git https://github.com/zcash/wallet.git
 ```
+
+### Choosing a chain backend
+
+Zallet supports two chain backends, selected at compile time:
+
+| Backend | Default | Platform | Requires |
+|---------|---------|----------|---------|
+| `zebra-state` | Yes | Linux only | co-located `zebrad` with `indexer` feature + `[indexer.read_state_service]` config |
+| `zaino` | No | Linux, macOS, Windows | co-located `zebrad` JSON-RPC endpoint |
+
+The **`zebra-state` backend** is the default. It reads finalized chain state directly from
+a co-located `zebrad`'s state database and is the recommended choice for production
+mainnet use on Linux.
+
+The **`zaino` backend** fetches chain data over JSON-RPC (and optionally reads state
+directly when `[indexer.read_state_service]` is configured). It is the only backend that
+supports regtest and runs on non-Linux platforms.
+
+To build or install with the `zaino` backend, pass `--no-default-features --features zaino`
+(along with any other features you need):
+
+```
+# Install from crates.io with the zaino backend
+cargo install --locked --no-default-features --features zaino,rpc-cli zallet
+
+# Install the latest development version with the zaino backend
+cargo install --locked --git https://github.com/zcash/wallet.git \
+  --no-default-features --features zaino,rpc-cli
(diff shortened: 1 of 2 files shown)


pacu:
Closing in favor of https://github.com/zcash/zallet/pull/538

y4ssi:
Heads-up: #538 (dual-backend CI/CD) touches the same doc files (`installation/README.md` + `setup.md`), so to avoid a conflict I incorporated this PR's "Choosing a chain backend" section into #538 — the comparison table (including the Linux-only vs cross-platform distinction), the `--no-default-features --features zaino` build commands, and the mutual-exclusivity compile-error warning are all carried over verbatim in spirit, and extended with the pre-built-artifact split (the `zallet` / `zallet-zebra-state` / `zallet-zaino` binaries the image + deb now ship). Credit to you for the source-build docs.

Either this can merge first and #538 rebases on it, or #538 can supersede it — whichever you and the reviewers prefer. Both close #533.

PR #535, opened 2026-07-01
Extract the shared amount parameter into a reusable payments type

### LINKED ISSUE #217: rpc: Design and implement `z_sendfromaccount`

dannywillems:
Extracting logic from https://github.com/zcash/zallet/pull/531. The idea is to have !531 as small as possible, and merge relatively unrelated changes to main, to decrease the reviewer time on the actual feature, while we're merging NU6.3 related changes.

-----

Move the `AmountParameter` request type out of `z_send_many` into `json_rpc::payments` (with accessor methods) and extract the shared `build_request` amount-parsing helper, so the type and parsing can be reused by other send-path RPC methods. Update `z_sendmany`, the method trait, and the OpenRPC generation to reference the relocated type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -616,6 +616,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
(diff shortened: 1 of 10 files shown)


nuttycom (approved):
utACK; essentially move-only.

PR #538, opened 2026-07-01
ci: ship both zebra-state and zaino backends in one image + deb

y4ssi:
## Problem

Zallet's chain-data backend is a **compile-time, mutually-exclusive** Cargo feature:

- **`zebra-state`** (default) — reads a co-located `zebrad`'s state directly via `ReadStateService`; **requires a `zebrad` built with the non-default `indexer` feature** + a shared state dir; **no regtest**.
- **`zaino`** (`--no-default-features --features zaino`) — talks to Zebra over **JSON-RPC** in separate containers; **supports regtest**.

The published `v0.1.0-alpha.4` image ships only the default `zebra-state` backend, so it can't run in split-container stacks like [z3](https://github.com/ZcashFoundation/z3): `zallet start` fails with `the zebra-state backend requires an [indexer.read_state_service] config section`. Because the backend is fixed at Zallet compile time, bumping Zebra/Zaino doesn't help — a zaino-backend image has to be published. This blocks [z3#49](https://github.com/ZcashFoundation/z3/pull/49).

## What this does

Build **both** backends and ship them together, **additively** — the existing `zallet` command is unchanged. Three invocable names:

| Name | Kind | Backend | Needs `zebrad` built with `indexer`? | Regtest |
|------|------|---------|:---:|:---:|
| **`zallet`** | real binary | `zebra-state` (default) | **Yes** | No |
| **`zallet-zebra-state`** | symlink → `zallet` | `zebra-state` | **Yes** | No |
| **`zallet-zaino`** | real binary | `zaino` | No | **Yes** |

`zallet` is the default and is byte-for-byte the previous default build. `zallet-zebra-state` is just an explicit alias (a symlink to the same binary). `zallet-zaino` is the only new binary.

> **Which one should I run?** `zallet` / `zallet-zebra-state` (zebra-state) **only work against a `zebrad` built with the non-default `indexer` feature** (plus a shared state dir). If your `zebrad` is **not** built with `indexer` — including the stock `zfnd/zebra` images, the z3 stack, any setup where Zebra and Zallet are separate containers over JSON-RPC, or regtest — run **`zallet-zaino`**. All three share the same CLI surface, config format, and subcommands; only the chain-data backend differs.

## Resulting artifacts

| Deliverable | zebra-state (default) | zaino |
|-------------|-----------------------|-------|
[...]

PR #539, opened 2026-07-02
fix(ci): stage the zallet-zaino binary in the release deb/binary job

y4ssi:
## Summary
- The "Stage exported binaries" step in `binaries-and-deb-release.yml` only ever copied `zallet` out of the runtime artifact root, never `zallet-zaino`, even though PR #538 made both binaries part of the runtime export (`Dockerfile.stagex` export stage for amd64, `build-arm64-nix.yml` for arm64).
- Two later steps in the same job already assumed `build/${PLATFORM_KEY}/zallet-zaino` existed (`cp` for the standalone binary artifact, `install` for `.deb` staging) and both fail with `No such file or directory` since nothing ever staged it.
- Adds the missing `install`, plus a matching sanity check for `zallet-zaino` mirroring the existing one for `zallet`.

## How found
Re-ran the release pipeline against the existing `v0.1.0-alpha.4` tag after #538 merged (`gh workflow run release.yml -f tag=v0.1.0-alpha.4`, run [28603304440](https://github.com/zcash/zallet/actions/runs/28603304440)). The container build + multi-arch manifest jobs succeeded, but both `linux-arm64` and `linux-amd64` legs of "Build standalone binaries and Debian packages" failed/cancelled on:
```
cp: cannot stat 'build/linux-arm64/zallet-zaino': No such file or directory
```

## Test plan
- [ ] `actionlint` passes (already verified locally, no errors)
- [ ] Re-run `release.yml` against `v0.1.0-alpha.4` after merge and confirm the deb/binary matrix jobs succeed for both architectures, producing `zallet-*-zaino` standalone binaries and `.deb`s with both `zallet-zaino` and `zallet-zebra-state` installed

### DIFF
--- a/.github/workflows/binaries-and-deb-release.yml
+++ b/.github/workflows/binaries-and-deb-release.yml
@@ -257,9 +257,16 @@ jobs:
           PLATFORM_KEY="${{ matrix.platform_key }}"
           ARTIFACT_ROOT="build/runtime-artifact/${PLATFORM_KEY}"
 
-          # Basic sanity check: ensure the exported binary exists at the root (as /zallet in the export image)
+          # Basic sanity check: ensure both backend binaries exist at the root
+          # (as /zallet and /zallet-zaino in the export image; zebra-state is
+          # the default binary name, zaino ships as the additive zallet-zaino).
           if [[ ! -s "${ARTIFACT_ROOT}/zallet" ]]; then
             echo "Missing zallet binary under ${ARTIFACT_ROOT}" >&2
+            find "${ARTIFACT_ROOT}" -maxdepth 5 || true
+            exit 1
+          fi
+          if [[ ! -s "${ARTIFACT_ROOT}/zallet-zaino" ]]; then
+            echo "Missing zallet-zaino binary under ${ARTIFACT_ROOT}" >&2
             find "${ARTIFACT_ROOT}" -maxdepth 5 || true
             exit 1
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/binaries-and-deb-release.yml
```diff
@@ -257,18 +257,28 @@ jobs:
           PLATFORM_KEY="${{ matrix.platform_key }}"
           ARTIFACT_ROOT="build/runtime-artifact/${PLATFORM_KEY}"
 
-          # Basic sanity check: ensure the exported binary exists at the root (as /zallet in the export image)
+          # Basic sanity check: ensure both backend binaries exist at the root
+          # (as /zallet and /zallet-zaino in the export image; zebra-state is
```
nuttycom, line 261: We should probably have `zallet-zebra` and `zallet-zaino` so as not to articulate a preference - we should update the book to ensure that users have enough information to make the choice themselves for which they want to run. In the future (https://github.com/zcash/zallet/issues/540) `zallet` may become a wrapper that chooses between implementations based on configuration (zallet-zebra and zallet-zaino binaries will both persist.)

gustavovalverde: Adding to this, but unrelated to this PR. This should preferably be a runtime configuration rather than a build-time configuration (which requires multiple binaries).

nuttycom: Yup, that's the path described by #540

nuttycom: See the rationale for multiple binaries in #540 - we can't have a slow-to-update Zaino block upgrading the Zebra backend.

gustavovalverde (approved):

.github/workflows/binaries-and-deb-release.yml
nuttycom: (I agree that it's not a blocker for this PR, can be a followup)

PR #541, opened 2026-07-02
rpc: Add JSON-RPC methods for working with PCZTs

### LINKED ISSUE #99: rpc: Add JSON-RPC methods for working with PCZTs
These will serve as the functional replacements for `createrawtransaction`, `fundrawtransaction`, and `signrawtransaction` (which operate directly on transaction encodings and are thus very brittle).

sellout:
Implements the PCZT (Partially Created Zcash Transaction) JSON-RPC methods requested in #99, as the modern, robust replacements for `createrawtransaction`, `fundrawtransaction`, and `signrawtransaction`.

Supersedes #354 (thanks @lamb356 for the initial implementation) — this branch rebuilds that work on current `main`, addresses all of @nullcopy's review feedback, and fills in the structural gaps.

Closes #99.

## Methods

A transaction is assembled by threading a PCZT through these roles:

| Method | Role(s) | Notes |
|---|---|---|
| `pczt_create` | `propose_transfer` + `create_pczt_from_proposal` | Build a funded PCZT from a payment request (≈ `createrawtransaction` + `fundrawtransaction`). |
| `pczt_combine` | Combiner | Merge several parties' contributions. |
| `pczt_prove` | Prover | Add the Sapling/Orchard proofs (run off the async runtime; caches the Orchard proving key). |
| `pczt_sign` | Signer | Sign with the wallet's keys. |
| `pczt_extract` | SpendFinalizer + TransactionExtractor | Finalize spends and **verify** before extracting the final transaction. |

The stateless roles (`combine`/`prove`/`extract`) live in the general `Rpc` trait; `create`/`sign` require wallet state.

## Notes for review

- **`pczt_create` vs `pczt_fund`.** #354 named this `pczt_fund`; @nullcopy noted that "fund" reads like *adding funds to an existing PCZT*. That operation (≈ `fundrawtransaction` on a partial PCZT) is effectively infeasible at the pinned `librustzcash` rev: `create_pczt_from_proposal` IO-finalizes its output (permanently clearing `tx_modifiable`), there is no `Constructor` implementation or Builder-from-PCZT path, and no input-selector API for a "cover the remaining value" target. So this method builds a *complete* PCZT from a payment request in one shot (matching `zcash-devtool`'s `pczt create`), and `pczt_create` is the accurate name.
- **Shared logic.** `pczt_create` reuses `payments::AmountParameter` + `build_request` (from #535) and a new shared `payments::propose_and_check` helper (privacy-policy validation → propose → enforce → Orchard-action-limit) that `z_sendmany` now also uses, so the two stay in sync.
- **Signing hints.** pczt 0.7 exposes no public getter for the native BIP 32 / ZIP 32 derivation metadata, so `pczt_create` records `zallet.v1.*` proprietary fields for an offline `pczt_sign` to read back — a stand-in for the native path until upstream adds accessors.
- **`pczt_extract` verifies.** It runs the spend finalizer and verifies all proofs/signatures (supplying the bundled Sapling verifying keys) before producing the transaction, rather than deferring to the network.

## Dependencies

Adds `pczt` 0.7, pinned to the same `librustzcash` revision as the other `[patch.crates-io]` crates, and enables the `zcash_client_backend/pczt` builder support plus `zcash_client_sqlite/serde` (required for `AccountUuid: Serialize`, which `create_pczt_from_proposal` needs).

## Testing

- Unit tests for the base64 decode size/format limits and the `pczt_combine` caps.
- `cargo clippy --all-targets -- -D warnings` clean; full `zallet` lib test suite green.
- :warning: **No automated end-to-end shielded test.** The `zcash_client_backend` `data_api::testing` framework can't be used from a downstream crate at the pinned rev — its only concrete store factory (`zcash_client_sqlite::testing`) is `#[cfg(test)]`-only, and there is no `zcash_client_memory` crate at this rev — so the full `create → prove → sign → extract` path is verified by **manual regtest** for now. A follow-up could vendor a minimal in-memory store factory to unlock a node-free CI test.

---

Rebased onto `main` @ 4fa38e1. The commits on this branch are currently unsigned.


schell:
> there is no zcash_client_memory crate at this rev — so the full create → prove → sign → extract path is verified by manual regtest for now. A follow-up could vendor a minimal in-memory store factory to unlock a node-free CI test

I think this is the biggest deal here, and obviously it's out of scope for this PR, so I've opened https://github.com/zcash/librustzcash/issues/2623 to track that.

Other than that the only thing that jumps out at me is that there's a number of discarded errors and I wonder if it might be good to either roll those into the mapped messages or create variants in the mapped-to error enum? Not a biggy though. 

Let me know when it's ready for another pass and I'll give it my stamp 🙇 .

sellout:
@schell Ok, this should be ready for re-review. I preserved more error context (and used a structured error as much as possible – opened #671 to recommend that approach more consistently). I also added i18n for the new errors and added the new RPC methods to the book.

sellout:
I’m pretty sure the two failures are flaky tests and not related to my changes.

- [Integration tests / Build zallet on Ubuntu 24.04](https://github.com/zcash/integration-tests/actions/runs/29936341555)
- [Integration tests / RPC tests Ubuntu (latest) zaino shard-5 (required)](https://github.com/zcash/integration-tests/actions/runs/29936341555)

I didn’t make any changes related to those failures, and [I’m re-running to verify](https://github.com/zcash/integration-tests/actions/runs/29942995625).

nullcopy:
@sellout can you move this to a branch on the upstream repo, so that integration tests can run?

sellout:
Well, on the second run, both failing tests passed, but a different one failed 😞

sellout:
@nullcopy This one _is_ on the upstream repo, no?

nullcopy:
Oh you're right. For some reason I thought I saw the branch as `sellout:pczt-rpc-methods` :facepalm:

And yeah, seems we have a flaky test :thinking:

nullcopy:
@sellout, take a look at this, which should fix the CI flakiness: https://github.com/zcash/integration-tests/pull/183

I hope to review your PR tomorrow... been super busy with other stuff

zallet-core/src/components/json_rpc/methods/pczt_common.rs
```diff
@@ -0,0 +1,135 @@
...
+pub(super) const PROP_SCOPE: &str = "zallet.v1.scope";
+
+/// Per transparent input: the non-hardened address index (`u32`, little-endian).
+pub(super) const PROP_ADDRESS_INDEX: &str = "zallet.v1.address_index";
+
+/// Encodes a transparent key scope as the `u32` stored in the [`PROP_SCOPE`] hint.
+///
+/// Inverse of [`decode_key_scope`]. Any scope that is neither external nor
+/// internal (e.g. ephemeral) maps to `2`.
+pub(super) fn encode_key_scope(scope: TransparentKeyScope) -> u32 {
```
nullcopy, line 80: Hmm, what do you think about impl `From<u32>` on `TransparentKeyScope` instead of this helper? Or perhaps an explicitely named `to_u32()` method on the type?

zallet-core/src/components/json_rpc/methods/pczt_common.rs
```diff
@@ -0,0 +1,135 @@
...
+    } else if scope == TransparentKeyScope::INTERNAL {
+        1
+    } else {
+        2
+    }
+}
+
+/// Decodes a [`PROP_SCOPE`] `u32` back into a key scope, or `None` if the value
+/// is not one this Zallet version writes.
+pub(super) fn decode_key_scope(value: u32) -> Option<TransparentKeyScope> {
```
nullcopy, line 86: Similar comment as above.

zallet-core/src/components/json_rpc/methods/pczt_common.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Encodes a transparent key scope as the `u32` stored in the [`PROP_SCOPE`] hint.
+///
+/// Inverse of [`decode_key_scope`]. Any scope that is neither external nor
+/// internal (e.g. ephemeral) maps to `2`.
+pub(super) fn encode_key_scope(scope: TransparentKeyScope) -> u32 {
+    if scope == TransparentKeyScope::EXTERNAL {
+        0
+    } else if scope == TransparentKeyScope::INTERNAL {
+        1
+    } else {
```
nullcopy: `TransparentKeyScope` allows custom scope definitions, which would all be caught by this arm and re-mapped to scope 2. That's incorrect.

zallet-core/src/components/json_rpc/methods/pczt_common.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/// Decodes a [`PROP_SCOPE`] `u32` back into a key scope, or `None` if the value
+/// is not one this Zallet version writes.
+pub(super) fn decode_key_scope(value: u32) -> Option<TransparentKeyScope> {
+    match value {
+        0 => Some(TransparentKeyScope::EXTERNAL),
+        1 => Some(TransparentKeyScope::INTERNAL),
+        2 => Some(TransparentKeyScope::EPHEMERAL),
+        _ => None,
```
nullcopy: Similar comment as above, about custom `TransparentKeyScope`s

zallet-core/src/components/json_rpc/methods/pczt_create.rs
```diff
@@ -0,0 +1,222 @@
...
+/// before doing any proposal work.
+const MAX_RECIPIENTS: usize = 1000;
+
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// Result of creating a PCZT.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct CreateResult {
+    /// The base64-encoded PCZT.
+    pub pczt: String,
```
nullcopy, line 61: Can we instead return `{pczt, privacy_policy}` where `privacy_policy` is the minimum required policy computed from the proposal?

Context: I plan to drop #531's `z_proposetransaction`/`z_finalizetransaction` and consolidate on these methods, with the review-first flow being "create with a permissive policy, inspect what it would reveal, then sign". That needs `pczt_create` to report the required policy. Everything else can be added using your pczt methods as-is, but this is the one piece that breaks the published response shape if added later, so I'd like it in now.

zallet-core/src/components/json_rpc/methods/pczt_create.rs
```diff
@@ -0,0 +1,222 @@
...
+            "err-pczt-too-many-recipients",
+            given = amounts.len(),
+            maximum = MAX_RECIPIENTS,
+        )));
+    }
+
+    let request = build_request(&amounts)?;
+
+    // Resolve `from_address` to an account.
+    let account = {
```
nullcopy: This is potentially misleading. Unless I'm mistaken, the user may provide a transparent address, expecting funds to be spent specifically from UTXOs owned by that address, but instead this maps to the account, and may select to spend any funds within the account, including funds _not_ bound to the specified funding address.

Maybe the fix is to check if the `from_address` is transparent? I'm actually not sure

zallet-core/src/components/json_rpc/methods/pczt_extract.rs
```diff
@@ -0,0 +1,77 @@
...
+pub(crate) type ResultType = ExtractResult;
+
+pub(super) const PARAM_PCZT_DESC: &str =
+    "The base64-encoded PCZT to extract a final transaction from.";
+
+/// Extracts a final, network-ready transaction from a completed PCZT.
+///
+/// The PCZT must already have all required proofs and signatures in place;
+/// extraction verifies them and fails otherwise.
+pub(crate) async fn call(pczt_base64: &str) -> Response {
```
nullcopy: Good find from Claude:
> The extracted transaction bypasses all wallet bookkeeping. zcash_client_backend at our pinned rev provides data_api::wallet::extract_and_store_transaction_from_pczt (wallet.rs:2984) — and create_pczt_from_proposal's own rustdoc directs callers to it. It finalizes spends, verifies proofs/signatures (same as here), stores the transaction in the wallet DB, and returns the txid.
>
> With the raw extractor, the wallet never learns about its own transaction until sync sees it in a block: the spent notes are never marked pending-spent, so a second pczt_create on the same account will happily propose the same notes, and the resulting broadcast rejection is exactly the failure mode that crashed the daemon in #531's testing (double-spend → network rejects → sync task treats it as unrecoverable and exits the process).
>
> Suggestion: make pczt_extract wallet-aware and use extract_and_store_transaction_from_pczt when the PCZT carries librustzcash's PROPRIETARY_PROPOSAL_INFO field (i.e. every PCZT this wallet created), keeping the raw path as a fallback for foreign PCZTs that lack it. That means moving the method from the stateless Rpc trait to WalletRpc, which also seems right conceptually — extraction of a wallet's own transaction is a wallet operation.

zallet-core/src/components/json_rpc/methods/pczt_prove.rs
nullcopy, line 1: This implementation needs to be updated for Ironwood

zallet-core/src/components/json_rpc/methods/pczt_sign.rs
nullcopy, line 1: This implementation needs to be updated for Ironwood

book/src/rpc/methods.md
```diff
@@ -67,6 +67,69 @@ REMINDER: It is recommended that you back up your wallet files regularly. If you
...
+#### Arguments
+- `pczt` (string, required) The base64-encoded PCZT to add proofs to.
+
+## `pczt_sign`
+
+*Only available in wallet builds of Zallet.*
+
+Signs a PCZT with the wallet's keys.
+
+Signs every input the wallet holds keys for. Inputs belonging to other
```
nullcopy, line 181: The current methods give no way for the use to verify that the PCZT they are signing matches the one they originally created. e.g. if I create a PCZT that involves other parties, I would like to know that they did not maliciously alter the PCZT before I ask my wallet to sign it. Perhaps a `pczt_inspect` method plus the afore-mentioned privacy-policy would address this?

At minimum, the docs need a warning to the user to inspect what they're about to sign

zallet-core/src/components/json_rpc/methods/pczt_extract.rs
```diff
@@ -0,0 +1,77 @@
...
+            // Fold partial transparent signatures into their `script_sig`s. This
+            // is a no-op when there are no transparent inputs.
+            let pczt = SpendFinalizer::new(pczt)
+                .finalize_spends()
+                .map_err(PcztError::FinalizeSpends)?;
+
+            // Supplying the Sapling verifying keys is required to extract a PCZT
+            // that has a Sapling bundle. The Orchard verifying key is generated
+            // on the fly by the extractor when one is not provided.
+            let (spend_vk, output_vk) = LocalTxProver::bundled().verifying_keys();
```
nullcopy: `LocalTxProver::bundled()` re-parses the bundled Sapling parameters on every call (prove and extract, which only needs the verifying keys). Since you already cache the Orchard proving key in a `OnceLock`, is it possible to cache the `LocalTxProver` the same way?

Consistent with `z_sendmany` today, so fine as a follow-up, but it would be a nice improvement to all three.

zallet-core/src/components/json_rpc/methods/pczt_create.rs
```diff
@@ -0,0 +1,222 @@
...
+    let proposal = propose_and_check(
+        wallet.as_mut(),
+        &params,
+        account.id(),
+        request,
+        privacy_policy,
+        confirmations_policy,
+        // pczt_create only spends shielded funds; the default policy permits every shielded
+        // pool present in the build and no transparent spending.
+        &SpendPolicy::default(),
```
nullcopy: `SpendPolicy::default()` permits shielded pools only, but `from_address` accepts transparent addresses.

Possible fix:  share `spend_policy_for` from z_send_many.rs (and probably move it next to `propose_and_check` in payments.rs)

nullcopy (changes requested):
Reviewed as of 4a11cd9e534d09d0ae8bb5b8cb0df83a275fcf98

zallet-core/src/components/json_rpc/methods/pczt_sign.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    // Sign Orchard actions.
+    let mut orchard_signed = 0;
+    let mut unsigned_orchard = Vec::new();
+    let orchard_ask = orchard::keys::SpendAuthorizingKey::from(usk.orchard());
+    for i in 0..orchard_count {
+        match signer.sign_orchard(i, &orchard_ask) {
+            Ok(()) => orchard_signed += 1,
+            Err(e) => {
+                tracing::debug!("Orchard action {i} left unsigned: {e:?}");
```
nullcopy: From Claude:
> pczt_sign reports pre-signed Orchard dummy actions as unsigned — pczt_sign.rs:210-235. pczt_create uses BundleType::DEFAULT, which pads Orchard spends with dummies, and the pczt IO Finalizer signs those dummies at creation time (confirmed in orchard 0.15's io_finalizer.rs, including its ..._signs_padding_dummy test). pczt_sign then blindly calls sign_orchard on every action; on a dummy this always returns WrongSpendAuthorizingKey, which the loop records in unsigned_orchard. So a plain 1-spend-plus-change send — fully signed and extractable — reports unsigned_orchard: [1], and strict: true errors out. Fix: skip actions whose spend_auth_sig() is already present (publicly readable at the pinned rev), and apply the same already-signed check to Sapling spends and transparent inputs so re-signing a combined PCZT also reports sanely.

book/src/rpc/methods.md
```diff
@@ -67,6 +67,69 @@ REMINDER: It is recommended that you back up your wallet files regularly. If you
...
+- `minconf` (numeric, optional) Minimum confirmations for inputs.
+- `privacy_policy` (string, optional) Privacy policy for the transaction.
+
+## `pczt_extract`
+
+Extracts the final, network-ready transaction from a completed PCZT.
+
+Finalizes the transparent spends and verifies all proofs and signatures
+before returning the hex-encoded transaction. Fails if the PCZT is not
+fully proven and signed.
```
nullcopy: From Claude:
> The "verify before extracting" claim is false for transparent inputs — pczt_extract.rs docs and the book. At the pinned rev, TransactionExtractor::extract batch-verifies Sapling/Orchard proofs, spend-auth sigs, and binding sigs, but the transparent bundle just has its script_sigs moved into the final tx — no script execution, no ECDSA check. SpendFinalizer only checks that a signature is present and the pubkey hash matches. A PCZT carrying a present-but-invalid transparent signature extracts "successfully" and fails only at broadcast — exactly the network deferral the PR body claims to avoid. Either soften the docs, or run script verification post-extraction (the extractor has script_pubkey and value per input, so it's cheap).

zallet-core/src/components/json_rpc/methods/pczt_create.rs
```diff
@@ -0,0 +1,222 @@
...
+    )?;
+
+    // Derivation info used to populate the zallet signing hints below.
+    let derivation = account.source().key_derivation().ok_or_else(|| {
+        LegacyCode::InvalidAddressOrKey.with_message(fl!("err-from-address-no-payment-source"))
```
nullcopy: From Claude:
> pczt_create hard-fails view-only accounts unnecessarily — pczt_create.rs:129-132. Upstream create_pczt_from_proposal deliberately supports accounts without key derivation (it adds ZIP 32 metadata only when available), and watch-only-creates/offline-signs is the flagship workflow this role split exists for. Zallet's derivation is only needed for the zallet.v1.* hints — which only matter to a signer this wallet could never be for a view-only account. Make the hint-writing conditional instead of erroring (and note the current error fires after the proposal work, with a misleading message).

zallet-core/src/components/json_rpc/methods/pczt_sign.rs
```diff
@@ -0,0 +1,246 @@
...
+pub(crate) async fn call(
+    wallet: DbHandle,
+    keystore: KeyStore,
+    pczt_base64: &str,
+    strict: Option<bool>,
+) -> Response {
+    let pczt = decode_pczt_base64(pczt_base64)?;
+
+    // Read the account signing hints from the global proprietary fields.
+    let seed_fp_bytes = pczt
```
nullcopy: From Claude:
> pczt_sign can only sign PCZTs this wallet created, contradicting its docs — pczt_sign.rs:63-72 and the trait doc/book, which promise "inputs belonging to other keys are left unsigned and reported." In reality it hard-errors when zallet.v1.* global hints are missing — i.e., on every foreign PCZT (zcash-devtool, another zallet, hardware flows) and on any multi-party PCZT where this wallet isn't the creator, since the hints name exactly one seed+account. Also, an unknown seed fingerprint maps to LegacyCode::Database (-20), a server-fault code for caller-supplied input. Minimum fix: document the limitation honestly and use a parameter-error code; better: fall back to trying the wallet's accounts when hints are absent.

zallet-core/src/components/json_rpc/methods/pczt_extract.rs
nullcopy: Follow-up from Claude:
> pczt_extract rebuilds the Orchard verifying key on every call — pczt_extract.rs:55-59. Distinct from the LocalTxProver comment already on the PR: when no Orchard VK is supplied, the upstream extractor runs VerifyingKey::build(...) — a full Halo2 keygen, seconds of CPU — per extraction. Cache one in a OnceLock like pczt_prove does for the proving key. Caveat for the fix: the extractor applies the one supplied VK to both the Orchard and Ironwood bundles, so key the cache by circuit version (or only supply it when the Ironwood bundle is empty) to stay compatible with the Ironwood work.

zallet-core/src/components/json_rpc/methods/pczt_prove.rs
nullcopy, line 1: From Claude:
> First pczt_prove will likely exceed the 30 s RPC timeout, and the work is uncancellable — pczt_prove.rs + the server's default request timeout. First call pays Orchard ProvingKey::build + Sapling parameter parse + actual proving. On timeout the response future is dropped but the detached spawn_blocking task runs to completion; a client retry loop stacks more uncancellable proving tasks (tokio allows up to 512 blocking threads). Warm the keys at startup or on first use in a background task, bound concurrent prove/extract with a semaphore, and document the first-call cost.

zallet-core/src/components/json_rpc/payments.rs
```diff
@@ -114,6 +129,207 @@ pub(super) fn build_request(amounts: &[AmountParameter]) -> RpcResult<Transactio
...
+/// when the transaction's net flows are fully transparent, i.e. it has no shielded input or
+/// output at all), but that is its invariant, not ours.
+pub(super) fn transparent_change_policy_for(spend_policy: &SpendPolicy) -> TransparentChangePolicy {
+    match spend_policy.transparent() {
+        Some(_) => TransparentChangePolicy::TransparentChangeAllowed,
+        None => TransparentChangePolicy::ShieldChange,
+    }
+}
+
+pub(super) fn propose_and_check(
```
nullcopy, line 198: The doc comment for this function looks like it got copy-pasta'd into the comment for `transparent_change_policy_for`

zallet-core/src/components/json_rpc/methods/pczt_create.rs
```diff
@@ -0,0 +1,222 @@
...
+    // Resolve `from_address` to an account.
+    let account = {
+        let address = Address::decode(wallet.params(), &from_address).ok_or_else(|| {
+            LegacyCode::InvalidAddressOrKey.with_message(fl!("err-invalid-from-address"))
+        })?;
+
+        get_account_for_address(wallet.as_ref(), &address)
+    }?;
+
+    let privacy_policy = match privacy_policy.as_deref() {
```
nullcopy: From Claude:
> `parse_privacy_policy` is dead code while this PR adds a second inline copy of it — pczt_create.rs:96-104 duplicates z_send_many.rs:151-158 byte-for-byte, next to a #[allow(dead_code)] helper extracted to prevent exactly this. Wire both call sites through it.

nullcopy (changes requested):
I asked Claude to give it a review pass too. Some of the findings were very helpful.

Here's a few more useful general comments from Claude:
- TEX recipients die with a generic "Failed to create PCZT": a ZIP 320 multi-step proposal passes propose_and_check, then create_pczt_from_proposal rejects it (steps().len() > 1 is unsupported upstream). Since z_sendmany supports TEX, pre-check and return a clear InvalidParameter, and document the limitation.
- Wire-visible naming inconsistency: from_address vs z_sendmany's fromaddress (matters for named-parameter calls, permanent once published), and pczt_create's privacy_policy doc lacks the value enumeration and default that z_sendmany's has.
- pczt_combine error-code inconsistency: a decode failure inside the array is re-stamped Deserialization (-22) even when the inner error was InvalidParameter (-8, oversized) — same input to pczt_prove reports differently.
- Malformed-hint logs write the HD address index at warn level (pczt_sign.rs:123-136) — mild privacy signal; the analogous per-input failures already use debug!.
- The decrypted seed binding is held to end-of-function though only needed for from_seed; scoping it tighter is free defense-in-depth. (Derived key non-zeroization matches the existing z_sendmany pattern — inherited, not new.)

nullcopy:
@sellout I went ahead and attempted to address this feedback myself. See https://github.com/zcash/zallet/pull/678 which adds a bunch of fixups you can squash into this PR if you agree. I'm not married to any of my proposed solutions, but it gives me something I can build on until you've had a chance to assess.

I'd especially love to hear your thoughts on the `pczt_inspect` idea. I'm not entirely sold on this approach, but I think it gets the job done :shrug:

sellout:
Thanks, @nullcopy. I’ve pulled in a lot of your PR, and made some other changes/refactorings. Appreciate the work. I’ll have the unified version pushed shortly.

sellout:
Ok, I’ve integrated @nullcopy’s #678, with a few changes

- `pczt_combine` preserves the underlying error code, rather than using `Deserialization` for everything,
- some more l10n
- changed `FundSource::Transparent`’s type to enforce non-emptiness
- some deduplication: 
  - The `zallet.v1.*` hint layout decoded in both `pczt_sign::read_signing_hints` and `pczt_inspect`
  - `enforce_privacy_policy` and `required_privacy_policy` both now derive from one `step_privacy_requirement` classifier
  - `orchard_proving_key` and `orchard_verifying_key` each had three `OnceLocks` that’s now one
  `PerCircuitVersion<T>`
  - `pczt_inspect` summarized the Orchard and Ironwood bundles twice, now one
  `orchard_info`
  - `pczt_extract` serialized the transaction in multiple branches, now one `serialize_transaction`

nullcopy (approved):
utACK ba76d0defe94bc3d2ead3e7aac80b35d451d2d1d

PR #543, opened 2026-07-03
Refactoring phase 1:  Make chain backend dynamically loadable

nuttycom:
This is the first stage of the refactoring to allow the zebrad and zaino backends to drift relative to one another with respect to the version of `zebra_state` that they depend on.


nuttycom (commented):
self-utACK

PR #546, opened 2026-07-04
Detect legacy transparent→shielded spends in z_viewtransaction

### LINKED ISSUE #260: Ensure we migrate the very-legacy transparent OVK derivation from `zcashd` to detect "spend from transparent to external shielded"
https://github.com/zcash/zcash/blob/b65b008a7b334a2f7c2eaae1b028e011f2e21dd1/src/zcash/address/zip32.cpp#L30-L45

sellout:
Very old `zcashd` wallets shielded funds directly from a transparent address to a Sapling address using an outgoing viewing key derived straight from the wallet’s HD seed (`ovkForShieldingFromTaddr` in zcashd’s `zcash/address/zip32.cpp`), predating the ZIP 316 transparent OVK derivation that `z_viewtransaction` already collects. Wallets imported from such `zcashd` installs therefore could not detect their own historical “spend from transparent to external shielded” transactions.

Port that derivation and feed it into the OVK set that `z_viewtransaction` tries against a transaction’s shielded outputs, deriving it for every mnemonic and legacy seed the keystore holds. This needs the raw seed, so it is best-effort: when the wallet is locked the legacy OVKs are skipped and the modern OVKs still apply.

Fixes #260.


nuttycom:
Converted back to draft for conflict resolution.

nuttycom (approved):
utACK 3b91eb5e7c881195fb4d76ffc222c8f8872d1418

Issue #540, opened 2026-07-02
Proposal: split the chain backends into independent resolution graphs (ending zebra-state/zaino feature mutual exclusivity)


nuttycom:
This is a proposal for team discussion — nothing here is decided. It was drafted after revisiting the decision to make the `zebra-state` and `zaino` cargo features mutually exclusive.

## Problem

Zallet has two chain-data backends — `zebra-state` (in-process `ReadStateService` + Zebra JSON-RPC) and `zaino` (embedded `zaino-state` `NodeBackedChainIndex`) — selected by mutually-exclusive cargo features, enforced by a `compile_error!` in `zallet/src/lib.rs`. We would like both backends to be available from a single installation, with the backend chosen by configuration rather than at compile time.

The obstacle is a diamond dependency:

```
zallet ──────────────► zebra-state (direct backend)
zallet ► zaino-state ► zebra-state (embedded Zaino backend)
```

`zebra-state` depends on `librocksdb-sys`, which declares `links = "rocksdb"`. Cargo permits at most one package per `links` value **per resolution graph** — i.e. per `Cargo.lock`. Two consequences follow:

1. Both edges of the diamond must resolve to the *same* `zebra-state` version. If we want a newer `zebra` release and Zaino has not yet updated to it, the workspace cannot resolve at all.
2. Feature flags cannot save us: the conflict fires at lockfile resolution time, before feature activation is considered. Mutual exclusivity of the features only protects against the *second-order* failure mode (a lagging Zaino failing to compile against a newer zebra); it does nothing about the resolution-level handcuff. Today the handcuff is held at bay by pinning both `zebra-*` and `zaino-*` to git revisions that agree on one `zebra-state`, which is exactly the fork-maintenance treadmill we want to escape.

Additionally, `zaino-state`'s public API exposes `zebra-rpc`/`zebra-chain` types, so even absent the `links` constraint, Zallet's Zaino backend must use version-identical `zebra-*` crates with Zaino itself.

### Assumptions (flag if you disagree)

- The Zaino backend **stays embedded**: `zallet + zebrad` must remain a sufficient deployment; we will not require an external `zainod` process.
- Packaging-level dispatch is acceptable: separate per-backend binaries behind a thin launcher satisfy "both backends enabled". A single fat binary is *not* a requirement.

## Proposal

[...]

PR #547, opened 2026-07-05
Refactoring phase 2: split the chain backends into independent resolution graphs

nuttycom:
Part of #540; stacked on #543 (base branch is `split-backend-workspaces` — only the last 7 commits are new here).

This is the core of the #540 proposal: each chain backend now lives in its own cargo workspace with its own lockfile, so the two backends' `zebra-state` versions are fully decoupled and the `links = "rocksdb"` constraint is satisfied per-graph. The `compile_error!` mutual-exclusivity guard and the `zaino`/`zebra-state` features are gone.

## Layout

```
Cargo.toml            workspace: zallet-core only; excludes backends/, crates/
zallet-core/          backend-generic wallet library (all commands, RPC, sync, config)
backends/zebra/       package `zallet` (bin `zallet`, unchanged name) — own workspace/lockfile
backends/zaino/       package `zallet-zaino` — own workspace/lockfile
crates/zebra-read-state/  shared ReadStateService construction, compiled per-graph
```

**Zero user-visible change in this phase:** the zebra backend package keeps the `zallet` name and binary, so deb assets, trycmd expectations, and the release workflow are untouched. The `zallet` launcher (phase 3) will claim the binary name and rename this package.

Backends implement the phase-1 `ChainFactory` capability and register it via the now two-argument `boot()`; the backend binary, not the library, chooses the backend. Backend packages declare dependencies directly (cross-workspace inheritance cannot reach a foreign root's `[workspace.dependencies]`) and carry verbatim copies of the librustzcash patch block, with sync comments.

## Verification

- Root, `backends/zebra`, `backends/zaino`: `cargo test` (features `zcashd-import,rpc-cli,tokio-console`), `cargo clippy --all-targets` (zero warnings), `cargo fmt --check`, `cargo doc` — all green per commit.
- The zebra backend's suite includes the existing acceptance + trycmd tests, unchanged.
- Lockstep: `zcash_client_sqlite` resolves to the identical librustzcash rev (`cac64ba2`) in all three lockfiles (an automated CI check for this lands in phase 4).
- CI (`ci.yml`, `lints.yml`) and the Docker build reshaped to per-workspace `--manifest-path` invocations.

## Reviewer notes

- CLI asset generation (completions/manpages/debian-copyright) stays in zallet-core's `build.rs` (it `include!`s `src/cli.rs`, and the fluent compile-time domain check is tied to the crate name); artifacts still land in each backend's `target/`, so packaging is unaffected.
- `crates/zebra-read-state` is deliberately independent of zallet-core (zebra-typed parameters, own error type) to avoid a dependency cycle.
- `ChainRuntime::run_migrate_zcashd_wallet` carries `#[allow(private_interfaces)]`: backends implement the trait but never call that method — sealed-style.

## Deferred to phase 4 (tracked in #540)

- Lockfile-lockstep CI check (the non-negotiable companion of this design).
- cargo-vet / cargo-deny coverage for the two backend workspaces (currently only the root graph is audited).
- **debian-copyright completeness**: embed-licensing collects zallet-core's graph, so zebra/rocksdb licenses are currently missing from the deb — must be fixed before the next release.
- Release/deb, stagex, `utils/build.sh`, and arm64-nix workflows still reference pre-split paths (release-time only, not PR CI).
- Zaino backend acceptance/trycmd coverage; installation-book update for the removed `--features zaino` instructions.


zallet-core/src/components/chain.rs
```diff
@@ -275,23 +260,28 @@ impl BlockLocator {
...
 }
 
 /// A transaction together with the chain metadata the wallet needs to ingest it.
-pub(crate) struct ChainTx {
-    pub(crate) inner: Transaction,
-    pub(crate) raw: Vec<u8>,
-    pub(crate) block_hash: Option<BlockHash>,
-    pub(crate) mined_height: Option<BlockHeight>,
-    pub(crate) block_time: Option<u32>,
+pub struct ChainTx {
```
nuttycom, line 632: Instead of pub fields, expose accessors and a constructor.

zallet-core/src/components/chain.rs
```diff
@@ -174,16 +157,16 @@ pub(crate) enum UpgradeStatus {
 /// Each [`Chain`] backend converts its own representation into this so that the
 /// consensus-compatibility check ([`check_consensus_compatibility`]) is backend-neutral.
 #[derive(Clone)]
-pub(crate) struct ReportedUpgrade {
+pub struct ReportedUpgrade {
     /// The consensus branch ID the node reports for this upgrade.
-    branch_id: u32,
+    pub branch_id: u32,
```
dannywillems: Nit: are we allowing public fields? I had the feeling we were creating accessors and always leave the fields private.

zallet-core/src/components/chain.rs
```diff
@@ -548,15 +531,17 @@ pub(crate) trait ChainView: Clone + Send + Sync + 'static {
 
 /// A block's height and hash.
 #[derive(Clone, Copy, Debug, PartialEq)]
-pub(crate) struct ChainBlock {
-    pub(crate) height: BlockHeight,
-    pub(crate) hash: BlockHash,
+pub struct ChainBlock {
+    /// The block's height.
+    pub height: BlockHeight,
```
dannywillems: Nit: same comment as above.

zallet-core/src/components/chain.rs
```diff
@@ -588,23 +573,28 @@ impl BlockLocator {
...
 
 /// A transaction together with the chain metadata the wallet needs to ingest it.
-pub(crate) struct ChainTx {
-    pub(crate) inner: Transaction,
-    pub(crate) raw: Vec<u8>,
-    pub(crate) block_hash: Option<BlockHash>,
-    pub(crate) mined_height: Option<BlockHeight>,
-    pub(crate) block_time: Option<u32>,
+pub struct ChainTx {
+    /// The parsed transaction.
```
dannywillems, line 662: Nit: same comment as above.

zallet-core/Cargo.toml
```diff
@@ -0,0 +1,166 @@
...
+authors.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+description = "A prototype wallet. Don't rely on this for privacy or with significant funds yet!"
+repository.workspace = true
+license.workspace = true
+categories.workspace = true
+
+[dependencies]
+abscissa_core.workspace = true
```
dannywillems: Nit: this is ordered alphabetically. Usually, we ordered between Zcash related packages and others.

dannywillems (approved):
utACK f6970c1fd77e526a21cb418823ea9a825068d905.
I rely on the CI for the compilation.

+18060, -2583

Image

zallet-core/src/components/chain.rs
nuttycom: Done. All three seam structs (`ChainTx`, `ChainBlock`, `ReportedUpgrade`) now have private fields with constructors and accessors, folded into the "Move the Zaino backend" commit (the earliest point where every touched path exists in the tree).

`ChainTx::new` deliberately does **not** constrain the three metadata options to move together — the zebra backend's side-chain case legitimately produces `block_hash: Some(..)` with `mined_height: None`; the constructor docs enumerate the three states (best-chain-mined / side-chain / mempool). One ownership-transfer method was needed: `into_parts(self) -> (Transaction, Vec<u8>)`, for the raw-transaction RPC path that moves `inner` out by value rather than cloning.

---
Drafted with the assistance of Claude Code (Claude Fable 5).

zallet-core/src/components/chain.rs
nuttycom: Done — fields are private with a constructor and accessors now (see the reply on the `ChainTx` thread for the details and the one ownership-transfer exception).

zallet-core/src/components/chain.rs
nuttycom: Done — fields are private with a constructor and accessors now (see the reply on the `ChainTx` thread for the details and the one ownership-transfer exception).

zallet-core/src/components/chain.rs
nuttycom: Done — fields are private with a constructor and accessors now (see the reply on the `ChainTx` thread for the details and the one ownership-transfer exception).

zallet-core/Cargo.toml
nuttycom: Done — `[dependencies]` regrouped into the same comment groups the root manifest uses (Zcash-ecosystem crates separated from the rest, alphabetical within groups); specs untouched.

PR #548, opened 2026-07-05
Refactoring phase 3: zallet launcher with config-selected chain backend

nuttycom:
Part of #540; stacked on #547 (base branch is `backend-split-phase2` — only the last 4 commits are new here).

With the backends in their own workspaces (#547), this phase makes backend selection a runtime configuration decision:

- **New top-level `backend = "zebra-state" | "zaino"` config key** (`ChainBackendKind`, kebab-case). Backend binaries validate it at startup and reject a mismatched config by name. Validation applies only to configs actually loaded from a file: abscissa hands `after_config` a default config when none exists, so configless commands (e.g. `zallet-zaino example-config` on a fresh system) must not be rejected.
- **The zebra backend package renames to `zallet-zebra-state`** (matching the binary-name convention #538/#539 established for the deb). The clap display name stays `zallet` for both backends, so usage text and trycmd expectations are unchanged.
- **New `zallet` launcher binary** (root workspace; deps: `toml` + `home` only). It scans argv for the config-locating globals (`--datadir`/`-d`, `--config`/`-c`, space/`=`/attached forms), peeks exactly the `backend` key, locates the sibling backend binary (next to `current_exe()`, then `$PATH`), and `exec()`s it with all arguments passed through (Windows: spawn + exit-code forwarding). No config file ⇒ default backend, mirroring the backends' own tolerance.
- **deb metadata moves to the launcher package**: the deb now ships launcher + `zallet-zebra-state` + `zallet-zaino` as real binaries (the release-CI symlink seeding for `zallet-zebra-state` is obsolete). The release workflow itself is reworked in phase 4; the asset comments document the seeding contract.
- **CI smoke test** (required Linux job): builds launcher + backends, verifies default dispatch, config-driven dispatch to `zallet-zaino`, argument passthrough, the missing-binary error, and a backend binary rejecting a mismatched config.

Verification: tests, `clippy --all-targets` (zero warnings), `fmt`, and `cargo doc` green per commit across all three workspaces; zebra acceptance/trycmd suites unchanged and passing; launcher unit tests cover argv scanning, config peek, and binary-name mapping.


zallet-core/src/config.rs
```diff
@@ -22,6 +22,29 @@ use {
     zip32::fingerprint::SeedFingerprint,
 };
 
+/// The chain backends a Zallet deployment can use.
+///
+/// Serialized in config files as the top-level `backend` key.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum ChainBackendKind {
```
nuttycom: This is a slight layering violation, as it means that `zallet-core` is at least nominally aware of and commits to a specific set of backends. What alternative designs might avoid the need for this enum?

nuttycom: Agreed on the layering violation — and it's worse than nominal: the closed set is currently encoded three times (this enum, the launcher's private `Backend` enum with its own parse/binary-name tables, and implicitly in the binary names). Alternatives considered:

**1. Open string + backend-declared name + binary-name convention (recommended).** The config field becomes a validated newtype (`BackendName`, charset `[a-z0-9-]+`); `ChainFactory` declares `const NAME: &'static str`; `after_config` validation compares the config's string against the registered runtime's name, exactly as today. The launcher drops its enum entirely and dispatches by convention: `backend = "foo"` → exec `zallet-foo` (the charset restriction also closes the path-injection hole that `exec("zallet-{}", name)` would otherwise open). Consequences:
- zallet-core carries no knowledge of which backends exist; a backend is just a name plus a `ChainFactory` impl.
- The *default* backend stops being core's decision too — it moves to the launcher, which is the packaging/deployment layer where that choice actually belongs.
- Out-of-tree backends become possible with zero changes to core or the launcher: implement the trait, ship a `zallet-<name>` binary.
- Cost: a typo'd `backend` value surfaces as a runtime dispatch error ("`zallet-zebrastate` not found beside the launcher or on `$PATH`") rather than a serde parse error enumerating valid variants. The launcher can soften this by listing the `zallet-*` siblings it can see, which makes valid values *discoverable* rather than compiled in. The example-config docs still name the in-tree backends, but as documentation rather than a code commitment.

**2. Relocate the enum to a shared micro-crate** consumed by core and the launcher: deduplicates the three copies but keeps the closed set as a code commitment — doesn't answer the criticism.

**3. Open enum** (`#[serde(untagged)]` known variants + `Other(String)`): keeps friendly serde errors for known values, but core still names the backends — a half-measure.

**4. Parameterize `ZalletConfig` over the backend-kind type**, letting each backend crate instantiate it: removes the commitment but infects every config consumer with a type parameter for the sake of one field — disproportionate.

Happy to rework the PR to (1) — the change is contained to this commit's scope (config newtype, `ChainFactory::NAME`, launcher table → convention + charset guard, fluent messages, example-config docs) and I'd fold it into the owning commit rather than appending.

---
Drafted with the assistance of Claude Code (Claude Fable 5).

nuttycom: Reworked to the open-name design and folded into the owning commits (the enum-era commit messages are rewritten to match):

- `BackendName` (validated string newtype: nonempty, lowercase alphanumeric plus hyphens) replaces `ChainBackendKind` in zallet-core; `ChainFactory` declares `const NAME: &'static str`. The wallet library no longer knows which backends exist.
- The launcher's own duplicate enum is gone too: dispatch is purely `backend = "foo"` → `zallet-foo` (charset-validated, so the name is safe to embed in a filename), and the missing-binary error now lists the `zallet-*` binaries actually installed next to the launcher — valid values are discovered, not compiled in.
- The default backend is now solely the launcher's decision. Consequence: a directly-invoked backend binary accepts a config with no `backend` key (choosing the binary is already an explicit backend choice); previously the non-default binary rejected it. Mismatch rejection for an explicitly-named different backend is unchanged.
- The lint-fix commit was subsumed by the rewritten launcher commit; #549 is rebased on top.

Verified: full test/clippy/fmt matrix across the three workspaces, beta clippy on the launcher, and live dispatch behavior (default, config-driven, discovery listing, mismatch rejection, absent-key acceptance).

---
Drafted with the assistance of Claude Code (Claude Fable 5).

zallet-core/src/config.rs
```diff
@@ -1071,6 +1154,14 @@ impl ZalletConfig {
 
         for field_name in Self::FIELD_NAMES {
             match *field_name {
+                // `backend` is the sole top-level *file-configurable* key; it must be
```
dannywillems, line 1157: LGTM for this design.

zallet/src/main.rs
```diff
@@ -0,0 +1,437 @@
+//! The Zallet launcher.
+//!
+//! Zallet's chain backends are separate binaries (`zallet-zebra`,
+//! `zallet-zaino`), each built in its own cargo workspace so their `zebra-state`
+//! dependency versions can move independently (see zcash/zallet#540). This binary is
+//! the user-facing `zallet` command: it reads the config file's top-level `backend`
+//! key and hands the entire invocation over to the corresponding backend binary.
+//!
+//! The launcher is deliberately dependency-light and performs only best-effort
```
dannywillems, line 9: > The launcher is deliberately dependency-light

I like the idea, but is it expected to not have `clap` (or any equivalent parser)?

dannywillems: We can use `clap` as discussed in person.

nuttycom: Deliberate, for two reasons that reinforce each other:

1. **The launcher must pass through the entire backend CLI surface without knowing it.** A clap parser here would either replicate the full CLI definition from zallet-core (a maintenance twin that drifts), or run in ignore-unknown mode — and clap's unknown-argument handling is not designed for faithful pass-through scanning (e.g. it cannot reliably distinguish `--unknown-flag value` from a positional, which changes how the remainder parses). The launcher needs exactly two globals (`--datadir`/`--config` in their space/`=`/attached forms) and total fidelity for everything else, which a ~40-line scanner delivers exactly.

2. **The scan is best-effort by design, with an authoritative safety net.** If the scanner ever mis-reads an exotic invocation, the worst case is dispatching to the wrong backend binary — which then refuses to run against a config naming a different backend (the `backend`-key validation in this PR). A wrong parse can cost an error message, never a wallet opened by the wrong backend.

Pulling in clap would add its parse semantics as a *second* source of truth about the command line without removing the need for the pass-through scan. If the two-global surface ever grows, that calculus changes and clap-with-`ignore_errors` would be worth revisiting.

---
Drafted with the assistance of Claude Code (Claude Fable 5).

dannywillems (approved):
utACK 5ca74add06438769172e600a06b77755272e1e73

PR #549, opened 2026-07-05
Refactoring phase 4: CI tooling, audits, and release paths for the split workspaces

### LINKED ISSUE #540: Proposal: split the chain backends into independent resolution graphs (ending zebra-state/zaino feature mutual exclusivity)

nuttycom:
Closes #540. Stacked on #548 (base branch is `backend-split-phase3` — only the last 5 commits are new here). Final phase of the backend workspace split.

## Lockstep check (`utils/check-lockstep.sh`)

The non-negotiable companion of the split design: both backend binaries open the same wallet database, so the wallet-critical crates must never diverge between the three lockfiles. The script fails CI if the resolved source/version of any librustzcash patch-block crate (plus `zcash_client_sqlite`, `rusqlite`, zewif, age) differs across the three graphs, or if the four package versions fall out of release lockstep. `zebra-*`/`zaino-*` are deliberately unconstrained — decoupling their cadences is the point of the split. Wired into the required Linux CI job. Current state: `Lockstep OK: 19 crates + 5 package versions consistent across 3 lockfiles.`

## Debian copyright across all shipped binaries

`tools/gen-copyright` unions the locked dependency graphs of the launcher and both backends (release feature set) — previously the file was generated from zallet-core's graph alone, so backend-graph-only licenses (e.g. `ring`, `aws-lc-rs`, `curve25519-dalek`) were missing from the deb. Generation moves out of zallet-core's `build.rs` into this tool, run on the release path. (MIT/Apache-matching crates are skipped, as before.)

## Release paths

`binaries-and-deb-release.yml`, `Dockerfile.stagex`, `utils/build.sh`, the arm64-nix workflow, and `flake.nix` (crane, per-workspace `cargoLock`) reworked for the three-workspace layout: build launcher + both backends, seed `zallet-zebra-state`, `zallet-zaino`, completions/manpages, and the union debian-copyright into root `target/release/` before `cargo deb -p zallet`. The old `zallet-zebra-state` *symlink* seeding is obsolete (real binary now).

**Naming decision for review:** the standalone release artifact `zallet-<ver>-<suffix>` remains the self-sufficient zebra-state backend binary (unchanged behavior for single-file consumers); the launcher ships additively as `-launcher`, while deb/Docker `zallet` is the launcher. Deliberately inconsistent to avoid breaking single-file consumers — flagging for team judgment.

## Audits for every resolution graph

cargo-vet + cargo-deny now run per workspace root (Audits workflow matrix). Audit **facts** are shared (`audits.toml` and `deny.toml` symlinked to the root's canonical copies); **policies and imports are per-graph real files**, since cargo-vet errors on policy entries for crates absent from the current graph. Maintainers can run plain `cargo vet` in any of the three roots, not just CI's `--locked` form. `aggregate-audits.yml` still watches the single canonical `audits.toml`. **Zero new exemptions**: the freshly generated backend lockfiles had resolved past the pre-split-audited union, fixed by precise-pinning five crates back (`quinn-proto`, `quinn-udp`, `fs-err`, `num-bigint`, `rapidhash`), which cascade-pruned the rest.

## Cleanups

Unused root patch entries dropped (`pczt` had zero consumers anywhere; `zcash_history` is backend-only, documented in the sync comments); zaino backend gains an acceptance smoke suite; installation book page rewritten for the launcher/`backend`-key model; CONTRIBUTING documents the workspace layout and the lockstep/divergence policy.

## Phase 5 note

The proposal's phase 5 (unpin zaino) required no code: the split itself decoupled the cadences, and the tree already pins upstream zaino with zebra from crates.io. A zebra bump now touches `backends/zebra` only; a zaino bump touches `backends/zaino` only.

## Follow-ups (not blocking)

- The everyday `Dockerfile`'s share-tree copy lines for debian-copyright are now dead code — run gen-copyright there or drop them.
- Full zebra/zaino nix builds are eval-validated only (multi-hour musl+rocksdb builds); exercise once before the next release.
- `cargo-deny-action`'s `manifest-path` input assumed supported at the pinned version — confirm on this PR's CI run.


.github/workflows/binaries-and-deb-release.yml
```diff
@@ -309,10 +314,15 @@ jobs:
         run: |
           set -euo pipefail
```
dannywillems, line 315: Nit: should we move into files these bash scripts, so we can run tools like `shellcheck` over time?

.github/workflows/binaries-and-deb-release.yml
```diff
@@ -429,23 +441,34 @@ jobs:
```
dannywillems, line 441: Nit: should we move into files these bash scripts, so we can run tools like `shellcheck` over time?

book/src/guide/installation/README.md
```diff
@@ -34,52 +41,58 @@ services/containers over JSON-RPC (for example, the stock `zfnd/zebra` images or
...
-To build or install with the `zaino` backend, pass `--no-default-features --features zaino`
-(along with any other features you need):
+Each backend is its own package in its own cargo workspace, so you install the one you
+want by name (plus the launcher, if you want config-driven dispatch):
 
 ```
-# Install from crates.io with the zaino backend
-cargo install --locked --no-default-features --features zaino,rpc-cli zallet
+# The zebra-state backend (Linux only)
+cargo install --locked --git https://github.com/zcash/zallet.git zallet-zebra
```
dannywillems, line 69: Nit: for a follow-up, it might be interesting to move the scripts into files to ensure that the commands are always correct and can be executed. Executing them in a CI jobs would confirm this.

book/src/guide/installation/README.md
```diff
@@ -34,52 +41,58 @@ services/containers over JSON-RPC (for example, the stock `zfnd/zebra` images or
...
+backend package with `--features` as usual.
 
 ## Pre-compiled binaries
 
 > WARNING: This approach does not have automatic updates.
 
 Executable binaries are available for download on the [GitHub Releases page].
 
-[GitHub Releases page]: https://github.com/zcash/zallet/releases
+[GitHub Releases page]: https://github.com/zcash/wallet/releases
```
dannywillems: Should stay `zallet`?

book/src/guide/installation/README.md
```diff
@@ -112,5 +125,5 @@ If you want to run the latest unpublished changes, then you can instead install
 directly from the main branch of its code repository:
 
 ```
-cargo install --locked --git https://github.com/zcash/zallet.git
+cargo install --locked --git https://github.com/zcash/wallet.git
```
dannywillems: Should stay `zallet`?

tools/gen-copyright/src/main.rs
```diff
@@ -0,0 +1,196 @@
...
+    let root = resolve
+        .root
+        .as_ref()
+        .ok_or("cargo metadata returned no root node")?;
+
+    let mut seen = BTreeSet::new();
+    let mut queue = vec![root];
+    let mut crates = BTreeSet::new();
+    while let Some(id) = queue.pop() {
+        if !seen.insert(id) {
```
dannywillems, line 103: Review note: that's to not process twice the same dep.

utils/check-lockstep.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
```
dannywillems, line 1: Review note: on MacOS, it requires to use a most recent bash version.

```
/opt/homebrew/Cellar/bash/5.3.15/bin/bash ./utils/check-lockstep.sh
```

utils/check-lockstep.sh
```diff
@@ -0,0 +1,117 @@
...
+)
+
+# The lockstep set: the union of [patch.crates-io] package names across the
+# three workspace manifests (honouring `package = "..."` renames), plus crates
+# that touch the shared wallet database but are not patched.
+EXTRA_CRATES=(rusqlite)
+
+patch_crates() {
+  awk '
+    /^\[patch\.crates-io\]/ { inpatch = 1; next }
```
dannywillems, line 38: Review notes: `awk` processes a file line by line. Here, the idea is to make changes to items in the section `[patch.crates.io]`.
Each line has the pattern `Pattern { Action }`, and `awk` executes all of these actions line by line.


https://quickref.me/awk.html for a quick ref.

utils/check-lockstep.sh
```diff
@@ -0,0 +1,117 @@
...
+    /^\[/ { inpatch = 0 }
+    inpatch && /^[A-Za-z0-9_-]+[[:space:]]*=/ {
+      name = $1
+      if (match($0, /package[[:space:]]*=[[:space:]]*"[^"]+"/)) {
+        renamed = substr($0, RSTART, RLENGTH)
+        gsub(/package[[:space:]]*=[[:space:]]*"|"$/, "", renamed)
```
dannywillems, line 44:
```
Substitute t for all occurrences of regex r in string s
```

utils/check-lockstep.sh
```diff
@@ -0,0 +1,117 @@
...
+        name = renamed
+      }
+      print name
+    }
+  ' "$@" | sort -u
```
dannywillems, line 49: Nit: `-u` might not be useful, as we might suppose that there isn't a duplicate already.

Cargo.toml
```diff
@@ -55,7 +55,9 @@ rust-embed = "8"
 # Packaging
 clap_complete = "4"
 clap_mangen = "0.2"
+cargo_metadata = "0.18"
 embed-licensing = "0.3"
+embed-licensing-core = "0.3"
```
dannywillems, line 60: Review note: https://codeberg.org/sbruder/embed-licensing. It doesn't seem to be actively maintained.

Cargo.toml
```diff
@@ -55,7 +55,9 @@ rust-embed = "8"
 # Packaging
 clap_complete = "4"
 clap_mangen = "0.2"
+cargo_metadata = "0.18"
```
dannywillems, line 58: `cargo_metadata` has `0.23.1` as a last release.
https://github.com/oli-obk/cargo_metadata/releases

dannywillems (approved):
utACK b115de6db42b2ccc980f1838f436e141c9fd3e3d.

I quickly reviewed the two new dependencies, focused (and learned at the same time `awk` a bit more).

Left a few nitpick for the future.

.github/workflows/binaries-and-deb-release.yml
dannywillems: Opened https://github.com/zcash/zallet/issues/553

.github/workflows/binaries-and-deb-release.yml
dannywillems: Opened https://github.com/zcash/zallet/issues/553

book/src/guide/installation/README.md
dannywillems: See https://github.com/zcash/zallet/issues/554

Cargo.toml
dannywillems: See https://github.com/zcash/zallet/issues/555

PR #551, opened 2026-07-06
Migrate zcashd wallets via ZeWIF

nuttycom:
Replaces `migrate-zcashd-wallet`'s hand-rolled account/transaction/transparent import logic with the composition **zewif-zcashd export → `zcash_client_sqlite` ZeWIF import** (zcash/librustzcash#2547), making the ZeWIF document the single interchange point between wallet.dat parsing and wallet-database population. Net −577 lines.

## Pipeline

1. `wallet.dat` dump and parse (unchanged), network reconciliation, and the multi-wallet-import pre-check — now sourced from the document's secret store fingerprints.
2. `zewif_zcashd::migrate_to_zewif` produces the interchange document.
3. Zallet enriches the document with what only it knows: chain-derived account birthdays written in as full tree states (Sapling, Orchard, and Ironwood frontiers) with `recover_until` at the chain tip; under `--no-scan`, the expiry-based birthday estimate. Pre-4.7.0 legacy seeds are normalized to their zcashd-style derived mnemonic, with the legacy account's key source re-pointed at the mnemonic fingerprint.
4. Secret material is decoded from its canonical ZeWIF encodings (BIP-39 phrase, WIF, Sapling Bech32), verified against the recorded public halves, and age-encrypted into the keystore via a new `SecretSink` adapter — before any account is created. Because the keystore and wallet DB share a database-wide lock, zallet drives the sink itself and hands `import_wallet` a `DiscardSecrets`; keystore inserts are idempotent, so the pattern is safe.
5. `zcash_client_sqlite::zewif::import_wallet` populates the wallet, and the returned report is logged through the usual warning conventions — including the address-book entry count, which the previous implementation dropped silently.
6. Watch-only (pubkey-only) transparent addresses are supplemented post-import (see gaps below).

## Behavior changes

- Regtest wallets are now rejected up front (activation-schedule equivalence cannot be verified).
- Under `--no-scan`, address exposure derives from the document's recorded addresses rather than transaction-output scraping.
- The address book size is now reported instead of silently discarded.
- `--buffer-wallet-transactions` retains its semantics: when unset, the document's transaction set is stripped before import.

## Dependencies

The librustzcash `[patch.crates-io]` pins move to the head of zcash/librustzcash#2547 (transiently on a fork; will be re-pointed when that merges), and zewif/zewif-zcashd to their current mains. This is a stacked draft until #2547 lands and zewif has a crates.io release. The pin bump also brings in the Ironwood scan-model work, so the chain backends gained Ironwood tree handling here.

## Upstream gaps observed (to be filed separately)

- **zaino** (pinned branch): `get_treestate` exposes no Ironwood tree, so the zaino backend errors on tree-state requests at or above NU6.3 activation. This needs a zaino fix before NU6.3 activates.
- *zewif-zcashd*: the synthesized legacy account's key source records the raw legacy-seed fingerprint where post-4.7.0 zcashd semantics imply the derived mnemonic's fingerprint; zallet normalizes, but the exporter should arguably emit the mnemonic fingerprint itself.
- *zcash_client_sqlite* ZeWIF import: no registration path for watch-only transparent addresses even though the document carries the pubkeys.
- *zewif*: no mutable access to wallets/accounts post-assembly, forcing whole-document rebuilds for enrichment.

## Testing

Eight new unit tests cover fingerprint decoding, network reconciliation (including regtest rejection), frontier conversion, and document enrichment (legacy-derivation normalization, birthday injection, transaction stripping); both backend matrices (zaino and zebra-state) build, test, and lint clean. An end-to-end migration of a real testnet wallet.dat is planned as manual validation before this leaves draft.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,7 +37,7 @@ dependencies = [
  "ident_case",
  "proc-macro2",
  "quote",
- "syn 2.0.118",
+ "syn",
  "synstructure",
 ]
 
@@ -243,18 +243,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "argon2"
-version = "0.5.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
-dependencies = [
- "base64ct",
- "blake2",
- "cpufeatures",
- "password-hash",
-]
-
-[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -274,7 +262,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.118",
+ "syn",
 ]
 
 [[package]]
@@ -306,7 +294,7 @@ checksum = "a273e731a7fb8c2268fd2932c9e9a3469f4fd171d85d070d2687190229653bd3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.118",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


nuttycom:
Closing in favor of #571

PR #557, opened 2026-07-07
Add utils/sync-librustzcash.sh to move the librustzcash pin in lockstep

dannywillems:
## Summary

Adds `utils/sync-librustzcash.sh`, the missing companion to
`utils/check-lockstep.sh`: a one-command way to move the librustzcash
`[patch.crates-io]` pin across all three workspaces (root, `backends/zebra`,
`backends/zaino`) so their lockfiles resolve identically.

## Why

All three binaries open the same wallet database, so the librustzcash stack
(notably `zcash_client_sqlite`) must resolve to one identical rev in every
workspace; `check-lockstep.sh` enforces that in CI. Its only remedy on drift
was a manual "edit each manifest and run cargo update per workspace", which is
error-prone. This script automates that.

## What it does

- `utils/sync-librustzcash.sh <rev>` rewrites the librustzcash rev in all three
  manifests, reconciles each lockfile with `cargo metadata`, and self-verifies
  with `check-lockstep.sh`.
- With no argument it reuses the rev already pinned in the root `Cargo.toml`
  (idempotent re-sync).
- Only the librustzcash pins are touched; the `age`/`zewif`/`zaino` pins are
  left alone.
- Uses `cargo metadata` rather than `cargo update -p <name>`, which is
  ambiguous when a crate such as `zcash_transparent` resolves to two versions
  in the graph.

## CI

A new step in the required job runs the script with no argument and asserts
`git diff --exit-code`, so an in-lockstep tree stays a no-op and any lockfile
drift is caught.

Verified locally: idempotent (no lockfile churn), `check-lockstep.sh` passes,
shellcheck-clean.

## Notes

Each patch block gains a comment naming the script as the only supported way to
move the rev.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,14 @@ jobs:
       # but crates touching the shared wallet database must stay in lockstep.
       - name: Check wallet-critical dependency lockstep
         run: ./utils/check-lockstep.sh
+      # Exercise the sync helper end to end: with no argument it re-pins every
+      # workspace to the rev already in the root Cargo.toml and regenerates the
+      # lockfiles, so on an in-lockstep tree it must be a no-op. A non-empty diff
+      # means the committed lockfiles drifted from what the manifests resolve to.
+      - name: Verify librustzcash sync script is idempotent
+        run: |
+          ./utils/sync-librustzcash.sh
+          git diff --exit-code
       - name: Run tests (zallet-core)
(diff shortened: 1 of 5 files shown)


nullcopy (approved):
utACK a414cc98a7d7ac318567f07d032a3a9a9a25a678

PR #560, opened 2026-07-08
Retry on BlockConflict instead of crashing the whole wallet

### LINKED ISSUE #556: steady_state sync crashes the whole process on BlockConflict, instead of using existing reorg-rewind recovery
## Summary

`zallet`'s steady-state wallet sync crashes the entire process on a `BlockConflict` error from `zcash_client_sqlite`, instead of recovering via the reorg-rewind logic that already exists in `steady_state_iteration` for other cases. This happened on testnet during ordinary near-tip syncing (no unusual user action beyond a normal `start`).

## Environment

- `zallet` @ `main` (`e09f8939`, merge of #549 "backend-split-phase4", includes #548 "backend-split-phase3")
- Backend: `zebra` (`zallet-zebra`, release build)
- `zebrad` @ `main` (`d9d29b658`), built with `--features indexer`, release
- Network: **testnet**
- rustc 1.88.0, macOS arm64
- `wallet.db` predates this session (previously synced under the older single-binary `zallet`)

## Reproduction context

Hit while validating #548/#549 locally: an old single-binary `zallet` was stopped, `zebrad` was updated to `main` (a substantial reorg/non-finalized-state rewrite + Ironwood migration) and resynced to tip, then the new launcher+backend `zallet` was rebuilt and started against the same `wallet.db`. It scanned forward from ~4130955, reached `steady_state` near the tip, then crashed. Both daemons were restarted a few times near the tip across the zebra version swap — flagging as a possible contributing factor, though testnet reorgs regardless of that.

## What happened

```
... Scanning block 4150382 ...
... Scanning block 4150383 ...
Wallet steady-state sync task exited wallet_sync_result=Err(... source: Some(Other(BlockConflict(BlockHeight(4150383)))) ...)
Exiting Zallet because an ongoing task exited; asking other tasks to stop
Shutting down Zallet
```

Full process exit, requires manual restart.

## Root cause

- `SqliteClientError::BlockConflict` (`zcash_client_sqlite-0.20.0/wallet.rs:4073`, in `put_block`) fires when writing a block whose hash mismatches what's stored at that height (reorg).
- `steady_state_iteration` (`zallet-core/src/components/sync.rs`) does handle reorgs — detects fork point, calls `truncate_to_height`, resumes — but only when `is_retryable()` returns true, which only matches `SyncError::Chain(ChainError::Unavailable(_))`.
- `BlockConflict` arrives as `SyncError::Other(Box<SqliteClientError>)`, uncovered by `is_retryable`, so it's fatal — kills `steady_state`, which `commands/start.rs` treats as grounds to shut down the whole process.

So any reorg that manifests this way — even a routine one — currently crashes the wallet instead of self-healing.

## Suggested direction (flagging for discussion, not a PR)

Consider whether `is_retryable` should also cover `SyncError::Other(SqliteClientError::BlockConflict(_))`, routing it through the same rewind-and-resume path as `ChainError::Unavailable`.

zcash-harry:
## Summary

- `BlockConflict` (the backend's best chain reorged away a block the wallet had already stored) wasn't recognized by `is_retryable`, so it propagated as a fatal error and crashed the whole wallet — requiring a manual restart to recover — even though reorg-recovery machinery to handle exactly this case (fork-point search + truncate + resume) already exists for other error paths.
- Rather than assuming the reorg is exactly one block deep, the fix treats the wallet's last known-good position (`height - 1`) as a fresh "previous tip" and runs it through the *existing* fork-point search (`locate_fork_point`/`step_back_to_best_chain`), which walks back as far as it actually needs to. `BlockConflict` only tells us the block at that height is wrong, not how deep the reorg goes, and `put_block` only checks for a collision at the exact height it's writing — not that its parent still matches what's stored one below it — so a fixed depth-1 rewind isn't safe in general.
- Also fixes the same class of bug in `initialize()`'s startup catch-up scan, which had no retry at all for the stale-view error (`ChainError::Unavailable`) that `steady_state`'s main loop already tolerates — so a reorg in progress right as the wallet started would also crash it.

Closes #556

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p zallet-core --all-targets -- -D warnings`
- [x] `cargo test -p zallet-core` — 172 passed
- [x] `cargo build` for all three chain backends (`zebra`, `zaino`, and the launcher)
- [x] Manually reproduced `BlockConflict` against a live testnet sync by injecting a synthetic conflicting block hash a few blocks ahead of the wallet's synced tip, and confirmed:
  - Before this fix: fatal crash (`Failed to synchronize zallet: ...`), process exits, requires manual restart.
  - After this fix: `WARN steady_state: Block at height N conflicts with previously stored data (likely a reorg); searching for the fork point and retrying` — wallet rewinds, resumes, and syncs cleanly to the chain tip afterward, no crash, no manual intervention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 During the alpha period, no Semantic Versioning is followed; all releases should
 be considered breaking changes.
 
+## [Unreleased]
+
+### Fixed
+
+- `steady_state` sync no longer crashes the whole wallet when the backend's best
+  chain reorgs away a block the wallet had already stored (`BlockConflict`).
+  Previously this error wasn't recognized as retryable, so it propagated as a
+  fatal error and required a full restart to recover. It's now handled the same
+  way as other reorgs: the wallet's last known-good position is run through the
+  existing fork-point search (which walks back as far as it actually needs to,
+  rather than assuming the reorg is exactly one block deep), then rewinds and
+  resumes syncing automatically.
+- `initialize()`'s startup catch-up scan could also crash the wallet on a
+  transient stale-view error (e.g. a reorg in progress right as the wallet
+  started) that `steady_state`'s main loop already tolerates. It now retries the
+  same way.
+
 ## [0.1.0-alpha.4] - 2026-06-25
 
 ### Added
--- a/zallet-core/src/components/sync.rs
+++ b/zallet-core/src/components/sync.rs
@@ -59,6 +59,7 @@ use zcash_client_backend::{
     sync::decryptor,
 };
 use zcash_client_sqlite::AccountUuid;
+use zcash_client_sqlite::error::SqliteClientError;
 use zcash_primitives::block::BlockHash;
 #[cfg(not(feature = "spend-index"))]
 use zcash_protocol::TxId;
@@ -294,7 +295,7 @@ async fn initialize<C: Chain>(
             }
         };
 
-        if steps::scan_blocks(
+        match steps::scan_blocks(
             chain_view,
             db_data,
             params,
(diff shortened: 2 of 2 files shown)


dannywillems:
As discussed in person, it would be nice to have tests to confirm the different paths are reached correctly in the scenarii this patch describes. It would also improve our regression tests coverage.

nuttycom (approved):

nuttycom:
> As discussed in person, it would be nice to have tests to confirm the different paths are reached correctly in the scenarii this patch describes. It would also improve our regression tests coverage.

Opened https://github.com/zcash/zallet/issues/564

PR #561, opened 2026-07-08
Integrate Ironwood (NU6.3) wallet support: treestate, RPC surface, and sync plumbing

dannywillems:
Integrate Ironwood (ZIP 2005, NU6.3) wallet support into Zallet end to end: the
node-facing treestate plumbing, the JSON-RPC surface, and the sync path, on top
of the librustzcash Ironwood wallet APIs.

## Dependency pins

- **librustzcash → `main` (`b43a4237`)**, advanced from the NU6.3 base through
  the #2539 Ironwood wallet-support work in the root, zebra, and zaino
  workspaces. Main librustzcash is built against `zcash/orchard` `main` (which
  adds the public `Builder::bundle_type()` accessor), so its
  `[patch.crates-io] orchard = { git = "zcash/orchard", rev = "475ef0ff" }` is
  replicated across all three manifests; without it the pinned rev does not
  compile against the crates.io `0.15.0-pre.2` release.
- **zaino → `dev` (`8048bf8d`)**, which surfaces the Ironwood note commitment
  treestate from `get_treestate` (now a `(sapling, orchard, ironwood)` triple of
  `PoolTreestate`). This required adapting the zaino backend to two accompanying
  `dev` API changes: `ValidatorConnector::State` gained `chain_tip_change` and
  `sync_task_handle`. The chain-tip-change watcher is threaded out of
  `init_read_state_service` (it was already produced by the zebra read-state
  init and discarded); `sync_task_handle` is left `None`, since the wallet keeps
  ownership of the syncer via its existing `AbortOnDrop` guard.

## librustzcash API adaptation (base commits)

- `WalletRead::get_ironwood_nullifiers` is now required (called on the scan
  path); forward it to `WalletDb` in `DbConnection`.
- Forward the `WalletCommitmentTrees` Ironwood methods (`with_ironwood_tree_mut`,
  `put_ironwood_subtree_roots`) to `WalletDb` rather than fall back to the
  trait's no-op defaults, which would silently drop Ironwood tree updates.
- `propose_transfer` now takes an explicit `SpendPolicy` and an
  `Option<TxVersion>`: pass `SpendPolicy::default()` and `None`.
- `Note::protocol()` became `Note::pool()`, classifying Orchard and Ironwood by
  value pool; update the call sites in `z_send_many` and the privacy-policy
  check in `payments`.
- `BlockMetadata::from_parts` and `ChainState::new` gained an Ironwood tree
  argument.

## RPC surface

Surface the Ironwood pool across the JSON-RPC methods (previously they elided
Ironwood or reported only transparent/sapling/orchard):

- `z_getbalanceforaccount`: add an `ironwood` pool.
- `z_gettotalbalance`: count Ironwood funds in `private`.
- `getbalances`: add a per-account `ironwood` balance.
- `z_listunspent`: list Ironwood notes with `pool == "ironwood"`.
- `listtransactions`: stop eliding Ironwood outputs; report their pool.

(`z_viewtransaction` is intentionally left for a follow-up: surfacing Ironwood
spends/outputs there needs trial-decryption under the Ironwood (version 3) note
domain, not the Orchard (version 2) one.)

## Sync path

- **Read the real Ironwood treestate** in the zaino backend's
  `tree_state_as_of` instead of hardcoding an empty tree. Once NU6.3 is active
  and Ironwood notes exist, an empty checkpoint frontier conflicts with the
  wallet's advancing Ironwood tree and `put_blocks` fails with a shardtree
  `CheckpointConflict`, crashing sync. The zebra backend still supplies an empty
  Ironwood frontier (its read-state view does not expose one yet).
- **Fetch and store Ironwood subtree roots** during sync (`update_subtree_roots`
  previously handled only Sapling and Orchard): add `get_ironwood_subtree_roots`
  to the `Chain` trait (implemented against zaino's `get_subtree_roots` for the
  Ironwood pool) and `put_ironwood_subtree_roots` into the wallet DB.

## Status

Ironwood works end to end with this PR: shielding coinbase into an Orchard UA
once NU6.3 is active mints version-3 (Ironwood) notes that are scanned,
positioned in the tree, counted as `private`, reported under the `ironwood`
pool, spendable, and spendable-out (a spend consumes an Ironwood note and its
value returns as new, spendable Ironwood notes). The `CheckpointConflict` sync
crash is fixed.

The `zcash/integration-tests` `wallet_ironwood.py` test exercises this full
receive-then-spend flow and passes. One prerequisite lives on the test side, not
here: the wallet database must be **created** with the NU6.3 activation height
already configured, because the librustzcash Ironwood migration bakes that
height into the `v_ironwood_shard_scan_ranges` SQL view; if NU6.3 is only
configured after the database exists, the view bakes a NULL height and Ironwood
notes never become spendable. On mainnet/testnet the NU6.3 height is a fixed
consensus parameter, so this only needs care in regtest.

### DIFF
--- a/backends/zaino/Cargo.toml
+++ b/backends/zaino/Cargo.toml
@@ -79,18 +79,18 @@ tempfile = "3"
 #
 # Update the librustzcash rev ONLY via utils/sync-librustzcash.sh (which syncs
 # all three manifests and lockfiles at once); do not edit these revs by hand.
-equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "52a19f0ec046b622fbd618b6434e0bdd4d896117" }
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
 
 # TEMPORARY: explicit checkpoint retention (ensure_retained / retained_checkpoints).
 # The librustzcash rev above requires these shardtree/incrementalmerkletree APIs
--- a/backends/zaino/Cargo.toml
+++ b/backends/zaino/Cargo.toml
@@ -79,18 +79,18 @@ tempfile = "3"
 #
 # Update the librustzcash rev ONLY via utils/sync-librustzcash.sh (which syncs
 # all three manifests and lockfiles at once); do not edit these revs by hand.
-equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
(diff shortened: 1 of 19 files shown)


backends/zebra/src/chain.rs
```diff
@@ -755,4 +766,25 @@ mod tests {
         // Above the captured tip.
         assert_eq!(view.resolve(BlockHeight::from_u32(11)).await.unwrap(), None);
     }
+
+    #[tokio::test]
+    async fn tree_state_as_of_supplies_an_empty_ironwood_tree() {
+        // The mock reader returns no tree bytes, so for a finalized height every pool's
+        // final tree is empty. This pins the Ironwood frontier that `ChainState::new` now
+        // requires: Zebra exposes no Ironwood treestate, so the zebra chain view must supply
+        // an empty one, matching the empty Sapling and Orchard trees.
```
nuttycom, line 787: This is a gap if it's still true of Zebra today - need to check.

zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -222,8 +222,13 @@ pub(crate) async fn call<C: Chain>(
         request,
         confirmations_policy,
         // Zallet does not yet spend transparent funds in a general transfer; `ANY_TADDR`
-        // spending is rejected above. This preserves the prior shielded-only behavior.
-        &TransparentSpendPolicy::ShieldedOnly,
+        // spending is rejected above. The default `SpendPolicy` permits every shielded pool
+        // present in the build and no transparent spending, preserving the prior
+        // shielded-only behavior.
+        &SpendPolicy::default(),
```
nuttycom, line 228: This will change once we update to current librustzcash `main` which now has transparent spend capability.

zallet-core/src/components/json_rpc/methods/list_unspent.rs
```diff
@@ -382,6 +386,62 @@ pub(crate) fn call(
...
+            }
+
+            let wallet_internal = note.spending_key_scope() == Scope::Internal;
+
+            let (memo, memo_str) =
+                get_memo(*note.txid(), ShieldedPool::Ironwood, note.output_index())?;
+
+            unspent_outputs.push(UnspentOutput {
+                txid: note.txid().to_string(),
+                pool: "ironwood".into(),
```
dannywillems, line 427: Nit: this follows orchard, but i guess we could have something using `ShieldedPool`

dannywillems:
Making it ready for review, as the changes seem to look legit. We can move step by step with the other patches if they need librustzcash updated.

backends/zebra/src/chain.rs
```diff
@@ -195,6 +197,16 @@ impl Chain for ZebraChain {
         self.reader().orchard_subtree_roots().await
     }
 
+    async fn get_ironwood_subtree_roots(
+        &self,
+    ) -> Result<Vec<CommitmentTreeRoot<orchard::tree::MerkleHashOrchard>>, ChainError> {
+        // TODO: The zebra read-state reader does not expose Ironwood subtree roots yet,
```
dannywillems, line 203: ^

backends/zaino/Cargo.toml
```diff
@@ -102,16 +102,24 @@ zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2
 shardtree = { git = "https://github.com/zcash/incrementalmerkletree", rev = "decefc469dbf1e686b20b7e7dcaa7cb4f122125b" }
 incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree", rev = "decefc469dbf1e686b20b7e7dcaa7cb4f122125b" }
 
+# Replicated from librustzcash b43a4237's own [patch.crates-io]: main librustzcash is
+# built against zcash/orchard main (adds the public `Builder::bundle_type()`
+# accessor), which the crates.io 0.15.0-pre.2 release predates. Required for the
+# librustzcash rev above to compile.
+orchard = { git = "https://github.com/zcash/orchard", rev = "475ef0ff77d45aebff93cb039d639250d82518a3" }
```
nullcopy, line 109: Can this point to main now that Orchard has released?

backends/zaino/Cargo.toml
```diff
@@ -102,16 +102,24 @@ zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2
...
-zaino-fetch = { git = "https://github.com/zingolabs/zaino.git", rev = "e7c9dfdb891d45c18ba53d75f5a54a90d3ec627c" }
-zaino-state = { git = "https://github.com/zingolabs/zaino.git", rev = "e7c9dfdb891d45c18ba53d75f5a54a90d3ec627c" }
+# TEMPORARY: pin the zaino crates to zingolabs/zaino `dev` (currently 8048bf8d),
+# which surfaces the Ironwood (NU6.3) note commitment treestate from
+# `get_treestate` (now a `(sapling, orchard, ironwood)` triple of
+# `PoolTreestate`). Reading the real Ironwood frontier is required for Ironwood
+# notes to be spendable; an empty frontier conflicts with the wallet's advancing
+# tree and sync fails with `CheckpointConflict`. Repoint at upstream zaino once
+# the NU6.3 work lands in a release.
+zaino-common = { git = "https://github.com/zingolabs/zaino.git", rev = "8048bf8d" }
```
nullcopy: Can we use full length commit refs?

backends/zaino/Cargo.toml
dannywillems: We're still using the branch `feat/ironwood`: https://github.com/zcash/orchard/tree/feat/ironwood. The commit is correct.

backends/zaino/Cargo.toml
dannywillems: Done.

backends/zebra/Cargo.toml
```diff
@@ -83,18 +83,18 @@ tokio-console = ["zallet-core/tokio-console"]
...
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
-zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "ea2992595d944d2fddc0d38ef24e242be7e037ba" }
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "b43a42372d8e43e690fe336c431266de17679a8f" }
```
nullcopy, line 86: These are not the latest commit on main... perhaps worth updating?

zallet-core/src/components/json_rpc/methods/z_get_balance_for_account.rs
```diff
@@ -193,7 +215,7 @@ mod tests {
     #[test]
     fn multiple_pools_are_rendered() {
         assert_eq!(
-            rendered(0, 5 * COIN, 625 * COIN / 100, None),
+            rendered(0, 5 * COIN, 625 * COIN / 100, 0, None),
```
nullcopy, line 218: Perhaps Ironwood should be non-zero here

zallet-core/src/components/json_rpc/methods/z_get_balance_for_account.rs
```diff
@@ -193,7 +215,7 @@ mod tests {
     #[test]
     fn multiple_pools_are_rendered() {
```
dannywillems, line 216: Implement a proptest for this.

zallet-core/src/components/json_rpc/methods/z_get_balance_for_account.rs
```diff
@@ -175,15 +193,19 @@ mod tests {
     #[test]
```
dannywillems, line 193: Use proptest for this.

nullcopy (commented):
[ignore this]

nullcopy (approved):
I think only one non-blocking comment in a test. Otherwise LGTM.

utACK f40a783dec02f0e1369e2d79482075d2033f0823

backends/zebra/src/chain.rs
nuttycom: @dannywillems are we pointing at a recent enough zebra?

PR #566, opened 2026-07-08
fix: 5 bugs blocking the arm64 (Nix) release path + gen-copyright's StageX build

### LINKED ISSUE #565: 5 bugs blocking the arm64 (Nix) release path + one StageX Dockerfile step, found in a full release dry-run
## Summary

Since the 3-binary workspace split (#540, merged via #543/#547/#548/#549), no release has actually been cut — so the arm64 (Nix) release path and one step of the amd64 (StageX) path were never exercised end-to-end. While testing a full release dry-run in a private fork, I found and fixed 5 real, reproducible bugs blocking `nix build .#zallet-zaino` / `.#zallet-zebra` / `.#zallet-launcher` and one step of `Dockerfile.stagex`. All are fixed and verified working on an actual `aarch64-linux` machine and in a full CI run (Docker + .deb + standalone binaries, all attested with SLSA provenance).

## Bug 1 — `Dockerfile.stagex`: `gen-copyright`'s bare `cargo run` fails on this builder

```
error: cannot produce proc-macro for `serde_derive v1.0.228` as the target
x86_64-unknown-linux-musl does not support these crate types
```

This stagex builder's rustc only has the musl target installed (no separate glibc host triple). A bare `cargo run -p gen-copyright` (no `--target`) still implicitly resolves to `x86_64-unknown-linux-musl`, but via a different cargo codepath than an explicit `--target` invocation — and that codepath fails building `serde_derive`'s proc-macro crate-type for that target, even though the three `cargo install --target ${TARGET_ARCH}` calls earlier in the same `RUN` build `serde_derive`-dependent crates for the identical target successfully.

**Fix:** pass `--target ${TARGET_ARCH}` explicitly to the `gen-copyright` invocation too, matching the pattern that already works for the three binary installs above it.

## Bugs 2–5 — `flake.nix`: four compounding crane/nixpkgs integration issues

All stem from `backends/zebra` and `backends/zaino` each being their own cargo workspace (excluded from the root `Cargo.toml`, by design, per #540) while `flake.nix` builds all three binaries from one shared `src`.

**Bug 2 — source filter drops backend lockfiles.** `craneLib.filterCargoSources` only recognises `Cargo.toml`/`Cargo.lock` files belonging to the workspace it can see from `.` — it silently drops `backends/*/Cargo.lock` from the filtered `$src` (`Cargo.toml` and `src/` survive because they're matched generically; `Cargo.lock` does not, because crane specifically special-cases lockfiles per-workspace).

**Bug 3 — crane's dummy-source generator only writes the lockfile at the tree root.** `mkDummySrc` (used internally by `buildPackage` for the deps-only/cacheable derivation) unconditionally copies the `cargoLock` argument to `$out/Cargo.lock` at the dummy tree's root — it has no concept of a lockfile living in a workspace subdirectory. Since `backends/zaino` is its own `[workspace]`, Cargo looks for the lockfile next to `--manifest-path backends/zaino/Cargo.toml`, not at the tree root, so it never sees a lockfile there and can't resolve `backends/zaino`'s git-patched `age` dependency:

```
error: failed to load source for dependency `age`
Caused by:
  Unable to update https://github.com/str4d/rage.git?rev=...
Caused by:
  the source ... requires a lock file to be present first before it can be
  used against vendored source code
```

**Bug 4 — `$CARGO_TARGET_DIR` mismatch once `--manifest-path` is involved.** With `--manifest-path backends/zaino/Cargo.toml`, Cargo resolves the default target dir relative to *that* workspace's root, not the build's `$PWD` — but crane's `installCargoArtifactsHook` looks for `$CARGO_TARGET_DIR` (defaulting to plain `target`) at `$PWD`:

```
line 13: cd: target: No such file or directory
```

**Bug 5 — `installFromCargoBuildLogHook`'s `cargo metadata` call resolves dev-dependencies that were never vendored.** This hook registers a *postBuild* hook (overriding `installPhaseCommand` does **not** skip it) that runs `cargo metadata --format-version 1` with no `--manifest-path`, resolving against the tree-root workspace (`zallet-core`, `zallet`, `tools/gen-copyright`) instead of the backend's. `cargo metadata` resolves dev-dependencies too, and `zallet-core`'s dev-deps (`proptest`, `trycmd`, ...) were never vendored for `backends/zaino`'s or `backends/zebra`'s independent lockfile/vendor dir (decoupling those graphs is the whole point of #540) — so that metadata call fails:

```
error: no matching package found
searched package name: `proptest`
```

...which corrupts the hook's downstream `jq` invocation with an unrelated, confusing error that hides the real failure above it:

```
jq: error: syntax error, unexpected '|' at <top-level>, line 4, column 7
jq: 1 compile error
```

## Fixes (all verified)

- Bug 1: add `--target ${TARGET_ARCH}` to the `gen-copyright` invocation in `Dockerfile.stagex`.
- Bug 2: explicitly whitelist `backends/*/Cargo.{toml,lock}` in `flake.nix`'s source filter regardless of what `filterCargoSources` decides.
- Bug 3: a `postPatch` that copies the (already-replaced) root `Cargo.lock` into the workspace subdirectory Cargo actually reads from, for each backend.
- Bug 4: pin `CARGO_TARGET_DIR` explicitly, scoped to only the backend workspaces (pinning it unconditionally breaks the launcher's own — currently working — install path, since `manifestDir == null` doesn't have this problem).
- Bug 5: use the hook's own documented escape hatch, `doNotPostBuildInstallCargoBinaries`, and install the already-built binary ourselves via `installPhaseCommand` for the two backends. That key must be *omitted entirely* (not set to `""`) for the launcher, so crane's own `args.installPhaseCommand or <default>` fallback still applies there.

## Verification performed

- Reproduced all 5 bugs on a scratch `aarch64-linux` EC2 instance (bare `nix build`, outside CI, with `--show-trace -L --keep-failed` to inspect the actual sandbox contents at failure time — confirmed e.g. that `backends/zaino/Cargo.lock` was genuinely absent from the filtered `$src`, and that the root `Cargo.lock` had been silently overwritten with the zaino lockfile's content).
- After the fixes: `nix build .#zallet-zebra .#zallet-zaino .#zallet-launcher` all succeed; the resulting binaries are correctly statically-linked `aarch64-unknown-linux-musl` ELF executables; placing all three side by side and running the launcher correctly dispatches to `zallet-zebra` by default (and to `zallet-zaino` when `zallet.toml` sets `backend = "zaino"`).
- Ran a **full release dry-run end-to-end** in a private fork (tag push → `release.yml`): StageX amd64 container build (with the gen-copyright fix) → arm64 Nix build (with the 4 flake.nix fixes) → multi-arch manifest → `.deb` + standalone tarball for both arches → GPG signing → SLSA provenance attestation → GitHub Release publish. All green. Verified the published SLSA attestations with `gh attestation verify` (cryptographically valid, Sigstore-backed) and confirmed both the `.deb` and the tarball ship all three binaries (`zallet`, `zallet-zebra`, `zallet-zaino`) and that the launcher dispatch works correctly from the actual released artifacts.

y4ssi:
Fixes #565.

## Summary

Five real, reproducible bugs blocking the arm64 (Nix) release path and one step of the amd64 (StageX) Dockerfile — none previously caught because no release has been cut since the 3-binary workspace split (#540 via #543/#547/#548/#549).

## Bug 1 — `Dockerfile.stagex`

`gen-copyright`'s bare `cargo run -p gen-copyright` fails on this builder because it only has the musl target installed, and a bare invocation resolves via a different cargo codepath than an explicit `--target` one — that codepath can't build `serde_derive`'s proc-macro crate-type for the target, even though the three `cargo install --target ${TARGET_ARCH}` calls right above it succeed building `serde_derive`-dependent crates for the same target. Fixed by passing `--target ${TARGET_ARCH}` explicitly, matching what already works above.

## Bugs 2–5 — `flake.nix`

All four stem from `backends/zebra` and `backends/zaino` each being their own cargo workspace (excluded from the root `Cargo.toml`, by design, per #540) while `flake.nix` builds all three binaries from one shared `src`.

- **Bug 2**: `craneLib.filterCargoSources` only recognises lockfiles belonging to the workspace it can see from `.`, so it silently drops `backends/*/Cargo.lock` from the filtered `$src`. Fixed by explicitly whitelisting `backends/*/Cargo.{toml,lock}` in the source filter.
- **Bug 3**: crane's `mkDummySrc` unconditionally copies the `cargoLock` argument to the dummy tree's root — it has no concept of a lockfile living in a workspace subdirectory, so `backends/zaino`'s git-patched `age` dependency couldn't resolve. Fixed with a `postPatch` that copies the root `Cargo.lock` into the workspace subdirectory Cargo actually reads from.
- **Bug 4**: with `--manifest-path backends/zaino/Cargo.toml`, Cargo resolves the target dir relative to that workspace's root, but crane's install hook looks for `$CARGO_TARGET_DIR` at `$PWD`. Fixed by pinning it, scoped to only the backend workspaces (the launcher's own — already working — install path breaks if this is applied unconditionally).
- **Bug 5**: `installFromCargoBuildLogHook` runs `cargo metadata` with no `--manifest-path`, which tries to resolve `zallet-core`'s dev-dependencies (`proptest`, `trycmd`, ...) that were never vendored for the backends' independent lockfiles, producing a confusing downstream `jq` error that hides the real failure. Fixed via the hook's own `doNotPostBuildInstallCargoBinaries` escape hatch + a manual `installPhaseCommand` for the two backends (must be *omitted*, not emptied, for the launcher so crane's own default still applies there).

Full root-cause detail with error output for each bug is in #565.

## Verification

- All five bugs reproduced and root-caused on a scratch `aarch64-linux` machine (bare `nix build --show-trace -L --keep-failed`, inspecting the actual sandbox contents at failure time).
- After the fixes: `nix build .#zallet-zebra .#zallet-zaino .#zallet-launcher` all succeed, producing correctly statically-linked `aarch64-unknown-linux-musl` binaries.
- Placing all three built binaries side by side and running the launcher: dispatches to `zallet-zebra` by default, and to `zallet-zaino` when `zallet.toml` sets `backend = "zaino"`.
- Ran a full release dry-run end-to-end (tag push → `release.yml`) in a private fork with these fixes: StageX amd64 container build → arm64 Nix build → multi-arch manifest → `.deb` + standalone tarball for both arches → GPG signing → SLSA provenance attestation → GitHub Release publish. All green.
- Verified the resulting SLSA attestations with `gh attestation verify` (cryptographically valid, Sigstore-backed) and confirmed both the `.deb` and the tarball ship all three binaries with correct launcher dispatch from the actual released artifacts.

## Test plan

- [ ] CI green on this PR (in particular the arm64/Nix build job, if it runs on PRs)
- [ ] `nix build .#zallet-zebra .#zallet-zaino .#zallet-launcher` succeeds
- [ ] StageX amd64 Docker build still succeeds (gen-copyright step)

### DIFF
--- a/Dockerfile.stagex
+++ b/Dockerfile.stagex
@@ -81,7 +81,19 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     # Generate the Debian copyright file across the UNION of the three shipped
     # binaries' dependency graphs (gen-copyright resolves each workspace's
     # lockfile; the per-binary graphs differ, so no single build can see all).
-    && cargo run --release --locked -p gen-copyright \
+    #
+    # This stagex builder's rustc has ONLY the musl target installed (there is
+    # no separate glibc host triple), so a bare `cargo run` (no --target) still
+    # implicitly resolves to x86_64-unknown-linux-musl, but by a different
+    # cargo codepath than an explicit `--target` invocation: it fails to build
+    # `serde_derive`'s proc-macro crate-type for that target ("cannot produce
+    # proc-macro ... as the target x86_64-unknown-linux-musl does not support
+    # these crate types"), even though the three `cargo install --target
+    # ${TARGET_ARCH}` calls above build serde_derive-dependent crates for the
+    # exact same target successfully. Passing --target explicitly here too
+    # (matching the pattern that already works above) avoids the bare-cargo
+    # codepath and fixes it.
+    && cargo run --release --locked -p gen-copyright --target ${TARGET_ARCH} \
         -- --out /usr/local/share/zallet/debian-copyright
 
 # --- Stage 2: layer for local binary extraction ---
--- a/flake.nix
+++ b/flake.nix
@@ -24,9 +24,25 @@
           targets = [ muslTarget ];
         };
         craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
+        # craneLib.filterCargoSources only recognises Cargo.toml/Cargo.lock
+        # files that belong to the ROOT workspace's resolution graph. Since
+        # zcash/zallet#540, backends/zebra and backends/zaino are each their
+        # own workspace (excluded from the root Cargo.toml's `[workspace]`),
+        # so filterCargoSources silently drops their Cargo.lock files from the
+        # filtered $src — Cargo.toml and src/ survive (matched generically),
+        # but Cargo.lock does not. `replaceCargoLockHook`'s `replaceCargoLock`
+        # then has no backends/*/Cargo.lock to move aside, and instead
+        # clobbers the ROOT Cargo.lock with the backend's lockfile content,
+        # leaving Cargo unable to resolve backends/zaino's git-patched `age`
+        # dependency correctly ("requires a lock file to be present first").
+        # Explicitly keep every backends/*/Cargo.lock (and Cargo.toml, for
+        # symmetry/robustness) regardless of what filterCargoSources decides.
+        isBackendCargoFile = path:
+          (builtins.match ".*/backends/[^/]+/Cargo\\.(toml|lock)$" path != null);
         ftlOrCargo = path: type:
           (builtins.match ".*\\.ftl$" path != null)
           || (builtins.match ".*/i18n\\.toml$" path != null)
+          || (isBackendCargoFile path)
           || (craneLib.filterCargoSources path type);
         src = pkgs.lib.cleanSourceWith { src = ./.; filter = ftlOrCargo; name = "source"; };
         # The *-sys C/C++ deps need a coherent musl toolchain. pkgsMusl.clangStdenv.cc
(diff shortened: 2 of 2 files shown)


nuttycom (approved):
utACK

PR #569, opened 2026-07-09
backend: support regtest in the zebra read-state backends

dannywillems:
## Summary

The `zebra` chain backend (and the `zaino` backend's read-state-service
mode) previously rejected regtest at startup with a "does not support
regtest" error. This wires regtest through so both backends can read a
co-located regtest zebrad's on-disk state.

## What changed

- `network_to_zebra` (in the shared `zallet-zebra-read-state` crate) now
  takes the wallet's full consensus parameters (`&impl Parameters`)
  instead of a bare `NetworkType`. For regtest it reads each network
  upgrade's activation height through the `Parameters` trait, builds
  Zebra's `ConfiguredActivationHeights`, and constructs a Zebra Regtest
  network via `Network::new_regtest`, so the read-state service
  interprets zebrad's state under matching consensus rules.
- Both backend call sites pass their network parameters through. The
  read-state crate stays decoupled from `zallet-core` (it only needs
  `zcash_protocol` and `zebra-chain`).
- Docs (`main.rs` module comment, `book/src/guide/setup.md`) updated to
  drop the "regtest unsupported" wording.

## Testing

- New unit test `network_to_zebra_builds_regtest_with_matching_heights`
  asserts a configured NU5 height is threaded end-to-end into the Zebra
  regtest network; `network_to_zebra_maps_main_and_test` still passes.
- `cargo clippy --all-targets` and `cargo fmt -- --check` clean for the
  zebra backend, the zaino backend, and the read-state crate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Added
+
+- The `zebra` chain backend (and the `zaino` backend's read-state-service mode)
+  now support regtest. The read-state service builds a Zebra Regtest network
+  whose network-upgrade activation heights mirror the wallet's configured
+  `regtest_nuparams`, so it interprets the co-located zebrad's on-disk state
+  under matching consensus rules. Previously these backends rejected regtest at
+  startup with an "does not support regtest" error.
+
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


nullcopy (approved):
utACK c941722abdb0d124639f358f6c3dc11abff38cc5

nuttycom (approved):

PR #570, opened 2026-07-09
rpc: surface Ironwood spends and outputs in z_viewtransaction

dannywillems:
## Motivation

`z_viewtransaction` was the last place in the wallet where Ironwood was still
stubbed out. Ironwood actions were elided: the received-note lookup and the
sent-to-address lookup both returned early for the Ironwood pool, so Ironwood
spends, outputs, and sent-to addresses were reported for neither. Every other
surface (balances, `list_transactions`, `list_unspent`, sync, chain readers)
already handled Ironwood.

## Solution

Ironwood notes are Orchard-shaped, so this mirrors the existing Orchard handling:

- `output_with_nullifier` now resolves Ironwood spends against the
  `ironwood_received_notes` table (indexed by action index, like Orchard).
- `sent_to_address` maps Ironwood to `sent_notes.output_pool = 4`.
- An Ironwood spend loop iterates `tx.ironwood_bundle()` actions and matches
  Orchard-shaped nullifiers; it runs before `funded_by_this_wallet` so Ironwood
  spends contribute to it.
- An Ironwood output loop trial-decrypts each action with the account's Orchard
  viewing keys (IVKs for incoming, OVKs for outgoing) under the Ironwood
  note-encryption domain (Ironwood carries version 3 note plaintexts).

## Test evidence

- `cargo build` / `cargo clippy --all-targets` / `cargo fmt --check` clean on
  `zallet-core`.
- `cargo clippy --all-targets` clean on the `zebra` backend.
- `cargo test -p zallet-core` green (187 passed).

Not yet exercised end-to-end against a real v6 transaction carrying an Ironwood
bundle; the logic is a structural mirror of the verified Orchard path.

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/view_transaction.rs
+++ b/zallet-core/src/components/json_rpc/methods/view_transaction.rs
@@ -2,7 +2,9 @@ use std::collections::BTreeMap;
 
 use documented::Documented;
 use jsonrpsee::core::RpcResult;
-use orchard::note_encryption::OrchardDomain;
+use orchard::note_encryption::{
+    DomainVersion, IronwoodVersion, NoteEncryptionDomain, OrchardVersion,
+};
 use rusqlite::{OptionalExtension, named_params};
 use schemars::JsonSchema;
 use serde::Serialize;
@@ -19,7 +21,7 @@ use zcash_protocol::{
     ShieldedPool, TxId,
     consensus::{BlockHeight, Parameters},
     memo::Memo,
-    value::{BalanceError, Zatoshis},
+    value::{BalanceError, ZatBalance, Zatoshis},
 };
 
 use crate::components::{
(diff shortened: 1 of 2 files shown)


zallet-core/src/components/json_rpc/methods/view_transaction.rs
```diff
@@ -540,6 +542,36 @@ pub(crate) async fn call<C: Chain>(wallet: &DbConnection, chain: C, txid_str: &s
         }
     }
 
+    // Ironwood spends. Ironwood actions are Orchard-shaped and carry Orchard nullifiers, so this
+    // mirrors the Orchard spend handling above; only the pool (and thus the received-notes table
+    // that `output_with_nullifier` consults) differs.
```
nullcopy: I feel like all these "Ironwood is the same shape as Orchard" changes should be abstracted into "Orchard family" methods

nullcopy (approved):
utACK 9bc88200b96a20e3a630d58d24bae7592f4afab9

PR #571, opened 2026-07-09
Update to the pure zewif-based zcashd wallet import stack.

nuttycom:
Subsumes #551

### DIFF
--- a/zallet-core/src/commands/migrate_zcashd_wallet.rs
+++ b/zallet-core/src/commands/migrate_zcashd_wallet.rs
@@ -148,7 +148,7 @@ impl MigrateZcashdWalletCmd {
         .map_err(|e| MigrateError::Zewif {
             error_type: ZewifError::BdbDump,
             wallet_path: wallet_path.to_path_buf(),
-            error: e,
+            error: e.into(),
         })?;
 
         let zcashd_dump =
@@ -156,7 +156,7 @@ impl MigrateZcashdWalletCmd {
                 MigrateError::Zewif {
                     error_type: ZewifError::ZcashdDump,
                     wallet_path: wallet_path.clone(),
-                    error: e,
+                    error: e.into(),
                 }
             })?;
 
@@ -165,7 +165,7 @@ impl MigrateZcashdWalletCmd {
                 MigrateError::Zewif {
                     error_type: ZewifError::ZcashdDump,
                     wallet_path,
-                    error: e,
+                    error: e.into(),
                 }
             })?;
 
@@ -173,15 +173,15 @@ impl MigrateZcashdWalletCmd {
     }
 
     fn check_network(
-        zewif_network: zewif::Network,
+        zewif_network: &zewif::Network,
         network_type: NetworkType,
     ) -> Result<NetworkType, MigrateError> {
         match (zewif_network, network_type) {
-            (zewif::Network::Main, NetworkType::Main) => Ok(()),
-            (zewif::Network::Test, NetworkType::Test) => Ok(()),
-            (zewif::Network::Regtest, NetworkType::Regtest) => Ok(()),
+            (zewif::Network::Mainnet, NetworkType::Main) => Ok(()),
+            (zewif::Network::Testnet, NetworkType::Test) => Ok(()),
+            (zewif::Network::Regtest(_), NetworkType::Regtest) => Ok(()),
             (wallet_network, db_network) => Err(MigrateError::NetworkMismatch {
-                wallet_network,
+                wallet_network: wallet_network.clone(),
                 db_network,
             }),
         }?;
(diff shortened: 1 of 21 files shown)


zallet-core/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -202,800 +187,530 @@ impl MigrateZcashdWalletCmd {
...
+                    .map(BlockHeight::from_u32)
             })
-            .collect::<Vec<_>>();
-        if skipped_unparseable_scripts > 0 {
-            warn!(
-                "Skipped {} watch-only redeem scripts that failed to parse in total.",
-                skipped_unparseable_scripts,
-            );
-        }
+            .unwrap_or(sapling_activation);
```
nullcopy, line 225: We should not default to sapling_activation height here. Perhaps genesis?

zallet-core/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -202,800 +187,530 @@ impl MigrateZcashdWalletCmd {
...
+        // constructs precise account birthdays with no further chain access. In
+        // no-scan mode, estimate a conservative birthday from transaction expiry
+        // heights; the importer will schedule a rescan from there.
+        let (birthday_chain_state, recover_until) = if let Some(chain_view) = chain_view.as_ref() {
+            let mut block_heights = HashMap::new();
+            for tx in document.transactions().values() {
+                if let Some(position) = tx.block_position() {
+                    let block_hash = BlockHash(*position.block_hash().as_bytes());
+                    if let Entry::Vacant(entry) = block_heights.entry(block_hash) {
+                        // Ignore any blocks that are not in the main chain.
```
nullcopy, line 342: Thinking about this some more and realizing maybe we shouldn't ignore these. I think the're still a good indication that an address was exposed around a given height, even if the transaction ultimately did not get included in the chain.

zallet-core/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -202,800 +187,530 @@ impl MigrateZcashdWalletCmd {
...
-        for (txid, wallet_tx) in wallet.transactions().iter() {
-            let block_hash = BlockHash(*wallet_tx.hash_block().as_ref());
-            tx_heights.insert(
-                txid,
-                (
-                    main_chain_block_heights.get(&block_hash).cloned(),
-                    buffer_wallet_transactions.then_some(wallet_tx),
-                ),
+            info!(
+                "Wallet document references {} mined main-chain blocks",
```
nullcopy, line 350: (nit) would be nice to indicate how many (if any) were recorded as mined, but not actually included in the chain

nullcopy (commented):
Flushing comments. Partial review as of 2311107cb30cd784bcc58aad1714eee59061a338

nullcopy (approved):
Some issues that should be fixed, but possible to defer to future work.

utACK 2311107cb30cd784bcc58aad1714eee59061a338

nullcopy (approved):
utACK 7d12301c3bc9230f6ec41e9d4e69bbee0cd5526b

zallet-core/src/commands/migrate_zcashd_wallet.rs
```diff
@@ -202,800 +187,530 @@ impl MigrateZcashdWalletCmd {
...
-                         the imported script will not be marked as exposed.",
-                    );
-                    debug_assert!(false, "sh() descriptor should always yield an address");
-                }
-                // `import_standalone_transparent_script` currently only accepts multisig
-                // redeem scripts within the P2SH size limit; treat those rejections
-                // (`SqliteClientError::BadAccountData`) as warn-and-skip so a single
-                // unsupported script does not abort the migration. Any other error
-                // (corrupted data, DB failure, conflict, ...) is propagated.
-                match db_data.import_standalone_transparent_script(target_account, script) {
```
nullcopy, line 819: P2SH import was removed and not replaced. Need to fix this at some point.

nullcopy: Created an issue for this https://github.com/zcash/zallet/issues/572

zallet-core/src/commands/migrate_zcashd_wallet.rs
nullcopy: Created an issue for this: https://github.com/zcash/zallet/issues/573

zallet-core/src/commands/migrate_zcashd_wallet.rs
nullcopy: Created an issue for later: https://github.com/zcash/zallet/issues/574

PR #576, opened 2026-07-09
backend: read the Ironwood note commitment tree in the zebra backend

zcash-harry:
## Summary

`tree_state_as_of` in the `zebra` chain backend always reported the Ironwood note
commitment tree as empty, regardless of the queried height. This was correct before
NU6.3 activated on any chain the backend could reach, but testnet has since crossed
activation, and the stale assumption now silently corrupts sync: every batch boundary
re-derives the Ironwood frontier as empty instead of reading the real (non-empty) tree,
so the wallet's local shardtree conflicts with itself on the very next batch — surfacing
as a checkpoint conflict or `Inserted root conflicts with existing root` error from
`zcash_client_sqlite`/`shardtree`.

Fix: read the tree the same way as Sapling and Orchard, via `ReadRequest::IronwoodTree`
on the `ChainReader` trait. Ironwood shares the Orchard tree's shape, so the decode path
mirrors the existing `orchard_tree_bytes` handling exactly.

## Test plan

- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo test` (chain module) passing
- [x] Verified live on a synced testnet `zebrad` past NU6.3 activation height (4,134,000):
      wallet Ironwood checkpoints now record real chain-scale positions instead of
      restarting from 0, `steady_state` and history-recovery sync both run without
      crashing, and a same-wallet Orchard→Ironwood `z_sendmany` mined and confirmed with
      the value correctly landing in the recipient's Ironwood pool.

### AI disclosure

Claude Code was used for root-cause analysis, the fix, and this PR description. I
reviewed and independently verified the fix end-to-end on testnet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,16 @@ be considered breaking changes.
   transient stale-view error (e.g. a reorg in progress right as the wallet
   started) that `steady_state`'s main loop already tolerates. It now retries the
   same way.
+- The `zebra` chain backend's `tree_state_as_of` always reported the Ironwood
+  note commitment tree as empty, regardless of the queried height. This was
+  correct before NU6.3 activated on any chain the backend could reach, but once
+  a chain crossed activation, every sync batch boundary re-derived the Ironwood
+  frontier as empty instead of reading the real (non-empty) tree, corrupting
+  the wallet's local shardtree state on the very next batch (surfacing as a
+  checkpoint or root conflict in `zcash_client_sqlite`, or an `Inserted root
+  conflicts with existing root` error from `shardtree`). The tree is now read
+  from the backend the same way as Sapling and Orchard, via
+  `ReadRequest::IronwoodTree`.
 
 ## [0.1.0-alpha.4] - 2026-06-25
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,16 @@ be considered breaking changes.
...
+- The `zebra` chain backend's `tree_state_as_of` always reported the Ironwood
+  note commitment tree as empty, regardless of the queried height. This was
+  correct before NU6.3 activated on any chain the backend could reach, but once
+  a chain crossed activation, every sync batch boundary re-derived the Ironwood
+  frontier as empty instead of reading the real (non-empty) tree, corrupting
+  the wallet's local shardtree state on the very next batch (surfacing as a
+  checkpoint or root conflict in `zcash_client_sqlite`, or an `Inserted root
+  conflicts with existing root` error from `shardtree`). The tree is now read
+  from the backend the same way as Sapling and Orchard, via
+  `ReadRequest::IronwoodTree`.
```
nuttycom, line 49: nonblocking, can clean up before release: this is fixing an error that was introduced since the last release, so we don't have to have it in the changelog

nuttycom (approved):
utACK

dannywillems (approved):
utACK

PR #577, opened 2026-07-09
rpc: surface Ironwood in z_getnotescount and decoderawtransaction

dannywillems:
## Motivation

Two informational RPCs still omitted the Ironwood (NU6.3, ZIP 2005) value pool
once it is active, even though the balance/listing RPCs already surface it:

- `z_getnotescount` reported only `sapling` and `orchard`, so Ironwood notes were
  counted in no field at all.
- `getrawtransaction` / `decoderawtransaction` rendered the `orchard` bundle but
  not the Ironwood one, so a v6 transaction's Ironwood actions never appeared in
  the decoded output.

## Changes

- **`z_getnotescount`**: add an `ironwood` field and count
  `ShieldedPool::Ironwood` notes from the account metadata.
- **`getrawtransaction` / `decoderawtransaction`**: add an `ironwood` object,
  gated on `tx.version().has_ironwood()`, encoding `tx.ironwood_bundle()` with the
  same encoder as the Orchard bundle (Ironwood actions are Orchard-shaped, so the
  object has the same shape as `orchard`).
- **Dependency bump (first commit)**: `z_getnotescount` depends on
  `WalletRead::get_account_metadata` counting Ironwood notes, which required
  advancing the librustzcash pin (to `29c7fb2`). That revision moved off the
  git-patched shardtree (0.6.2) onto the published `shardtree 0.7`, so the
  workspace `shardtree` requirement moves to `0.7` and the temporary
  `shardtree`/`incrementalmerkletree` git `[patch.crates-io]` is dropped from all
  three manifests (only the `orchard` patch remains, still required).

## Testing

Verified against the Z3 integration-tests stack (zebrad + zaino + zallet) with
the `zaino` backend:

- `z_getnotescount` returns `{sprout, sapling, orchard, ironwood}` and reports the
  correct Ironwood note count once the notes' transactions are mined (the count
  keys off `transactions.mined_height`, which backfills a scan pass after a note
  first becomes visible).
- `decoderawtransaction` of a v6 Ironwood spend now includes an `ironwood` object
  carrying the bundle's actions.

Corresponding integration-tests assertions are added in
zcash/integration-tests#159.

### DIFF
--- a/backends/zaino/Cargo.toml
+++ b/backends/zaino/Cargo.toml
@@ -79,28 +79,19 @@ tempfile = "3"
 #
 # Update the librustzcash rev ONLY via utils/sync-librustzcash.sh (which syncs
 # all three manifests and lockfiles at once); do not edit these revs by hand.
-equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
-f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
-transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
-zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
-zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
-zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
-zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "9fe5abf78a15940aa84fd895fed75162f21fcb53" }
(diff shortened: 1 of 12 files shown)


backends/zaino/Cargo.toml
```diff
@@ -79,28 +79,19 @@ tempfile = "3"
...
-# [patch] from the ROOT workspace, so it must be replicated in every manifest.
-# Points at zcash/incrementalmerkletree main (includes #184); replace with a
-# published shardtree release once one is cut. Keep in lockstep with the root and
-# backends/zebra manifests.
-shardtree = { git = "https://github.com/zcash/incrementalmerkletree", rev = "decefc469dbf1e686b20b7e7dcaa7cb4f122125b" }
-incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree", rev = "decefc469dbf1e686b20b7e7dcaa7cb4f122125b" }
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "29c7fb28889a69df15b44fc9465cf4a464f766e9" }
+f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "29c7fb28889a69df15b44fc9465cf4a464f766e9" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "29c7fb28889a69df15b44fc9465cf4a464f766e9" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "29c7fb28889a69df15b44fc9465cf4a464f766e9" }
```
dannywillems, line 85: Gonna be using the release after.

nuttycom (approved):
utACK b450f6d2a45baa769646648d9051342ed87cf5cd

Issue #578, opened 2026-07-09
Rescan already-scanned ranges when a key/account is imported with an earlier birthday


oxarbitrage:
Follow-up to #563. That fix makes `z_importkey` / `z_getnewaccount` / `z_recoveraccount` reload the sync engine's viewing keys so a key added to a running wallet is scanned **going forward**, without a restart.

It does **not** cover the other half: `reload_keys()` only affects blocks scanned *after* the reload. A key whose **birthday falls in an already-scanned range** still needs an explicit rescan of that range to find its historical notes.

`WalletWrite::rewind_to_height` could drive this (see the TODO in `import_key.rs`), but it currently rewinds the **entire wallet** (every account), not just the imported key's range — so it's deferred until that global side effect is acceptable/desired.

Notes from @nuttycom (#563):
- Rewinds are cheaper than they used to be (with optimization scope remaining) — taking this on is high-value.
- Rewinds aren't required for **transparent** key import; transactions-involving-address queries can find what's needed there.

oxarbitrage:
Design note while implementing this, in case anyone revisits the choice.

The TODO this issue came from named `rewind_to_height`, and an earlier revision of #581 used `truncate_to_height`. **That primitive cannot close this issue**, for a structural reason worth recording:

`truncate_to_height` does not lower an account's birthday, and suggested scan ranges are floored at `MIN(birthday_height)` across *all* accounts (`zcash_client_sqlite` at the currently pinned librustzcash `1f6bb20`: `wallet.rs:3240`, `wallet/scanning.rs:649`). That floor is deliberate — the comment at `scanning.rs:645-648` explains it exists because account creation requires a tree frontier, so the wallet has no tree state below the birthday to scan against. Consequently a truncation below the wallet birthday is either clipped or rejected with `RequestedRewindInvalid`, and this issue's headline case — a key "imported with an earlier birthday" — is exactly that case.

It also means the documented default `startHeight=0` could never work via truncation: `select_truncation_height` requires a `blocks` row at or below the target, and none exists at height 0.

`rewind_to_chain_state` is the API that moves the floor: it takes the `ChainState` frontier for the new birthday and lowers the birthdays of the accounts it is given. It is also what the *new-key* import path already uses (`add_account`). #581 now routes the existing-key path through the same call.

One caveat that outlives #581: while #588 is open, any RPC-driven rewind leaves the `(tip - 100, tip]` band unscanned, so a rescan is only fully effective below that boundary.

PR #579, opened 2026-07-09
fix: reload sync engine viewing keys after key-adding RPCs

### LINKED ISSUE #563: `z_importkey` into a running wallet: imported key not scanned until restart
### Summary

Importing a key into a **running** wallet (`z_importkey`) doesn't make the sync engine scan for it.
The account + UFVK are written and the scan range is re-queued, but the batch decryptor
(`zcash_client_backend::sync::decryptor`) keeps trial-decrypting with the key set it loaded at
startup — it only refreshes on `Handle::reload_keys()`, which nothing in zallet ever calls. So the
imported key's notes go undetected (`z_gettotalbalance` stays 0 even at the chain tip) until a
restart, which reloads keys. Surfaced recovering a real Sapling balance via `z_importkey` (#400).

### Fix + open questions

Call `reload_keys()` after an import. Either **push** (thread the decryptor handle to the RPC layer,
call it in `z_importkey` — precise, but per-RPC plumbing) or **pull** (the sync engine already holds
the handle; reload there — one-file change in `sync.rs` that covers all import paths, at some
reload-frequency cost).

- Push vs pull?
- Scope: `z_importkey` only, or also `z_importviewingkey` / `importaddress` (same blind spot)?

Caveat: `reload_keys()` only affects blocks scanned *after* the reload; a key whose birthday is in an
already-scanned range still needs an explicit rescan (cf. the `rewind_to_height` TODO in
`import_key.rs`). Recover-from-genesis is unaffected.

oxarbitrage:
Importing a key or adding an account to a **running** wallet never made the sync engine scan for it: the batch decryptor loads its viewing keys once at startup and only refreshes on `reload_keys()`, which nothing called. So imported/created keys went undetected (`z_gettotalbalance` stayed 0 at the chain tip) until a restart.

This threads the decryptor handle to the affected RPCs and reloads keys after each one. To avoid delaying RPC startup, the decryptor is now built up front and its handle shared with the RPC server, rather than reordering sync ahead of the RPC server.

**Scope** — reloads after all three key/account-adding RPCs: `z_importkey`, `z_getnewaccount`, `z_recoveraccount`. (`z_importviewingkey` isn't implemented; `z_importaddress` is transparent-only and doesn't use the shielded decryptor.) The reload is fire-and-don't-block (queued, not awaited) with a `warn!` if the engine has shut down, so it never hangs the RPC during heavy sync.

Rewind-to-height for already-scanned ranges (the other half of the caveat in #563) is tracked separately in #578.

**Commits:**
- fix(rpc): reload sync engine viewing keys after z_importkey
- fix(rpc): also reload after z_getnewaccount and z_recoveraccount
- refactor(sync): use WalletDecryptorHandle alias for existing decryptor usages

Closes #563

### DIFF
--- a/zallet-core/src/commands/start.rs
+++ b/zallet-core/src/commands/start.rs
@@ -61,6 +61,9 @@ impl StartCmd {
         // returns the height at which to shut down before reaching them.
         let shutdown_height = check_consensus_compatibility(&chain).await?;
 
+        // Build the decryptor up front so the RPC server has its handle before the initial scan.
+        let (decryptor_handle, decryptor_engine) = WalletSync::build_decryptor();
+
         // Launch RPC server.
         let rpc_task_handle = JsonRpc::spawn(
             &config,
@@ -68,6 +71,8 @@ impl StartCmd {
             #[cfg(zallet_build = "wallet")]
             keystore,
             chain.clone(),
+            #[cfg(zallet_build = "wallet")]
+            decryptor_handle.clone(),
         )
         .await?;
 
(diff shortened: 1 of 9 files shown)


nuttycom (approved):
utACK 0e696a19a9c9398331ce46142e791a1ebe6f5d10

PR #580, opened 2026-07-09
Reduce the risk of lost funds by adding backup warnings

defuse:


nullcopy (approved):
utACK b41bce34e789674a51f2b9efb507cdb1e957dcb7

pacu (approved):
Although the reliability of zcashd wallet.dat migration will keep improving over the next days as we keep addressing partners' feedback on the process, these are good remarks of the _current_ implications of this important operations and it's a good documentation to have. 

Thanks @defuse!

nuttycom (approved):
ACK, these are good warnings.

Issue #76, opened 2025-04-03
rpc: Implement `z_getnotescount`


str4d:

str4d:
#14 implemented the basic version of this. Still needs finishing work to support the `minconf` and `as_of_height` parameters.

str4d:
This was finished in #318.

PR #581, opened 2026-07-10
fix(rpc): rescan already-scanned range when z_importkey re-imports a known key with rescan=yes

### LINKED ISSUE #578: Rescan already-scanned ranges when a key/account is imported with an earlier birthday

oxarbitrage:
Closes #578.

Follow-up to #579 (merged), which made key imports scan **going forward** without a restart. #578 is the other half: re-importing an **already-known** key with `rescan="yes"` did nothing (`z_importkey` short-circuits on `is_new_key`). New keys are already covered by the account import queuing their birthday range; this wires up the existing-key path.

**What changed** (`import_key.rs`): when the key is already known and `rescan="yes"`, rewind to the chain state just before `startHeight` and reset *that account's* birthday to it, so the sync engine re-scans `startHeight..=tip` and finds notes in blocks that were scanned before this key was known. This reuses the same `fetch_account_birthday` → `rewind_to_chain_state` path the new-key import already goes through, rather than introducing a second mechanism.

**Why `rewind_to_chain_state` and not `truncate_to_height`.** An earlier revision of this PR used `truncate_to_height`, because at librustzcash `29c7fb2` `rewind_to_chain_state` did not snap its truncation target to a checkpoint shared by all pools — it asserted they matched and failed with *"Data DB is corrupted: Sapling and Orchard should have the same checkpoints"*. **That is fixed upstream.** At the currently pinned `1f6bb20` it computes a `window_floor` as the minimum over every active pool (Sapling / Orchard / Ironwood) of `min_checkpoint_id_at_or_above` (`zcash_client_sqlite/src/wallet.rs:4866-4884`), with dedicated tests for straddling and lagging pool checkpoints.

That matters because `truncate_to_height` **cannot lower an account's birthday**, and suggested scan ranges are floored at `MIN(birthday_height)` across *all* accounts (`wallet.rs:3240`, `wallet/scanning.rs:649`). That floor exists because account creation requires a tree frontier (`scanning.rs:645-648`). So a plain truncation below the wallet birthday is either clipped or rejected — which is exactly the case #578 leads with ("imported with an earlier birthday"). It also made the documented default `startHeight=0` fail with a bare `-1 "Rescan failed"`, since `select_truncation_height` needs a `blocks` row at or below the target and none exists at height 0.

**Behaviour notes:**
- This is a **wallet-wide** rewind: the note commitment trees are shared across accounts, so every account re-scans from `startHeight`. Only the re-imported account's birthday is lowered.
- A rewind that would land below every account's birthday without acknowledgement returns `RewindBeyondBirthdays`; that is mapped to `-8` (invalid parameter) rather than a generic failure.

**Testing.** `cargo check` / `fmt` / `clippy` clean; 388 unit tests pass, including 4 new ones covering the `(is_new_key, rescan)` decision matrix extracted as `should_rescan_existing_key`.

End-to-end against zcash/integration-tests#76 (`wallet_import_export_key.py`, branch `export-import-key-test`) on regtest with `zebrad 6.0.0-rc.0`: **this PR's scenario passes.** Instrumenting the run to print raw `z_getbalances`, the re-imported account reports `sapling.spendable.valueZat = 100000000` within ~20s of `z_importkey(key, "yes", funding_height)`, with `wallet_tip = fully_synced_height`, stable across repeated polls. The account birthday is correctly lowered (124 → 123) and the previously-missed note is attributed to it.

⚠️ **The full IT#76 run still fails, in a later phase, on #588 — not on this change.** After the fresh-import phase the wallet is left with `blocks` at 1..145 while the node is at 245, and a `Historic(144..245)` scan range that is never serviced: `recover_history` only handles ranges below the `tip - 100` boundary, and `steady_state` never consults the scan queue. **This reproduces identically with a build of this branch's previous `truncate_to_height` revision, and is unchanged at both a 450s and a 1200s timeout** (stuck, not slow) — so it is a pre-existing sync bug, not a regression from this PR. It matches #588, where a rewind-installed `Historic` range lands in a band that `recover_history` (bounded below `tip - 100`) will not service and that `steady_state` never looks at, because it follows an in-memory tip rather than the scan queue.

The practical consequence for this PR: any RPC-driven rescan currently leaves the `(tip - 100, tip]` band unscanned until #588 is fixed. `rescan="yes"` on a known key is still a strict improvement over the current behaviour of doing nothing at all.

_(Note: the previous revision of this description claimed IT#76 was fully green. That was measured against librustzcash `29c7fb2` before the pin moved, and no longer holds; it has been corrected above.)_

ZIT-Revision: export-import-key-test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Changed
+
+- `z_importkey` with `rescan="yes"` now re-scans when the key is already known to
+  the wallet; previously it scanned nothing. The account's birthday is lowered to
+  `startHeight`, so the rescan reaches blocks below the wallet's previous earliest
+  birthday. This rewinds the **entire wallet** to `startHeight`: the note commitment
+  trees are shared, so every account re-scans from there.
+
 ## [0.1.0-beta.3] - 2026-08-24
 
 ### Added
--- a/zallet-core/src/components/json_rpc/methods/import_key.rs
+++ b/zallet-core/src/components/json_rpc/methods/import_key.rs
@@ -124,11 +124,6 @@ pub(crate) async fn call<C: Chain>(
         //   historical blocks from that point.
         // - "no" → use the current chain tip so the sync engine only tracks new
         //   transactions going forward.
-        //
-        // TODO: When rescan is "yes" and the key already exists, zcashd would force a
-        // rescan from start_height. `WalletWrite::rewind_to_height` could now drive this,
-        // but it rewinds the *entire* wallet (every account) rather than just this key, so
-        // we defer wiring it up until that global side effect is the desired behaviour.
         let effective_height = match rescan {
             "yes" | "whenkeyisnew" => start_height,
             "no" => chain_tip.unwrap_or_else(|| {
@@ -161,6 +156,14 @@ pub(crate) async fn call<C: Chain>(
                 "sync engine has shut down; imported key won't be scanned until restart"
             );
         }
+    } else if rescan == "yes" {
+        // The key is already known but the caller requested a rescan: truncate to just before
+        // `start_height` so the sync engine re-scans `start_height..=tip`. Use `truncate_to_height`,
+        // not the account-scoped `rewind_to_chain_state` — the latter doesn't snap to a shared
+        // Sapling/Orchard checkpoint and corrupts the DB on a deep rewind (integration-tests #76).
+        wallet
+            .truncate_to_height(start_height.saturating_sub(1))
+            .map_err(|e| LegacyCode::Misc.with_message(format!("Rescan failed: {e}")))?;
     }
 
     Ok(ResultType {
--- a/zallet-core/src/components/json_rpc/methods/import_key.rs
+++ b/zallet-core/src/components/json_rpc/methods/import_key.rs
@@ -1,8 +1,12 @@
+use std::collections::HashSet;
+
 use documented::Documented;
 use jsonrpsee::core::RpcResult;
 use schemars::JsonSchema;
 use serde::Serialize;
-use zcash_client_backend::data_api::{AccountPurpose, WalletRead, WalletWrite};
+use zcash_client_backend::data_api::{
+    Account, AccountPurpose, WalletRead, WalletWrite, error::RewindError,
+};
 use zcash_client_sqlite::error::SqliteClientError;
 use zcash_keys::{encoding::decode_extended_spending_key, keys::UnifiedFullViewingKey};
 use zcash_protocol::consensus::{BlockHeight, NetworkConstants};
@@ -44,6 +48,13 @@ fn validate_rescan(rescan: Option<&str>) -> RpcResult<&str> {
         Some(_) => Err(LegacyCode::InvalidParameter
             .with_static("Invalid rescan value. Must be \"yes\", \"no\", or \"whenkeyisnew\".")),
     }
+}
+
+/// Whether re-importing an already-known key should rewind the wallet to rescan.
+///
+/// A new key needs no rewind: its account import already queues the birthday range.
+fn should_rescan_existing_key(is_new_key: bool, rescan: &str) -> bool {
+    !is_new_key && rescan == "yes"
 }
 
 /// Decodes a Sapling extended spending key and derives the default payment address.
@@ -131,10 +142,10 @@ pub(crate) async fn call<C: Chain>(
         .map_err(|e| LegacyCode::Wallet.with_message(e.to_string()))?;
 
     // Check if the key is already known to the wallet.
-    let is_new_key = wallet
+    let existing_account = wallet
         .get_account_for_ufvk(&ufvk)
-        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?
-        .is_none();
+        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?;
+    let is_new_key = existing_account.is_none();
 
     // For a new key, resolve its birthday before opening the transaction, since the birthday
     // fetch is async and may query the chain.
@@ -144,11 +155,6 @@ pub(crate) async fn call<C: Chain>(
         //   historical blocks from that point.
         // - "no" → use the current chain tip so the sync engine only tracks new
         //   transactions going forward.
-        //
-        // TODO: When rescan is "yes" and the key already exists, zcashd would force a
-        // rescan from start_height. `WalletWrite::rewind_to_height` could now drive this,
-        // but it rewinds the *entire* wallet (every account) rather than just this key, so
-        // we defer wiring it up until that global side effect is the desired behaviour.
         let effective_height = match rescan {
             "yes" | "whenkeyisnew" => start_height,
             "no" => chain_tip.unwrap_or_else(|| {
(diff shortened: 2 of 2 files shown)


oxarbitrage:
### Handoff from zcash/integration-tests #76: the `rescan="yes"` rewind corrupts the wallet DB

While wiring up the rescan-on-import test scenarios in [zcash/integration-tests#76](https://github.com/zcash/integration-tests/pull/76) (the ones this PR will run via `ZIT-Revision: export-import-key-test`), I ran the new `qa/rpc-tests/wallet_import_export_key.py` **locally against this branch** and the core rewind path fails.

**Binaries under test**
- `zallet` + `zallet-zaino` built from this branch (`import-rescan-rewind-578` @ `38610e3`).
- `zebrad` built from `zcash/zebra` `main` (`2f1fd8531`, reports `6.0.0-rc.0`).
- Regtest, NU6.2 active at height 1 (Orchard active; no Ironwood).

**Symptom.** Re-importing an already-known key with `rescan="yes"` aborts:

```
z_importkey("<sapling key already in the wallet>", "yes", 1)
→ error: Rescan failed: Wallet data source error:
  Data DB is corrupted: Sapling and Orchard should have the same checkpoints
```

Reproduced **twice independently** in the test: (a) importing the key with an early birthday so it finds a funded note, then re-importing with `rescan="yes"`; and (b) importing with a birthday *after* the note (so it is missed), then re-importing with `rescan="yes", startHeight=1` to rewind and pick it up. Both hit the identical corruption at the moment of the rewind. Normal scanning is fine — the `#579` fresh-early-birthday scenario finds the historical note without issue, so this is specific to the rewind, not to scanning.

**Where.** The new already-known-key branch this PR adds in `import_key.rs`:

```rust
wallet
    .rewind_to_chain_state(
        birthday.prior_chain_state().clone(),
        std::iter::once(account_id).collect(),
    )
    .map_err(|e| LegacyCode::Misc.with_message(format!("Rescan failed: {e}")))?;
```

is what raises the `Rescan failed: …` above.

**Root-cause hypothesis.** `zcash_client_sqlite` enforces that after any rewind/truncate every shielded pool has checkpoints at the *same* heights. The account-scoped `rewind_to_chain_state(birthday.prior_chain_state(), {account_id})` here appears to leave the Sapling and Orchard note-commitment-tree checkpoints out of lockstep — likely because the `ChainState` from `fetch_account_birthday(start_height).prior_chain_state()` doesn't carry matching per-pool frontiers for a wallet that has already scanned Orchard-active blocks, so the rewind lands Sapling and Orchard checkpoints at different heights. (The removed TODO noted that `WalletWrite::rewind_to_height` rewinds the *entire* wallet across pools/accounts; the cross-pool checkpoint invariant may be exactly why the whole-wallet primitive is needed rather than the account-scoped one.)

**Reproducer.** `qa/rpc-tests/wallet_import_export_key.py` on branch `export-import-key-test` (integration-tests #76). Run with:

```
ZEBRAD=<zebra main>/target/release/zebrad \
ZALLET=<this branch launcher, with zallet-zaino beside it> \
./qa/pull-tester/rpc-tests.py wallet_import_export_key
```

Minimal sequence against a running regtest wallet synced to tip (NU6.2 active), with a confirmed Sapling note at an address whose key the wallet doesn't yet hold:
1. `z_importkey(<sapling key>, "whenkeyisnew", <height>)` — imports, scans forward (fine).
2. `z_importkey(<same key>, "yes", 1)` — the rewind branch → `Data DB is corrupted: Sapling and Orchard should have the same checkpoints`.

**What already passes against this branch** (so the rest of the fix is validated): the `#579` path (fresh import with an early birthday finds the historical note), all the basic import/export/isolation/input-validation assertions, and — separately — the confirmation that spending from an imported key currently fails with `Invalid from address, no payment source found for address.` (tracked as a distinct issue, not part of this PR).

Happy to iterate on the test once the rewind is adjusted here.

oxarbitrage:
Confirmed: rebuilt `zallet-zaino` from `3e96317` (the `truncate_to_height` revision) and re-ran `wallet_import_export_key` (zcash/integration-tests#76) against it + `zebrad` from `zcash/zebra` `main` — **all three scenarios pass** (green, 196s). The checkpoint corruption from the earlier `rewind_to_chain_state` revision is gone. Added `ZIT-Revision: export-import-key-test` to the description so interop CI runs these end-to-end.

oxarbitrage:
Interop status update: the `ZIT-Revision` is working — the dispatch now correctly picks up the test branch (`TEST_SHA: export-import-key-test`). But the `zallet-interop-request` run currently aborts in the **Define CI matrix** job for a reason unrelated to this PR.

Root cause is on the integration-tests side: after the `zcash/wallet` → `zcash/zallet` rename, the interop's repo-id mapping still pointed at `zcash/wallet`, so the App-token lookup (`repos/zcash/wallet/installation`) 404s before anything builds. Fixed by zcash/integration-tests#161 (one-liner).

Once #161 merges I'll re-trigger this PR's interop so it builds `3e96317` and runs the three rescan-on-import scenarios end-to-end. They already pass locally against this branch + `zebrad` from zebra `main`.

pacu:
Scheduling context, and a rebase request.

Zallet has moved to a **weekly release train** — a cut every Tuesday, shipping whatever is
merged and green (#781). This PR is targeted at the **beta.6 cut, Tue 2026-09-15**.

It **conflicts with `main`** at the moment. It is a single-file change, so the rebase should be
cheap — would you mind? Nothing can be reviewed until it is mergeable, and it is otherwise one
of the more straightforward items in the queue.

Thanks for your patience on this one; it has been open a while, and the queue is being worked
through in dependency order rather than age order.

pacu:
This conflicts with `main` — could you rebase? It can't be reviewed until it's mergeable.

oxarbitrage:
Rebased onto `main` and force-pushed. Two things worth flagging, since this needed more than a rebase:

**1. The conflict was semantic.** `e25dad2` ("Make z_importkey import the account and store its key atomically") replaced the `if is_new_key { … }` block this patch hung its `else if` on, and made `reload_keys()` unconditional. The fix is re-applied against the new structure rather than textually merged.

**2. The librustzcash pin moved (`29c7fb2` → `1f6bb20`), which invalidated this PR's approach.** The previous revision used `truncate_to_height` specifically to avoid a `rewind_to_chain_state` bug that corrupted the wallet DB on a deep rewind. That bug is fixed upstream (it now snaps to a `window_floor` across all pools), and `truncate_to_height` turns out to be unable to lower an account birthday — which is precisely what #578 asks for, and which also made the documented default `startHeight=0` fail outright. So this now uses `rewind_to_chain_state`, the same path the new-key import already takes. Details in the updated description.

I also want to correct the record: **the "IT#76 all scenarios green" claim in the old description was stale** — it predated the pin move. I re-ran it. This PR's own scenario passes (verified with instrumented `z_getbalances`: the re-imported account reports the recovered note within ~20s, birthday correctly lowered). The full run still fails in a later phase, but that failure **reproduces identically on a build of the previous revision**, and is unchanged at 450s and 1200s — it matches #588, a pre-existing sync bug where a `Historic` range in the `(tip - 100, tip]` band is serviced by neither `recover_history` nor `steady_state`.

So: ready for review, with the caveat that a fully green interop run depends on #588.

PR #583, opened 2026-07-11
Move the librustzcash and age pins to the latest git revisions

dannywillems:
## Summary

Moves the `[patch.crates-io]` pins in all three workspaces (root,
`backends/zebra`, `backends/zaino`) to the latest upstream git revisions and
bumps the crate version requirements to their released numbers. The librustzcash
family and `age` stay on git pins because the wallet uses APIs that are not yet
published to crates.io. `orchard` is now unpinned (its 0.15.0 release is on
crates.io).

Both the zebra and zaino backends build, and `utils/check-lockstep.sh` passes.

## Pending upstream releases (tracking)

This section is updated as each dependency releases; a pin is dropped when its
release lands.

- [ ] **`zcash_client_backend` 0.24.0** (librustzcash): the sync component uses
      the `sync-decryptor` batch decryptor (`sync::decryptor`), only on
      librustzcash `main`. This blocks the entire librustzcash family pin; when
      0.24.0 ships, drop every librustzcash entry and use the crates.io releases.
      Currently pinned to `main` `a97a3d5f`.
- [ ] **`age`** (str4d/rage): the keystore uses `age::encrypted::EncryptedIdentity`
      and `Send + Sync` identities from `IdentityFile::into_identities`, which the
      crates.io 0.11.3 release lacks (it renamed the type and dropped the
      `Send + Sync` bound). Un-pin `age` and drop the `=0.11.1` requirement once a
      release carries these APIs. Currently pinned to `92f437bc`.
- [ ] **`zaino`** (zingolabs/zaino): NU6.3 / Ironwood `get_treestate` support.
      Un-pin once it lands in a release. Currently pinned to `dev` `8048bf8d`.
- [ ] **`zewif` / `zewif-zcashd`** (zcash): first crates.io release; un-pin then.

## Already released (done)

- [x] **`orchard` 0.15.0**: released; librustzcash `main` builds against it, so
      the orchard patch was removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,7 +1224,7 @@ checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
 [[package]]
 name = "equihash"
 version = "0.3.0"
-source = "git+https://github.com/zcash/librustzcash.git?rev=29c7fb28889a69df15b44fc9465cf4a464f766e9#29c7fb28889a69df15b44fc9465cf4a464f766e9"
+source = "git+https://github.com/zcash/librustzcash.git?rev=a97a3d5f46d096b94ceb71271c7d38f20af4e1f1#a97a3d5f46d096b94ceb71271c7d38f20af4e1f1"
 dependencies = [
  "blake2b_simd",
  "corez",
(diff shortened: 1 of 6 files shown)


nuttycom:
Subsumed by #589. I don't know when `rage` will get a release so we should just pare back functionality to what is available with 0.11.3.

PR #584, opened 2026-07-11
fix(keystore): await aborted relock task in lock()

nullcopy:
Subsumes https://github.com/zcash/zallet/pull/468

KeyStore::lock() aborted the background relock task but did not await its completion, unlike unlock(). A rapid walletpassphrase after walletlock could therefore leave the wallet locked despite reporting success when the stale task cleared identities after unlock() repopulated them.

Fixes GHSA-8cfr-hmvc-2333.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,9 @@ be considered breaking changes.
   embedding them in the HTTP URL.
 
 ### Fixed
+- `walletlock` now awaits aborted relock tasks before returning, preventing a race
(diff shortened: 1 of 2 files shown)


zcash-harry:
Should we make a test for lock/unlock cycles?

nuttycom (approved):
utACK

ouicate:
Great, thanks for getting this fixed and merged. 

Could we go ahead and publish the advisory now that the fix has landed in `main`? Let me know if any additional details are needed from my side. @nullcopy @nuttycom

Issue #384, opened 2026-02-24
Implement "legacy `zcashd` transparent pool of funds" semantics


str4d:
# Background

`zcashd` had two kinds of legacy balance semantics:

* The transparent JSON-RPC methods inherited from Bitcoin Core treated all spendable funds in the wallet as being part of a single pool of funds. RPCs like `sendmany` didn't allow the caller to specify which transparent addresses to spend funds from, and RPCs like `getbalance` similarly computed a balance across all transparent addresses returned from `getaddress`.
* The early shielded JSON-RPC methods added for Sprout treated every address as a separate pool of funds, because for Sprout there was a 1:1 relationship between addresses and spend authority. RPCs like `z_sendmany` only spent funds that were sent to the specified addressed, and RPCs like `z_getbalance` similarly computed a separate balance for each address (which became complex and unintuitive with the introduction of Sapling diversified addresses).

With the advent of Unified Addresses and HD-derived spending keys, `zcashd` gained its modern balance semantics: each full viewing key in the wallet is a separate pool of funds, and treated as a separate "account". These are the semantics used throughout Zallet, and that should be used by everyone going forward. They are also incompatible with various legacy JSON-RPC methods that were deprecated in `zcashd`, as well as some fields of general RPC methods; these methods and fields are unavailable in Zallet by default.

This issue is about implementing the first of the above legacy behaviours. (We don't need to implement the second, because if we ever added Sprout support then we'd represent each Sprout address as a separate account.)

Due to how the legacy transparent semantics in particular were defined by Bitcoin Core, this can only be done for a single `zcashd` wallet at a time. Given that every `zcashd` wallet in production in 2025 had a single mnemonic seed phrase in its wallet, we use its ZIP 32 seed fingerprint as the `zcashd` wallet identifier in this setting.

# What to implement

When a `zcashd` wallet is imported via the `migrate-zcashd-wallet` CLI command:

* If the imported wallet is old enough that it doesn't have an HD seed (i.e. pre-Sapling), then a new mnemonic should be generated, and used for subsequent steps.
  * This is effectively behaving as if the imported wallet had first been loaded into a modern `zcashd`, with the caveat that the non-mnemonic seed never existed.
* The legacy account created from that seed as part of the import (from which all legacy Sapling addresses are generated) should have its `key_info` field set to something identifying the account as imported from `zcashd`.
[...]

PR #585, opened 2026-07-11
Spend transparent funds in z_sendmany

dannywillems:
Stacked on #586 (Docker fix), which is unrelated but must land first for CI to be green. Base auto-retargets to `main` on its merge.

## Problem

`z_sendmany` passed `SpendPolicy::default()` to `propose_transfer`. That policy permits the shielded pools and **no transparent spending**:

```rust
impl Default for SpendPolicy {
    fn default() -> Self {
        Self::shielded_pools([Sapling, Orchard, Ironwood])   // transparent: None
    }
}
```

With `transparent()` returning `None`, `GreedyInputSelector` skips the transparent gather entirely (`=> Vec::new()`), so **no proposal could ever contain a transparent input**. Transparent-to-transparent transfers were impossible, and the `AllowFullyTransparent` privacy policy was unreachable — `enforce_privacy_policy` already implemented it correctly, but no proposal it could accept could be built.

The `// TODO: Once zcash_client_backend supports spending transparent coins arbitrarily` comment was stale. The pinned librustzcash already provides `SpendPolicy::with_transparent`, `TransparentSpendPolicy`, `CoinbasePolicy`, and `TransparentChangePolicy`. **No dependency bump.**

## Solution

Derive the spend policy from the source address instead of hardcoding the default:

```rust
let spend_policy = match &address {
    Address::Transparent(taddr) => SpendPolicy::shielded_pools([])
        .with_transparent(TransparentSpendPolicy::from_one_address(*taddr)),
    _ => SpendPolicy::default(),
};
```

- A **bare transparent `fromaddress`** permits transparent spending restricted to that one address's UTXOs. A shielded send can never silently reach into transparent funds, and the named address is not linked to the account's other transparent receivers.
- **Every other source is unchanged** (`SpendPolicy::default()`), so shielded behaviour is untouched.
- **Coinbase stays unspendable** (`CoinbasePolicy::NonCoinbase` is the default): consensus requires it be spent to a single shielded output, which remains `z_shieldcoinbase`'s job. A transparent spend needs a non-coinbase UTXO.
- **`ANY_TADDR` remains unsupported** (#384) and is explicitly out of scope.

The privacy policy is deliberately *not* consulted during selection: the selector returns its best proposal, and `enforce_privacy_policy` rejects it afterwards. Resulting lattice, all enforced and all tested:

| Flow | Required policy |
| --- | --- |
| transparent source -> shielded recipient | `AllowRevealedSenders` |
| transparent source -> transparent recipient/change | `AllowFullyTransparent` |

### Transparent change

Transparent change is permitted **only when the spend policy carries a transparent component**:

```rust
let transparent_change_policy = match spend_policy.transparent() {
    Some(_) => TransparentChangePolicy::TransparentChangeAllowed,
    None => TransparentChangePolicy::ShieldChange,
};
```

This keeps a fully transparent send transparent end to end rather than sweeping its change into a shielded pool. The change strategy independently enforces the same thing (it emits transparent change only when the net flows are fully transparent, i.e. no shielded input or output at all) — but that is *its* invariant, not ours, so we state ours locally rather than depend on it. **A shielded send cannot acquire a transparent change output by this route.** Change lands at an internal-scope (BIP 44) address, freshly reserved per transaction, so consecutive sends cannot be linked through a reused change address.

`fallback_change_pool` stays `ShieldedPool::Orchard`, deliberately. It is consulted only when the flows are fully transparent, and `select_change_pool` promotes it to Ironwood itself once NU6.3 is active (the turnstile forbids value entering Orchard), keyed on the transaction's **target height** — which is not known where the change strategy is constructed. Naming `Ironwood` outright would send change to a pool that does not exist yet on a pre-NU6.3 chain.

## Three latent defects this surfaced

All were unreachable while transparent spending was impossible. **Two panicked the wallet process** rather than returning an error, because the trait methods are *defaulted* to `unimplemented!()` — so omitting them compiles cleanly and only fails at run time.

1. `get_account_for_address` compared the source against `list_addresses()`, which yields **unified** addresses. A bare transparent receiver never compares equal to the UA it belongs to, so a `taddr` `fromaddress` always failed with `Invalid from address, no payment source found for address.` Now resolved via `find_account_for_address`, which maps an address back through its receivers.
2. `DbConnection` did not implement `InputSource::select_spendable_transparent_outputs` — **panic** on selecting any transparent input.
3. `DbConnection` did not implement `WalletWrite::reserve_next_n_internal_addresses` — **panic** on producing transparent change.

Note for reviewers: **`z_sendfromaccount` and `z_proposetransaction` sit on the same two methods.** Landing this first means they will not rediscover these panics.

## Testing

Covered by zcash/integration-tests#163, which adds `wallet_transparent_spend.py` and reactivates `wallet_changeaddresses.py` and `regtest_signrawtransaction.py` — both were disabled pending the account/UA migration, and both exist to exercise transparent spending. The trailer at the bottom of this description points this PR's CI at that branch.

Verified locally on both sides of the NU6.3 boundary:

- `wallet_transparent_spend.py` (NU6.3 off) — 4 ZEC sent t-to-t; asserts the tx is fully transparent (no shielded components), the recipient balance on-chain, and change at a fresh internal address. Pins **both** policy rejections: `AllowRevealedSenders` and `AllowRevealedRecipients` are each insufficient alone, and they fail via *different* checks (the recipient check runs on the request; the fully-transparent check runs on the built proposal).
- `wallet_changeaddresses.py` (NU6.3 off) — two consecutive transparent sends use different change addresses.
- `regtest_signrawtransaction.py` (NU6.3 off) — a transparent input signs under the correct consensus branch id.
- `wallet_ironwood.py` (NU6.3 **on**) and `wallet_z_shieldcoinbase.py` — regressions, unaffected.

Clippy clean; 189/189 `zallet-core` unit tests pass.

ZIT-Revision: t-t-tests

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/z_send_many.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_send_many.rs
@@ -13,11 +13,14 @@ use zcash_client_backend::{
         Account,
         wallet::{
             ConfirmationsPolicy, create_proposed_transactions,
-            input_selection::{GreedyInputSelector, SpendPolicy},
+            input_selection::{GreedyInputSelector, SpendPolicy, TransparentSpendPolicy},
             propose_transfer,
         },
     },
-    fees::{DustOutputPolicy, StandardFeeRule, standard::MultiOutputChangeStrategy},
+    fees::{
+        DustOutputPolicy, StandardFeeRule, TransparentChangePolicy,
+        standard::MultiOutputChangeStrategy,
+    },
     wallet::OvkPolicy,
 };
 use zcash_client_sqlite::ReceivedNoteId;
@@ -64,6 +67,45 @@ pub(super) const PARAM_FEE_DESC: &str = "If set, it must be null.";
 pub(super) const PARAM_PRIVACY_POLICY_DESC: &str =
     "Policy for what information leakage is acceptable.";
 
+/// The sources of funds a transfer from `source` may draw upon.
+///
+/// Spending from a bare transparent address draws only on that address's UTXOs: the funds are
+/// already public, and confining selection to the named address avoids linking it to the
+/// account's other transparent receivers. Every other source stays shielded-only, so a
+/// shielded send can never silently reach into transparent funds.
+///
+/// Coinbase UTXOs are excluded: `TransparentSpendPolicy` defaults to
+/// `CoinbasePolicy::NonCoinbase`, and consensus requires coinbase to be spent to a single
+/// shielded output, which is `z_shieldcoinbase`'s job.
+///
+/// The privacy policy deliberately does not narrow this: the selector returns its best
+/// proposal, and `enforce_privacy_policy` rejects it afterwards if it leaks more than the
+/// caller permitted.
+fn spend_policy_for(source: &Address) -> SpendPolicy {
+    match source {
+        Address::Transparent(taddr) => SpendPolicy::shielded_pools([])
+            .with_transparent(TransparentSpendPolicy::from_one_address(*taddr)),
+        _ => SpendPolicy::default(),
+    }
+}
+
+/// Whether change may be returned to the transparent pool.
+///
+/// Permitted exactly when `spend_policy` can spend transparent funds in the first place, which
+/// keeps a fully transparent send transparent end to end rather than sweeping its change into a
+/// shielded pool. A shielded send therefore cannot acquire a transparent change output by this
+/// route.
+///
+/// The change strategy independently enforces the same thing (it emits transparent change only
+/// when the transaction's net flows are fully transparent, i.e. it has no shielded input or
+/// output at all), but that is its invariant, not ours.
+fn transparent_change_policy_for(spend_policy: &SpendPolicy) -> TransparentChangePolicy {
+    match spend_policy.transparent() {
+        Some(_) => TransparentChangePolicy::TransparentChangeAllowed,
+        None => TransparentChangePolicy::ShieldChange,
+    }
+}
+
 #[allow(clippy::too_many_arguments)]
 pub(crate) async fn call<C: Chain>(
     mut wallet: DbHandle,
@@ -88,7 +130,7 @@ pub(crate) async fn call<C: Chain>(
 
     let request = build_request(&amounts)?;
 
-    let account = match fromaddress.as_str() {
+    let (account, spend_policy) = match fromaddress.as_str() {
         // Select from the legacy transparent address pool.
         // TODO: Support this if we're going to. https://github.com/zcash/zallet/issues/138
         "ANY_TADDR" => Err(LegacyCode::WalletAccountsUnsupported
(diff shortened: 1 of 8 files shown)


zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -198,19 +222,53 @@ pub(crate) async fn call<C: Chain>(
...
+    let transparent_change_policy = match spend_policy.transparent() {
+        Some(_) => TransparentChangePolicy::TransparentChangeAllowed,
+        None => TransparentChangePolicy::ShieldChange,
+    };
+
+    // Where shielded change goes when the transaction has no shielded flows to infer a pool
+    // from. A transaction that does have shielded flows ignores this and keeps its change in
+    // the pool it is already using.
+    //
+    // This stays Orchard rather than Ironwood: the change strategy promotes it to Ironwood
```
dannywillems, line 274: Ah, smart.

zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -88,7 +130,7 @@ pub(crate) async fn call<C: Chain>(
 
     let request = build_request(&amounts)?;
 
-    let account = match fromaddress.as_str() {
+    let (account, spend_policy) = match fromaddress.as_str() {
         // Select from the legacy transparent address pool.
         // TODO: Support this if we're going to. https://github.com/zcash/zallet/issues/138
         "ANY_TADDR" => Err(LegacyCode::WalletAccountsUnsupported
```
nuttycom: It should now be possible to support `ANY_TADDR`; there is a specific transparent input selection policy that this can now be implemented using.

zallet-core/src/components/json_rpc/methods/z_send_many.rs
nuttycom: We have the wallet handle here, so we could just get the chain tip height and use that to determine the fallback change pool. I'm not sure what the behavior is if the fallback change pool is Orchard after nu6.3 activation.

dannywillems: > I'm not sure what the behavior is if the fallback change pool is Orchard after nu6.3 activation.

It goes to Ironwood. This is in librustzcash. I'm writing a test for this.

dannywillems: See https://github.com/zcash/librustzcash/pull/2608/. This confirms it goes to ironwood after activation, and stays in orchard before activation. The logic already happens in librustzcash.

nuttycom: Maybe a little fragile in the future but fine for now.

nuttycom (approved):
utACK, 🎉

PR #589, opened 2026-07-11
Update to zewif 1.0.0-rc.2; support passphrase-encrypted wallet.dat migration

nuttycom:
This updates zallet to the released ZeWIF stack — `zewif 1.0.0-rc.2` and `zewif-zcashd 0.1.0-rc.2` from crates.io — and adds interactive passphrase support to `migrate-zcashd-wallet` for encrypted `wallet.dat` files.

## Dependency changes

- `zewif` and `zewif-zcashd` move from git pins to the crates.io releases. zewif 1.0.0-rc.2's wire format flattens the tagged-union encoding to `[variant-id, body?]` (documents produced against earlier revisions do not decode) and moves the codec to minicbor 2.
- librustzcash is pinned to the rev that migrates it to the released zewif (zcash/librustzcash#2607); `orchard` moves to the released 0.15.0, dropping its git pin.
- The rage pin moves to current rage main (`d28f10ee`): age 0.11.3 plus the unreleased APIs the keystore relies on — `EncryptedIdentity`, and `Send + Sync` bounds on `IdentityFile::into_identities` — which the 0.11.x maintenance releases exclude. The pin is temporary: str4d expects to publish a rage 0.12 release shortly, at which point it should be replaced by the crates.io release.
- Both backend workspaces are synced to the root pins (librustzcash via `utils/sync-librustzcash.sh`, plus the rage-main pin); `check-lockstep.sh` reports the three lockfiles consistent.

## Encrypted wallet.dat migration

`migrate-zcashd-wallet` now handles passphrase-encrypted zcashd wallets: parsing first runs under the default reject policy, and when the dump reports encrypted key material the command interactively prompts for the wallet passphrase (up to three attempts on a wrong passphrase) and decrypts via `zewif-zcashd`'s `EncryptedKeyPolicy::Decrypt`. The passphrase is held as a `SecretVec` and zeroized on drop. zewif-zcashd's `migrate_to_zewif` also gained a `regtest_activations` parameter; zallet passes `None` since regtest migrations are rejected earlier by `check_network`.

## Verification

- Root workspace: 214 tests pass with `--all-features`; clippy and rustfmt clean.
- Both backends compile with `--all-features` against the synced pins.
- `utils/check-lockstep.sh`: 14 crates + 4 package versions consistent across the 3 lockfiles.

## Known follow-up

- Replace the rage-main pin with the rage 0.12 release once published.
- `cargo vet` / `cargo deny` have not yet been reconciled with the updated dependency graph.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ be considered breaking changes.
 
 ### Added
 
+- `migrate-zcashd-wallet` now supports passphrase-encrypted zcashd wallets:
+  when the wallet's key material is encrypted, the command interactively
+  requests the wallet passphrase and decrypts the key material during the
+  migration.
 - The `zebra` chain backend (and the `zaino` backend's read-state-service mode)
   now support regtest. The read-state service builds a Zebra Regtest network
   whose network-upgrade activation heights mirror the wallet's configured
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ dependencies = [
  "p256",
  "pin-project",
  "pinentry",
- "rand 0.8.6",
+ "rand 0.8.7",
  "rpassword",
  "rust-embed",
  "scrypt",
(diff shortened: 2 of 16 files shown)


nullcopy (approved):
utACK 7c3427adf963a6e0c52caf12d79f259786ee9dfa

PR #598, opened 2026-07-13
fix(sync): don't shut down the wallet when a transaction data request fails

### LINKED ISSUE #599: Wallet crash-loops permanently when a data-requested transaction is reorged out
**Version:** zallet v0.1.0-beta.1, zebrad 6.0.0, testnet (post-NU6.3)

## Summary

When a transaction the wallet has a pending `TransactionDataRequest` for is reorged out
of the best chain, the data-requests sync task aborts the whole process. Because the
failed request persists in the wallet DB and is re-serviced first on every startup, the
wallet enters a permanent crash-loop (systemd `start-limit-hit` after a few restarts).
The only recovery we found is a manual `repair truncate-wallet` to a height **below**
the reorged block.

## Reproduction

1. zebrad + zallet (zaino backend), `miner_address` = wallet-owned t-addr.
2. Mine blocks to the wallet; keep mining until one of your blocks is reorged out
   (easy on testnet; on mainnet this is any orphaned block).
3. The wallet queues `TransactionDataRequest::GetStatus`/`Enhancement` for the coinbase
   txid while the block is in the best chain; after the reorg, zebra's
   `getrawtransaction` returns RPC `-5`.

## Observed behavior

```
INFO zallet_core::commands::start: Wallet data-requests sync task exited
wallet_sync_result=Err(Error(Context { kind: Sync, source: Some(Chain(Backend(ChainIndexError {
  kind: InternalServerError,
  message: "critical error in backing block source: could not fetch transaction data:
            unexpected error response from server: RPC Error (code: -5):
            No such mempool or main chain transaction" })))
INFO zallet_core::commands::start: Exiting Zallet because an ongoing task exited; asking other tasks to stop
...
zallet.service: Failed with result 'start-limit-hit'.
```

Restarts do not help: the same persisted request fails first every time.

s1ckoleg:
Fixes zcash/zallet#599.

When servicing a `GetStatus`/`Enhancement` transaction data request fails with a backend error — e.g. the transaction was reorged away and the validator answers `RPC -5: No such mempool or main chain transaction` — the data-requests sync task exits and takes the whole wallet process down. The request is regenerated from the wallet database on every startup, so one stale request crash-loops the wallet until systemd gives up (`start-limit-hit`).

This change logs the failure and leaves the request for a later sync iteration instead, matching how `TransactionsInvolvingAddress` servicing errors are already handled. Once the chain index catches up with the reorg, the request resolves through the existing `None → TxidNotRecognized` path. Wallet database errors still abort as before.

Tested with `cargo check -p zallet-core` and `cargo fmt`; the failure mode was reproduced live on testnet (v0.1.0-beta.1, zebrad 6.0.0) — details in the linked issue.

### DIFF
--- a/zallet-core/src/components/sync.rs
+++ b/zallet-core/src/components/sync.rs
@@ -886,12 +886,10 @@ async fn data_requests<C: Chain>(
                     }
 
                     info!("Getting status of {txid}");
-                    let status = chain_view
-                        .get_transaction_status(txid)
-                        .await
-                        .map_err(SyncError::Chain)?;
-
-                    db_data.set_transaction_status(txid, status)?;
+                    match chain_view.get_transaction_status(txid).await {
+                        Ok(status) => db_data.set_transaction_status(txid, status)?,
+                        Err(e) => warn!("Failed to get status of {txid} (will retry): {e}"),
+                    }
                 }
                 TransactionDataRequest::Enhancement(txid) => {
                     if txid.is_null() {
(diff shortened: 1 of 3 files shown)


zallet-core/src/components/sync.rs
```diff
@@ -886,36 +886,36 @@ async fn data_requests<C: Chain>(
...
                     info!("Getting status of {txid}");
-                    let status = chain_view
-                        .get_transaction_status(txid)
-                        .await
-                        .map_err(SyncError::Chain)?;
-
-                    db_data.set_transaction_status(txid, status)?;
+                    match chain_view.get_transaction_status(txid).await {
+                        Ok(status) => db_data.set_transaction_status(txid, status)?,
+                        Err(e) => warn!("Failed to get status of {txid} (will retry): {e}"),
```
pacu, line 894: I think that we want to differentiate the kind of error. here. The previous code was incorrect because it treated all the `ChainError` values as if they were as bad, and this code does that in by assuming all the possible `ChainError` values that can be returned can be recovered from.  Ultimately what we want here is to fail when `ChainErrod::InvalidData` is returned and retry on the rest (we don't have much fine-grained errors here).

zallet-core/src/components/sync.rs
```diff
@@ -886,36 +886,36 @@ async fn data_requests<C: Chain>(
...
+                                tx.mined_height(),
+                            )?;
+                        }
+                        Ok(None) => {
+                            db_data.set_transaction_status(
+                                txid,
+                                TransactionStatus::TxidNotRecognized,
+                            )?;
+                        }
+                        Err(e) => warn!("Failed to enhance {txid} (will retry): {e}"),
```
pacu, line 924: Same as above

pacu (changes requested):

zallet-core/src/components/sync.rs
s1ckoleg: Done in c57a2e5 — `ChainError::InvalidData` now propagates (fatal) in both arms; `Unavailable`/`Backend` are logged and retried.

zallet-core/src/components/sync.rs
s1ckoleg: Done in c57a2e5 — `ChainError::InvalidData` now propagates (fatal) in both arms; `Unavailable`/`Backend` are logged and retried.

pacu (approved):
utACK

nuttycom (approved):
utACK 76eea49e88723bbd8588c2bbf79905e647eda34d

nuttycom (approved):
re-utACK c115f17cde6fead96df650f124dc3b7726349c19

PR #614, opened 2026-07-15
rpc: Implement z_exportviewingkey for Sapling

### LINKED ISSUE #70: rpc: Implement `z_exportviewingkey`

nullcopy:
Subsumes https://github.com/zcash/zallet/pull/461
Closes #70
Closes COR-255

Implements z_exportviewingkey, which returns the bech32-encoded Sapling extended full viewing key for a given Sapling payment address. This is the export counterpart needed to support z_importviewingkey (PR #431) and its integration test (zcash/integration-tests#99).

The address lookup scans each account's UFVK Sapling component via `decrypt_diversifier`, since bare Sapling receivers derived from unified accounts aren't found by the standard `get_account_for_address` path. Wallet unlock is required for seed access to re-derive the EFVK. Imported view-only accounts are rejected (the stored `DiversifiableFullViewingKey` can't reconstruct the chain-code-bearing `ExtendedFullViewingKey`).

Integration tests can be found here: https://github.com/zcash/integration-tests/pull/115

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Added
+
+- RPC methods:
+  - `z_exportviewingkey`. In addition to the `zcashd` behaviour (exporting the
+    Sapling extended full viewing key for a Sapling address), it accepts
+    unified addresses (returning the account's unified full viewing key), and
+    an optional `ivk` argument that exports the account's unified incoming
+    viewing key instead.
+
 ### Changed
 
 - `zallet rpc help` is now answered locally instead of being sent to the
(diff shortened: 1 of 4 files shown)


nuttycom (changes requested):
Overall looks good; please ensure that the book has been updated to reflect `z_exportviewingkey` support.

I also think that we should modify the semantics of this method to support export of ufvks and uivks; we can add an optional argument to specify when an IVK is requested.

schell:
utACK 89dc36ac571f2340feddbf4326b70ea55fcc666c

mnm458:
Once this merges, I will update https://github.com/zcash/zallet/pull/431 and its integration tests

nuttycom (approved):
utACK

PR #619, opened 2026-07-15
rpc-cli: answer `help` locally without a wallet or running server

### LINKED ISSUE #550: `zallet rpc help` should not required a wallet state to show help
when running `zallet rpc help` it shouldn't be necessary to have a config file and an initialized wallet to access the help menu. 

```
 ./target/release/zallet rpc help  
2026-07-06T10:34:31.747783Z  INFO zallet::commands: Loading configuration config=/Users/pacu/.zallet/zallet.toml
2026-07-06T10:34:31.748406Z  WARN zallet::commands::rpc_cli: Failed to read cookie file: Failed to initialize zallet: No such file or directory (os error 2)

[ Tell us: https://github.com/zcash/zallet/issues                      ]
JSON-RPC request failed: client error (Connect)

[ Tell us: https://github.com/zcash/zallet/issues    
```

zcash-harry:
Closes #550.

The `help` RPC response is generated entirely from static method metadata (the `METHODS` map in `openrpc.rs`), so `zallet rpc help` can print it directly instead of requiring a config file, an initialized wallet, and a running JSON-RPC server just to list the available commands.

Changes:
- Split the text generation out of the `help` RPC handler into `help::text()`; the RPC `call()` now wraps it, so server-side behavior is unchanged and the CLI output is guaranteed to match it.
- `RpcCliCmd::run` short-circuits `help` before reading credentials or building the HTTP client, guarded with `#[cfg(zallet_build = "wallet")]` to match the cfg on the `help` module.
- The command argument may now be passed bare (`zallet rpc help getwalletinfo`) in addition to the JSON-quoted form the HTTP path requires; extra parameters still error as before.

Verified against a freshly created empty datadir (no config, no wallet, no server): `rpc help`, `rpc help getwalletinfo` (bare and quoted), unknown-command, and too-many-params paths all behave, with no cookie warning or connection error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Changed
+
+- `zallet rpc help` is now answered locally instead of being sent to the
+  wallet's JSON-RPC server, so it no longer requires a config file, an
+  initialized wallet, or a running `zallet start`. The command argument may
+  now be passed bare (`zallet rpc help getwalletinfo`) in addition to the
+  JSON-quoted form.
+
 ## [0.1.0-beta.1] - 2026-07-12
 
 ### Added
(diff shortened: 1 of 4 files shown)


nuttycom (approved):
utACK

PR #621, opened 2026-07-16
Make age recipient initialization atomic

### LINKED ISSUE #616: initialize_recipients inserts age recipients non-atomically. Should be atomic.
initialize_recipients inserts age recipients in a loop with no transaction, so each INSERT autocommits (keystore.rs:423-441). If a write fails partway (disk full, I/O error, crash), a subset commits. The one-shot guard (keystore.rs:413-419) then refuses every retry because the table is non-empty, leaving the wallet stuck with an incomplete set. All later key material is encrypted only to that subset, so a missing offline backup identity can never decrypt the wallet - surfacing only at recovery time. 

You could consider wrapping the loop in conn.transaction() + commit like store_encrypted_standalone_transparent_keys already does (keystore.rs:644-668), making init all-or-nothing.

nullcopy:
`initialize_recipients` previously inserted age recipients in a loop with no transaction, so each INSERT autocommitted. A failure partway through (disk full, I/O error, crash) could commit a partial recipient set, after which the one-shot emptiness guard would refuse every retry, permanently wedging wallet initialization with key material encrypted to an incomplete set of recipients.

Wrap the emptiness guard and the insert loop in a single explicit transaction, following the pattern used by
`store_encrypted_standalone_transparent_keys`. The guard is now an in-transaction `SELECT COUNT(*)` on the recipients table (equivalent to the previous `maybe_encryptor().is_empty()` check, which used a separate connection and could not participate in the transaction), which also closes the check-then-insert race.

Closes https://github.com/zcash/zallet/issues/616
[COR-1579](https://linear.app/zodl/issue/COR-1579)

### DIFF
--- a/zallet-core/src/components/keystore.rs
+++ b/zallet-core/src/components/keystore.rs
@@ -410,31 +410,51 @@ impl KeyStore {
         &self,
         recipient_strings: Vec<String>,
     ) -> Result<(), Error> {
-        // If the wallet has any existing recipients, fail (we would instead need to
-        // re-encrypt the wallet).
-        if !self.maybe_encryptor().await?.is_empty() {
-            return Err(ErrorKind::Generic
-                .context(fl!("err-keystore-already-initialized"))
-                .into());
-        }
-
         let now = ::time::OffsetDateTime::now_utc();
 
         self.with_db_mut(|conn, _| {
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK

PR #622, opened 2026-07-16
Make z_importkey perform keystore and wallet-db writes in lock-step

### LINKED ISSUE #615: z_importkey should perform keystore and wallet-db writes in lock-step
z_importkey writes the keystore first, then the wallet-DB account, in two separate transactions with no rollback (import_key.rs:103-154). If a step after the keystore write fails (e.g. fetch_account_birthday or import_account_ufvk), the call returns an error but the keystore row persists. Since z_exportkey reads only the keystore, the "failed" key is exportable while the wallet never scans it (zero balance / funds look lost until a re-import happens). As a fix, you may consider wrapping both writes in one transaction (same DB, so feasible), or writing the keystore last so failures fail closed.

nullcopy:
Depends on https://github.com/zcash/librustzcash/pull/2639
Closes https://github.com/zcash/zallet/issues/615
[COR-1578](https://linear.app/zodl/issue/COR-1578)

### DIFF
--- a/backends/zaino/Cargo.toml
+++ b/backends/zaino/Cargo.toml
@@ -78,6 +78,25 @@ tempfile = "3"
 # rage 0.12 ships.
 age = { git = "https://github.com/str4d/rage.git", rev = "d28f10ee776ce7391f7065695997f998631b113a" }
 
+# Temporary patch for `WalletDb::transactionally_with_extension` (adds atomic
+# z_importkey account+key writes), merged upstream but not yet released. The API
+# lives in zcash_client_sqlite, but the repo is a workspace: pinning only that
+# crate pulls its path-dep siblings from git while their crates.io copies remain,
+# duplicating the librustzcash stack, so the whole stack is pinned to one rev.
+# Replace with the crates.io releases once the next zcash_client_sqlite release
+# ships.
+equihash = { git = "https://github.com/zcash/librustzcash.git", rev = "d09d1f390f2e4e3d8978bedfe84dd18aeceef008" }
+f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "d09d1f390f2e4e3d8978bedfe84dd18aeceef008" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "d09d1f390f2e4e3d8978bedfe84dd18aeceef008" }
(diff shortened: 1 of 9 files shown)


nullcopy:
Self-ack ba2116805521fa445f17a8bc303062a4840ccfdf

backends/zaino/Cargo.toml
```diff
@@ -78,6 +78,23 @@ tempfile = "3"
...
+zcash_address = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_client_backend = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_client_sqlite = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_encoding = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_keys = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_primitives = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_protocol = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zcash_transparent = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
+zip321 = { git = "https://github.com/nullcopy/librustzcash.git", rev = "3adfaac98eac9c46a3875b50a8fee7b3e152853d" }
```
nuttycom: This shouldn't be necessary now that the librustzcash PRs have merged, right?

nullcopy: I can update the pin to LRZ main if you prefer, but since the deps are pointing to the latest lrz crate release, we do need some crate-patch... unless I'm mistaken

nullcopy: done

nuttycom (approved):
utACK

PR #623, opened 2026-07-16
zallet-core: add zcash_ironwood_migration_backend dependency scaffold

dannywillems:
Note: I would be in favor to add a flag `experimental` to not encourage users to use some RPCs.

------

Wires the new backend-agnostic `zcash_ironwood_migration_backend` crate (the Orchard -> Ironwood migration engine) into zallet as a dependency scaffold, independent of the crate's still-evolving API.

- Adds `zcash_ironwood_migration_backend` pinned to the `feat/ironwood-denominations` git branch of `zcash/librustzcash` (the crate is unpublished; to be repointed at a release later).
- Wires it through `zallet-core` via a thin `zallet_core::migrate` scaffold module (a whole-crate re-export, no coupling to the evolving API), so all three workspaces (root, backends/zebra, backends/zaino) resolve its git-sourced transitive librustzcash crates identically and `check-lockstep.sh` stays green.
- Regenerates all three lockfiles via `utils/sync-lockfiles.sh`; CHANGELOG entry included.

Note: the separate "sync existing librustzcash deps" commit was a verified no-op (all pins already at their latest published versions), so this is a single commit. Verified: `cargo test` 194 passed, both backends `cargo check` clean, fmt/clippy clean, lockstep OK.

Draft; part of the Ironwood migration effort. Do not merge yet.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/pool_migration.rs
+++ b/zallet-core/src/components/json_rpc/methods/pool_migration.rs
@@ -23,7 +23,10 @@ use zcash_client_backend::data_api::{Account, WalletRead};
 use zcash_client_sqlite::AccountUuid;
 use zcash_keys::keys::UnifiedSpendingKey;
 use zcash_pool_migration_backend::engine::{
-    MigrationState, MigrationStatus, MigrationTxId, MigrationTxKind, MigrationTxState,
+    MigrationState, MigrationStatus, MigrationTxKind, MigrationTxState,
+};
+use zcash_pool_migration_backend::state::{
+    Blocker, NextAction, TransactionStatus as EngineTransactionStatus,
 };
 use zcash_protocol::consensus::{NetworkUpgrade, Parameters};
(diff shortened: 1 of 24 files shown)


zallet-core/src/components/json_rpc/methods/pool_migration.rs
```diff
@@ -0,0 +1,343 @@
...
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+pub(crate) struct MigrationProgress {
+    /// The number of planned transactions that have been mined so far.
+    completed_transactions: u32,
+    /// The total number of transactions the migration requires.
+    total_transactions: u32,
+}
+
+#[cfg(test)]
+mod tests {
```
dannywillems: Useless. Can be removed.

PR #639, opened 2026-07-17
z_getbalances: split transparent balance into regular and coinbase; z_listunspent: add 'generated' flag

### LINKED ISSUE #635: z_getbalances/getbalances: transparent regular balance includes coinbase; coinbase field never populated; total.spendable ignores transparent
## Summary

`z_getbalances` / `getbalances` reports transparent balance figures that are internally inconsistent with their own documented contract, and every RPC-visible balance number ends up either including un-spendable coinbase funds or omitting the transparent pool entirely. There is no RPC that correctly answers "how much can I currently pay to a transparent recipient?", which makes automated payout flows (e.g. mining pools) fail unpredictably.

## Details

`TransparentBalance` in [`zallet-core/src/components/json_rpc/methods/get_balances.rs`](<../blob/main/zallet-core/src/components/json_rpc/methods/get_balances.rs#L99-L110>) documents:

```rust
struct TransparentBalance {
    /// The transparent balance excluding coinbase outputs.
    regular: Option<Balance>,
    /// The transparent balance in coinbase outputs.
    /// Omitted if no coinbase funds are present.
    coinbase: Option<Balance>,
}
```

But the implementation (same file, [L180-212](<../blob/main/zallet-core/src/components/json_rpc/methods/get_balances.rs#L180-L212>)) does not honor this:

```rust
// TODO: Separate out transparent coinbase and standalone balances.
let transparent_regular = account.unshielded_balance();
...
Ok(Some(TransparentBalance {
    regular: opt_balance_from(regular)?,
    coinbase: None, // never populated
}))
```

`regular` is populated from `AccountBalance::unshielded_balance()`, which is the *entire* transparent balance including coinbase outputs — contradicting the doc comment. `coinbase` is hardcoded to `None` and is never populated regardless of actual coinbase holdings.

Separately, `total.spendable` (from `balance_from` → `AccountBalance::spendable_value()`) only covers the shielded pools, so the transparent balance never contributes to `total` at all.

## Reproduction

Environment: Zallet 0.1.0-beta.1, zebra backend, Zebra v28, mainnet, mining pool wallet with coinbase paid to a wallet-owned t-address.

Every transparent UTXO was independently classified against the node (`getrawtransaction <txid> 1` → `vin[0].coinbase`): two coinbase outputs of 1.25 ZEC each, plus one non-coinbase 1.24035 ZEC output (from a prior `z_sendmany` z→t). The real spendable-to-transparent balance is therefore 1.24035 ZEC.

`z_getbalances` reported:

```json
"transparent": { "regular": { "spendable": { "valueZat": 874035000 } } }
"sapling":     { "spendable": { "valueZat": 9375920000 } }
"total":       { "spendable": { "valueZat": 9375920000 } }
```

* `regular` = 8.74035 ZEC — the two coinbase outputs are folded in; the true non-coinbase balance is 1.24035 ZEC.
* No `coinbase` key is present at all.
* `total.spendable` is byte-identical to `sapling.spendable` — transparent contributes 0.

Reproduced on a second sample: transparent balance 102.49 ZEC, sapling 0.00985 ZEC → `total.spendable` reported as exactly 985000 zats (i.e. only the sapling balance).

## Impact

Since coinbase outputs can only be shielded, never spent transparently, every currently available balance figure is wrong for the "can I pay this transparent recipient" decision:

* `getwalletinfo` is a stub (zcash/zallet#55).
* `z_gettotalbalance.transparent` and `z_getbalances.transparent.regular` both include coinbase.
* `total.spendable` ignores the transparent pool entirely.

Acting on any of these numbers and calling `z_sendmany` fails with `Insufficient balance (have 0, need N including fee)`. This is what took down payouts on our pool: `z_getbalances` reported 102.49 ZEC available, the pool attempted to pay out, and `z_sendmany` immediately failed with `have 0`, halting the payout cycle.

## Current workaround

Enumerate `z_listunspent`, then call `getrawtransaction` on every txid to identify and exclude coinbase outputs — one extra node round-trip per transaction, per payout cycle. `z_listunspent` itself carries no coinbase marker (zcashd's `listunspent` exposed a `generated` field for this).

## Suggested fix

Either of:

1. Finish the `// TODO: Separate out transparent coinbase and standalone balances.` in `get_balances.rs` — likely needs `AccountBalance` (upstream, in `zcash_client_backend`) to carry the coinbase/non-coinbase split so `regular`/`coinbase` can be populated correctly, and `total.spendable` to include the non-coinbase transparent balance.
2. Lower-cost for integrators in the meantime: add a `generated`/coinbase flag to `z_listunspent` output entries, mirroring zcashd's `listunspent`, so callers can filter without an extra `getrawtransaction` round-trip per UTXO.

Reported by a mining pool operator running Zallet in production.

pacu:
Fixes #635.

## Summary

`z_getbalances` now delivers its documented transparent balance contract, and `z_listunspent` identifies coinbase UTXOs — so integrators (mining pools in particular) can finally read "how much can I pay to a transparent recipient" directly from the wallet.

- **`z_getbalances`** (`fe6ce4b`): `transparent.regular` contains only non-coinbase funds; `transparent.coinbase` is populated (immature coinbase reported as `pending`, mature as `spendable`); the account `total` now includes transparent value in its `spendable`/`pending` buckets, so `total == sum of parts` across all five pools. Previously `regular` silently included coinbase, `coinbase` was never emitted, and `total.spendable` ignored the transparent pool entirely — the combination that halted the reporting pool's payouts with `Insufficient balance (have 0, …)`.
- **`z_listunspent`** (`dd87ed2`, standalone/cherry-pickable): transparent entries gain `"generated": true|false` (coinbase origin, mirroring zcashd's `listunspent`; omitted on shielded entries), replacing the per-UTXO `getrawtransaction` round-trip integrators needed. Implemented via two disjoint `CoinbaseFilter` queries; requires no librustzcash changes.
- **Dependency** (`68ad944`): the balance split comes from zcash/librustzcash#2654; `[patch.crates-io]` (workspace root + both backend workspaces) pins that branch (`196f12439a`) until a release containing it ships. `z_gettotalbalance` semantics are intentionally unchanged (pinned by new tests).

## Verification

- `cargo test -p zallet-core`: 208 passed (14 new unit tests: full rendering matrix for the split incl. dust-counted-once and total-equals-sum-of-parts; semantics pins for `z_getbalanceforaccount`/`z_gettotalbalance`; `generated` serialization contract). Both backends `cargo check` clean against the pin.
- End-to-end on a real zebrad+zainod+zallet regtest stack (new `wallet_transparent_coinbase_balance.py`, see the accompanying integration-tests PR): immature coinbase → `coinbase.pending`; mature → `coinbase.spendable` with exact zatoshi sums; post-shield `z_sendmany` z→t output lands in `regular.spendable` only; `generated` flags correct on every entry; `z_gettotalbalance` cross-check passes.

Notes for review: `z_listunspent` transparent output ordering changed (non-coinbase before coinbase per address; no ordering was documented). Immature coinbase remains excluded from `z_listunspent` (upstream maturity clause), while `z_getbalances` now *does* show it as pending — intentional and now-documented behavior.

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/get_balances.rs
+++ b/zallet-core/src/components/json_rpc/methods/get_balances.rs
@@ -177,18 +177,7 @@ pub(crate) fn call(wallet: &DbConnection, minconf: Option<u32>) -> Response {
         .account_balances()
         .iter()
         .map(|(account_uuid, account)| {
-            // TODO: Separate out transparent coinbase and standalone balances.
-            let transparent_regular = account.unshielded_balance();
-
-            Ok(AccountBalance {
-                account_uuid: account_uuid.expose_uuid().to_string(),
-                transparent: opt_transparent_balance(transparent_regular)?,
-                transparent_watchonly: vec![],
-                sapling: opt_balance_from(account.sapling_balance())?,
-                orchard: opt_balance_from(account.orchard_balance())?,
-                ironwood: opt_balance_from(account.ironwood_balance())?,
-                total: balance_from(account)?,
-            })
+            account_balance_response(account_uuid.expose_uuid().to_string(), account)
         })
         .collect::<RpcResult<_>>()?;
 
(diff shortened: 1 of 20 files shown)


zallet-core/src/components/json_rpc/methods/get_balances.rs
```diff
@@ -201,26 +190,69 @@ pub(crate) fn call(wallet: &DbConnection, minconf: Option<u32>) -> Response {
...
+        )?,
+        // `AccountBalance::change_pending_confirmation` and
+        // `AccountBalance::value_pending_spendability` cover the shielded pools only.
+        // Immature transparent coinbase funds are reported by the coinbase bucket's
+        // pending fields.
+        (b.change_pending_confirmation()
+            + b.value_pending_spendability()
+            + regular.change_pending_confirmation()
+            + regular.value_pending_spendability()
+            + coinbase.change_pending_confirmation()
```
nuttycom, line 248: This is a little bit of an oxymoron; it should always be zero but it's fine.

zallet-core/src/components/json_rpc/methods/list_unspent.rs
```diff
@@ -440,9 +454,107 @@ pub(crate) fn call(
...
+fn transparent_unspent_output(
+    txid: String,
+    outindex: u32,
+    confirmations: u32,
+    is_watch_only: bool,
+    address: Option<String>,
+    account_uuid: String,
+    wallet_internal: bool,
+    value: Zatoshis,
+    generated: bool,
```
nuttycom, line 481: Why is this called "generated"?

zallet-core/src/components/json_rpc/methods/list_unspent.rs
```diff
@@ -77,6 +77,12 @@ pub(crate) struct UnspentOutput {
     #[serde(rename = "walletInternal")]
     wallet_internal: bool,
 
+    /// `true` if the output was produced by a coinbase transaction.
+    ///
+    /// Omitted if this is a shielded output.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    generated: Option<bool>,
```
nuttycom, line 84: Weird name for this field.

nuttycom: I guess this comes from zcashd's `listunspent`.

nuttycom (approved):
utACK 27dd5ee9297c5379e6b7780e0c9a326f4c68f0c3

PR #640, opened 2026-07-17
sync: clamp suggested scan ranges to the current chain tip in initialize

### LINKED ISSUE #636: sync::initialize busy-loops with no backoff when suggested scan range exceeds current chain tip
## Summary

The `initialize` step of the sync task in [`zallet-core/src/components/sync.rs`](../blob/main/zallet-core/src/components/sync.rs) can enter a tight busy-loop with no backoff, burning a full CPU core and writing dozens of log lines per second indefinitely.

## Details

The `initialize` loop ([L231-341](../blob/main/zallet-core/src/components/sync.rs#L231-L341)) is structured as:

```rust
let (current_tip, starting_boundary) = loop {
    let chain_view = chain.snapshot().await?;
    let current_tip = chain_view.tip().await?;
    info!("Latest block height is {}", current_tip.height());
    db_data.update_chain_tip(current_tip.height())?;
    let starting_boundary = update_boundary(BlockHeight::from_u32(0), current_tip.height());
    let scan_range = match db_data.suggest_scan_ranges()? ... ;
    match steps::scan_blocks(chain_view, db_data, params, &scan_range, &decryptor, shutdown_height).await {
        Ok(flow) if flow.is_break() => break (current_tip, starting_boundary),
        Ok(_) => {}                                       // no sleep, immediate re-loop
        Err(error) if is_retryable(&error) => { time::sleep(REORG_RETRY_BACKOFF).await; }
        Err(error) => return Err(error),
    }
};
```

The retryable-error arm backs off via `REORG_RETRY_BACKOFF` (200ms), but the "made no progress" arm (`Ok(_) => {}`) does not. If `scan_range` ends up above the current chain tip, `scan_blocks` has nothing to scan, returns `Ok` without breaking, and the loop immediately restarts with no delay and no state change that could resolve the condition — spinning until the chain view happens to advance past the range on its own.

## Reproduction

Environment: Zallet 0.1.0-beta.1, mainnet.

The following two log lines repeated continuously, ~40 times/second:

```
19:05:39.948  INFO initialize: zallet_core::components::sync: Latest block height is 3413617
19:05:39.950  INFO initialize: zallet_core::components::sync::steps: Scanning blocks Historic(3413618..3414133)
19:05:39.951  INFO initialize: zallet_core::components::sync: Latest block height is 3413617
19:05:39.952  INFO initialize: zallet_core::components::sync::steps: Scanning blocks Historic(3413618..3414133)
```

Note the suggested scan range (`3413618..3414133`) starts above the tip height the same iteration just logged (`3413617`) — there is nothing in that range to scan.

Likely cause (inference, not confirmed): the wallet DB had already recorded a higher chain tip (~3414133) from an earlier run, while the current chain view (Zebra's read-only secondary, serving finalized on-disk state) only serves up to 3413617 — i.e. the chain view can lag or regress relative to what the DB has already observed. Regardless of root cause, any scan range suggestion starting above `current_tip` reproduces the same busy-loop.

## Impact

Burns a full CPU core and writes ~40 log lines/second to journald for as long as the condition holds. Not data-threatening, but on a production host it is indistinguishable from a hang, and the log volume is significant over time.

## Suggested fix

Any of:
1. When `scan_range.start > current_tip`, skip scanning and await a chain-tip change instead of immediately retrying.
2. Apply a backoff on the `Ok(_)` arm when zero blocks were actually scanned (mirroring the existing retryable-error backoff).
3. Clamp suggested scan ranges to `current_tip` before selecting one, so an above-tip range is never handed to `scan_blocks` in the first place.

Reported by a mining pool operator running Zallet in production.

pacu:
Fixes #636.

## Summary

`initialize`'s catch-up loop busy-looped at full speed — pinning a core and writing ~40 log lines/second indefinitely — whenever `suggest_scan_ranges` returned a range starting above the current chain view's tip. `scan_blocks` has nothing to scan for such a range and returns `Ok` without breaking, and the `Ok(_)` arm re-ran the loop instantly with no state change, so nothing inside the loop could ever resolve the condition (as reported: `Scanning blocks Historic(3413618..3414133)` against `Latest block height is 3413617`, repeating indefinitely).

The wallet database can legitimately be ahead of the chain view: it records the highest tip it has ever observed, while Zebra's read-only secondary serves finalized on-disk state that can lag (or effectively regress across a restart).

## Fix

Clamp every suggested range to the current tip when selecting the next range to scan (`select_initial_scan_range`, extracted for testability):

- a range straddling the tip scans its available prefix — real progress each iteration;
- a range entirely above the tip is dropped, and when nothing scannable remains the existing no-range arm exits `initialize` to `steady_state`, which waits for chain-tip changes properly instead of polling.

This also covers the silent variant of the same spin where `tree_state_as_of(start - 1)` returns `None` (range starting more than one block above the tip), which previously re-looped without even logging.

## Verification

- 7 new unit tests on `select_initial_scan_range`, including the reported repro (range `3413618..3414133` against tip `3413617` → dropped), straddle clamping, Verify-vs-Historic boundary handling, and above-tip ranges not masking later scannable candidates. `cargo test -p zallet-core`: 201 passed.
- End-to-end regression: `wallet_z_shieldcoinbase.py` (fresh-wallet `initialize` + 100-block catch-up scan) passes against a release build of this branch on the zebrad+zainod regtest stack.

### DIFF
--- a/zallet-core/src/components/sync.rs
+++ b/zallet-core/src/components/sync.rs
@@ -52,7 +52,8 @@ use zcash_client_backend::data_api::{
 };
 use zcash_client_backend::{
     data_api::{
-        TransactionDataRequest, TransactionStatus, WalletRead, WalletWrite, scanning::ScanPriority,
+        TransactionDataRequest, TransactionStatus, WalletRead, WalletWrite,
+        scanning::{ScanPriority, ScanRange},
         wallet::decrypt_and_store_transaction,
     },
     scanning::ScanningKeys,
@@ -203,6 +204,44 @@ fn update_boundary(current_boundary: BlockHeight, tip_height: BlockHeight) -> Bl
     current_boundary.max(tip_height - 100)
 }
 
+/// Selects the next scan range for [`initialize`]'s catch-up loop from the wallet's
+/// suggestions.
+///
+/// Every candidate is first clamped to `current_tip`: `suggest_scan_ranges` can return
+/// ranges beyond the current chain view's tip when the wallet database previously
+/// observed a higher tip than the view now serves (e.g. Zebra's read-only secondary
+/// exposing finalized state that lags what the wallet saw before a restart). Such
+/// ranges contain nothing scannable, and before this clamp existed, handing one to
+/// `scan_blocks` returned `Ok` with no progress and no state change, so the
+/// `initialize` loop spun at full speed — burning a core and logging tens of lines per
+/// second — until the chain view happened to advance past the range (#636). Clamping
+/// scans whatever prefix of a range is actually available now; a range entirely above
+/// the tip is dropped, and if nothing scannable remains the caller's no-range arm exits
+/// to `steady_state`, which waits for tip changes properly.
+///
+/// Beyond the clamp, `Verify` ranges are taken as-is, and ranges at `Historic` priority
+/// or above are truncated to start at `starting_boundary` (history below the boundary
+/// belongs to the `recover_history` task). Lower-priority ranges are not scanned here.
+fn select_initial_scan_range(
+    suggested: impl IntoIterator<Item = ScanRange>,
+    current_tip: BlockHeight,
+    starting_boundary: BlockHeight,
+) -> Option<ScanRange> {
+    suggested
+        .into_iter()
+        .filter_map(|r| {
+            let r = r.truncate_end(current_tip + 1)?;
+            if r.priority() == ScanPriority::Verify {
+                Some(r)
+            } else if r.priority() >= ScanPriority::Historic {
+                r.truncate_start(starting_boundary)
+            } else {
+                None
+            }
+        })
+        .next()
+}
+
 /// Prepares the wallet state for syncing.
 ///
 /// Returns the boundary block between [`steady_state`] and [`recover_history`] syncing.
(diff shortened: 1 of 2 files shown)


zallet-core/src/components/sync.rs
```diff
@@ -203,6 +204,44 @@ fn update_boundary(current_boundary: BlockHeight, tip_height: BlockHeight) -> Bl
...
+/// ranges beyond the current chain view's tip when the wallet database previously
+/// observed a higher tip than the view now serves (e.g. Zebra's read-only secondary
+/// exposing finalized state that lags what the wallet saw before a restart). Such
+/// ranges contain nothing scannable, and before this clamp existed, handing one to
+/// `scan_blocks` returned `Ok` with no progress and no state change, so the
+/// `initialize` loop spun at full speed — burning a core and logging tens of lines per
+/// second — until the chain view happened to advance past the range (#636). Clamping
+/// scans whatever prefix of a range is actually available now; a range entirely above
+/// the tip is dropped, and if nothing scannable remains the caller's no-range arm exits
+/// to `steady_state`, which waits for tip changes properly.
```
nullcopy, line 220: (nit) the history of the bug that is being fixed is not helpful, IMO. I'd rather the comment just explained what the function does, with some minimal "evergreen" rationale

nullcopy (approved):
utACK 18a89035b4b7ec242406e8ccb050590553032310

nullcopy (approved):
utACK a0fdf580d188450f5463d54c62e03c9c1eb319bf

PR #643, opened 2026-07-17
Verify security-relevant wallet.db records against derived key material before use

### LINKED ISSUE #642: Verify security-relevant wallet.db records against derived key material before use
Several records read from wallet.db are trusted in security-relevant decisions. A DB-write attacker is largely out of scope, but the following specific reads can be verified fail-closed against material already in memory at the point of use (decrypted seed, derived USK/UFVK) or anchored outside the DB, without any operation loading secrets it doesn't already load:

1. **Keystore ciphertexts**: age recipients are public, so a ciphertext row can be replaced with an encryption of attacker-controlled seed material under the victim's fingerprint; the fingerprint is never recomputed on read. Fix: after decrypting, recompute the fingerprint (or pubkey, for standalone keys) from the decrypted material and require it to match the row key.

2. **Account UFVKs at spend time**: the spend path signs with the seed-derived USK without checking it corresponds to the account record used for input selection. Fix: call the existing (currently unwired) `validate_seed` after seed decryption, before signing.

3. **Cached transparent change/ephemeral addresses**: inputs re-derive from the seed, but change outputs trust the cached address rows, so a substituted row silently redirects change while the payment succeeds. Fix: before broadcast, require every transparent output to either match a requested payment or re-derive from the USK-derived account pubkey at its claimed (scope, index).

4. **The age recipient set**: an appended recipient row causes all subsequently stored secrets to also be wrapped to the attacker's key. Requiring the identity at encryption time would break encrypting-while-locked, so instead anchor the authoritative recipient list in a manifest beside the identity file and require exact set equality when constructing an encryptor.

nullcopy:
wallet.db records are trusted as-written, but several feed directly into key handling and transaction construction. This PR verifies the ones that can be checked against material the wallet already holds at the point of use.

Closes https://github.com/zcash/zallet/issues/642
[COR-1594](https://linear.app/zodl/issue/COR-1594/improve-zallet-unlock-error-handling)

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/z_send_many.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_send_many.rs
@@ -23,8 +23,11 @@ use zcash_client_backend::{
     },
     wallet::OvkPolicy,
 };
-use zcash_client_sqlite::ReceivedNoteId;
-use zcash_keys::{address::Address, keys::UnifiedSpendingKey};
+use zcash_client_sqlite::{AccountUuid, ReceivedNoteId};
+use zcash_keys::{
+    address::Address,
+    keys::{UnifiedFullViewingKey, UnifiedSpendingKey},
+};
 use zcash_proofs::prover::LocalTxProver;
 use zcash_protocol::{
     PoolType, ShieldedPool,
@@ -39,8 +42,8 @@ use crate::{
             asyncop::{ContextInfo, OperationId},
             payments::{
                 AmountParameter, IncompatiblePrivacyPolicy, PrivacyPolicy, SendResult,
-                broadcast_transactions, build_request, enforce_privacy_policy,
-                get_account_for_address, get_legacy_pool_account,
+                build_request, enforce_privacy_policy, get_account_for_address,
+                get_legacy_pool_account, verify_and_broadcast_transactions,
             },
             server::LegacyCode,
         },
@@ -405,6 +408,8 @@ pub(crate) async fn call<C: Chain>(
         run(
             wallet,
             chain,
+            account.id(),
+            usk.to_unified_full_viewing_key(),
             proposal,
             #[cfg(feature = "zcashd-import")]
             SpendingKeys::new(usk, standalone_keys),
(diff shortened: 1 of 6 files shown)


book/src/cli/init-wallet-encryption.md
```diff
@@ -12,6 +12,13 @@ be set with the `keystore.encryption_identity` [config option].
 > encryption identity file to already exist. You can generate a plain or
 > passphrase-encrypted identity with [`zallet generate-encryption-identity`].
 
+Alongside the identity file, Zallet maintains a recipients manifest (the identity file's
+path with `.recipients` appended) holding an authoritative copy of the age recipient set
+stored in the wallet database. Zallet refuses to encrypt key material if the two sets
+differ, so a recipient added to the database alone cannot silently receive copies of
+newly stored secrets. If you change the recipient set by editing the database directly,
+you must update the manifest to match.
```
nuttycom: So is the point of this just to have the recipients manifest have read-only file permissions? How much difference does this actually make? This seems like it's just adding a third source of truth to get out of sync.

zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -379,13 +382,22 @@ pub(crate) async fn call<C: Chain>(
...
     )
     .map_err(|e| LegacyCode::InvalidAddressOrKey.with_message(e.to_string()))?;
 
+    // The account record (including its stored UFVK, which input selection and address
+    // derivation rely on) is not integrity-protected, so require it to match the spending
+    // key before signing with it.
+    let expected_ufvk = usk.to_unified_full_viewing_key().encode(wallet.params());
+    if account.ufvk().map(|ufvk| ufvk.encode(wallet.params())) != Some(expected_ufvk) {
+        return Err(LegacyCode::Wallet.with_message(fl!("err-account-key-mismatch")));
+    }
```
nuttycom: Doesn't the call to `create_proposed_transactions` do this same validation internally? Is there something that requires us to make the check earlier?

nullcopy: It does indeed. Good catch. I've dropped that commit out.

book/src/cli/init-wallet-encryption.md
nullcopy: Agree. It only offers a very niche benefit that does not justify the complexity. I'm removing that commit as well.

nuttycom (approved):
utACK

PR #645, opened 2026-07-17
rpc: Opt a UA z_sendmany source into transparent spending per the privacy policy

### LINKED ISSUE #138: rpc: Consider supporting spends from the legacy pool
There are several RPCs in `zcashd` that support arguments like `ANY_TADDR`, which in the case of `z_sendmany` means "Select non-coinbase UTXOs from any transparent addresses belonging to the wallet". The flag was introduced before UAs, and was intended to make spending funds from the legacy transparent bucket easier (as they are historically treated as a single pool with multiple spend authorities, unlike Sapling or UAs).

Currently in e.g. #103 we reject `ANY_TADDR` because we don't yet support spending from the legacy pool. This issue is to decide whether or not we should support the legacy semantics anywhere.

Note that for UAs we do get these semantics for transparent, because the multiple transparent spend authorities are all derived from the same account. In `zcashd` we migrated the legacy transparent APIs to a "legacy account", but that has slightly different semantics for UAs (specifically for Sapling address derivation, but we should check how we mapped transparent over). We could instead enable the legacy transparent APIs to be used only when connected to a specific UUID-named account.

### LINKED ISSUE #644: A clean wallet cannot spend its own transparent funds in one transaction: change fragments by design, and the only multi-address source is gated behind the zcashd legacy pool
**Version:** `zallet 0.1.0-beta.1` (rebuilt against `zebra-rpc 11.1.0`), zebra backend,
`zcash_client_backend 0.24.0-rc.1`, testnet.

## Summary

A wallet created from scratch — no `migrate-zcashd-wallet` — reaches a state where it holds N ZEC
of spendable transparent funds and **cannot pay N ZEC in one transaction**: the funds sit on
several of its own addresses and no `z_sendmany` source can gather them. The wallet creates this
condition itself, on every payment.

## Mechanics

**Change fragments by design.** Transparent change goes to *the next available* internal address —
`reserve_next_n_internal_addresses(account_id, …)` in `zcash_client_backend/src/data_api/wallet.rs`:
*"Add any transparent change outputs, sending them to the next available internal-scope (change)
transparent addresses of the account"*. `GreedyInputSelector` also stops once the target is covered,
leaving tails behind. After a few rounds the funds are spread over A/B/C/D…

**No source gathers them.** `fromaddress` accepts only `"ANY_TADDR"`, a t-addr, a z-addr, or a UA
(`Address::decode` — an account UUID is rejected, contrary to what one might expect):

| source | policy | result |
|---|---|---|
| bare t-addr | `TransparentSpendPolicy::from_one_address` | that address's UTXOs only |
| UA | `SpendPolicy::default()` → shielded-only | asserted by your own `shielded_source_permits_no_transparent_spending` test — *"even though the unified address it names does carry a transparent receiver"* |
| `"ANY_TADDR"` | `any_account_addr()` — does gather | but only via `get_legacy_pool_account()` |

Across all of Zallet, `TransparentSpendPolicy` is constructed in exactly two places:
`z_send_many.rs:87` (`from_one_address`) and `z_send_many.rs:110` (`any_account_addr`, inside
`legacy_pool_spend_policy`). `TransparentSpendPolicy::from_addresses(…)`, which accepts a *set* of
addresses, is never called.

`get_legacy_pool_account()` requires **both** `features.legacy_pool_seed_fingerprint` **and** an
account at ZIP 32 index `0x7FFFFFFF` — `is_legacy_pool_account` =
`seed_fingerprint == legacy_seed_fp && account_index == ZCASH_LEGACY_ACCOUNT`. A clean wallet's
account is index 0, and *"a regular account of the legacy seed is therefore not the legacy pool"*.

The capability is not missing from the backend: `any_account_addr()` is documented as *"non-coinbase
UTXOs from arbitrary transparent receivers belonging to the account"*, with no legacy requirement.
Only the RPC layer gates it.

## Why it matters

We run a mining pool. Miners are paid **transparent→transparent**, and that is not negotiable: most
payout addresses are exchange deposits, and exchanges block a deposit whose transaction has shielded
inputs. So the path is coinbase → `z_shieldcoinbase` → unshield to our own t-addr → one `z_sendmany`
to all payees. Splitting the payout is not an option either — ZIP 317 charges per output, and a
500-payee round is expensive enough once.

Real state after a single payout: `151.43815934` on the internal change address, `0.1` on the public
receiver, `0` on the address our config names. A 170 ZEC round is unpayable while the wallet reports
the total as spendable.

## Workaround, verified end-to-end

`is_legacy_pool_account` checks only the seed fingerprint and the ZIP 32 index, not provenance — so a
clean wallet can manufacture the legacy pool:

```bash
zallet rpc z_recoveraccounts '[{"name":"LEGACYPOOL","seedfp":"<this wallet's own seedfp>","zip32_account_index":2147483647,"birthday_height":<tip>}]'
# then in zallet.toml:  features.legacy_pool_seed_fingerprint = "<the same seedfp>"
```

Before the option, `"ANY_TADDR"` → `-11 "The legacy pool of funds is disabled"`. After it, it
resolves. With the account's funds split across four addresses — `3.0` / `2.0` / `2.0` / `0.99985`,
the last being a change address the wallet itself created — a single `z_sendmany "ANY_TADDR"` of
**7 ZEC**, more than any one address holds, succeeded:

```
txid a9efd22b75628c340bb7c1cee6dff7b1bd31cc2a5581526d3a28b7ce2c396429
transparent inputs : 4
transparent outputs: 2
```

So the capability works perfectly well for an ordinary HD account. The only thing standing between a
clean wallet and it is the `0x7FFFFFFF` index check.

## Ask

Let a regular account spend its own transparent pool — any of:

- allow `any_account_addr()` for a normal account (an explicit account/UA source opting into
  transparent selection, guarded by `AllowLinkingAccountAddresses`, which already governs the
  linkage);
- expose `TransparentSpendPolicy::from_addresses(…)` through `z_sendmany` (accept a list in
  `fromaddress`);
- or drop the `0x7FFFFFFF` condition so `legacy_pool_seed_fingerprint` alone enables one-pool
  semantics for the named seed.

Otherwise please bless the workaround above and document it — but as it stands, the documented modern
semantics (*"each account is a separate pool of funds … the semantics used throughout Zallet, and
that should be used by everyone going forward"*) are unusable for transparent funds: the wallet
scatters its change and then refuses to gather it unless you pretend to be an imported zcashd wallet.

pacu:
Fixes #644.

## Design evaluation (is the lack of `ANY_TADDR` a gap?)

Research across #138, #384, and #217 settles it as follows:

- **`ANY_TADDR`'s legacy-pool gating is by design.** #384 defines the legacy pool as *one specific account of one specific seed* (the migrated `zcashd` wallet at ZIP 32 index `0x7FFFFFFF`), and nuttycom closed #138 with exactly that resolution. Relaxing the index check would make a regular account of the named seed impersonate the legacy pool, which #384's semantics deliberately exclude. This PR does not touch it.
- **The actual gap is the unified-address source.** str4d in #138: *"Note that for UAs we do get these semantics for transparent, because the multiple transparent spend authorities are all derived from the same account."* And Zallet's `z_sendmany` help text already documents `AllowLinkingAccountAddresses` as *"Allow selecting transparent coins from the full account, rather than just the funds sent to the transparent receiver in the provided Unified Address"* — `zcashd`'s exact semantics. But `spend_policy_for` mapped every UA source to shielded-only, so the documented capability was unreachable outside the legacy pool. A clean wallet consequently scattered its transparent change (by design) and then refused to gather it (the bug).

## What this changes

For a UA `fromaddress`, the caller's privacy policy now opts the source into transparent spending, mirroring `zcashd`:

| requested policy | transparent funds selectable |
|---|---|
| `FullPrivacy` (default), `AllowRevealedAmounts`, `AllowRevealedRecipients` | none (unchanged) |
| `AllowRevealedSenders`, `AllowFullyTransparent` | the provided UA's own transparent receiver |
| `AllowLinkingAccountAddresses`, `NoPrivacy` | any of the account's transparent receivers (`TransparentSpendPolicy::any_account_addr`, the same source `ANY_TADDR` uses) |

The opt-in must be derived from the requested policy rather than left to post-hoc enforcement: the `GreedyInputSelector` gathers transparent UTXOs up front (targeting the full request amount) whenever a `TransparentSpendPolicy` is present, so a policy-blind opt-in would make the default fully-shielded send propose transparent inputs and then fail `enforce_privacy_policy`. Enforcement itself stays post-hoc and unchanged. Bare-taddr sources and `ANY_TADDR` are unchanged; the legacy-pool-disabled error now points regular accounts at the UA route.

For the miner's t→t payout in #644 (multiple linked input addresses + transparent recipients), the call is now:

```
z_sendmany "<account UA>" '[...]' 1 null 'NoPrivacy'
```

(`NoPrivacy` because linking multiple input addresses *and* revealing transparent recipients is the lattice meet of `AllowLinkingAccountAddresses` and `AllowRevealedRecipients`.) No manufactured legacy account required. Note one behavioral caveat inherited from the greedy selector: transparent funds are drawn first, and if they cannot cover the request, shielded notes cover the remainder — callers who need a guaranteed all-transparent transaction (e.g. exchange payouts) should ensure the account's transparent balance covers amount + fee. The longer-term explicit fund-source selection belongs to `z_sendfromaccount` (#217).

## Testing

- `cargo test -p zallet-core --lib`: 197 passed (new proptests cover each policy tier, the shielded-only-UA divergence between the `AllowRevealedSenders` and `AllowLinkingAccountAddresses` tiers, bare-taddr confinement under every policy, and the change-policy derivation).
- `cargo clippy -p zallet-core --all-targets` and `cargo fmt --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ be considered breaking changes.
 
 ### Changed
 
+- `z_sendmany` with a unified-address `fromaddress` can now spend the account's
+  transparent funds when the caller's privacy policy permits it, matching
+  `zcashd`'s documented privacy-policy semantics for a UA source. Under
+  `AllowRevealedSenders` or `AllowFullyTransparent` it spends the non-coinbase
+  UTXOs of the transparent receiver in the provided UA; under
+  `AllowLinkingAccountAddresses` or `NoPrivacy` it spends the non-coinbase UTXOs
+  of any of the account's transparent receivers, linking them on-chain when more
+  than one is needed. Previously a UA `fromaddress` was always shielded-only, so
+  a regular account had no way to gather its transparent change in one
+  transaction; only the legacy-pool `ANY_TADDR` source could spend from multiple
+  transparent addresses.
 - `zallet rpc help` is now answered locally instead of being sent to the
   wallet's JSON-RPC server, so it no longer requires a config file, an
   initialized wallet, or a running `zallet start`. The command argument may
--- a/zallet-core/src/components/json_rpc/methods.rs
+++ b/zallet-core/src/components/json_rpc/methods.rs
@@ -596,10 +596,20 @@ pub(crate) trait WalletRpc {
     ///     on-chain, so such a call requires a privacy policy of
     ///     `AllowLinkingAccountAddresses` (or of `NoPrivacy`, if it also has a transparent
     ///     recipient or transparent change). Use `z_shieldcoinbase` to shield coinbase
-    ///     UTXOs from multiple transparent addresses.
-    ///   If a unified address is provided for this argument, the TXOs to be spent will be
-    ///   selected from those associated with the account corresponding to that unified
-    ///   address, from value pools corresponding to the receivers included in the UA.
+    ///     UTXOs from multiple transparent addresses. For a regular (non-legacy-pool)
+    ///     account, the equivalent is to pass one of the account's unified addresses as
+    ///     `fromaddress` with a privacy policy of `AllowLinkingAccountAddresses` (or
+    ///     `NoPrivacy`, when the transaction also has a transparent recipient or transparent
+    ///     change).
+    ///   If a unified address is provided for this argument, the TXOs to be spent are
+    ///   selected from the account corresponding to that unified address. Its shielded funds
+    ///   are always available. Transparent funds become available as the privacy policy
+    ///   permits: under `AllowRevealedSenders` or `AllowFullyTransparent`, the non-coinbase
+    ///   UTXOs of the transparent receiver in the provided UA; under
+    ///   `AllowLinkingAccountAddresses` or `NoPrivacy`, the non-coinbase UTXOs of any of the
+    ///   account's transparent receivers (which links them on-chain when more than one is
+    ///   spent). A transaction that both spends from multiple transparent addresses and has a
+    ///   transparent recipient or transparent change requires `NoPrivacy`.
     /// - `amounts` (array, required) An array of JSON objects representing the amounts to
     ///   send, with the following fields:
     ///   - `address` (string, required) A taddr, zaddr, or Unified Address.
--- a/zallet-core/src/components/json_rpc/methods/z_send_many.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_send_many.rs
@@ -40,7 +40,7 @@ use crate::{
             payments::{
                 AmountParameter, IncompatiblePrivacyPolicy, PrivacyPolicy, SendResult,
                 broadcast_transactions, build_request, enforce_privacy_policy,
-                get_account_for_address, get_legacy_pool_account,
+                get_account_for_address, get_legacy_pool_account, parse_privacy_policy,
             },
             server::LegacyCode,
         },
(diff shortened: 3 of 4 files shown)


pacu:
Closing after design discussion: this PR replicated zcashd's coupling where the privacy policy doubles as a coin selector (a looser policy silently changes which funds are selected). That is exactly the behavior Zallet's design moves away from — selection and privacy acknowledgement are deliberately decoupled (the privacy policy is enforced post-hoc against the proposal, never used to pick funds), and #217 explicitly aims to leave privacy-policy-driven APIs behind.

The capability gap #644 describes is real, but the right home for it is `z_sendfromaccount` (#217) with an explicit fund-source argument (`orchard | sapling | any_transparent | {specific taddrs}`): the caller names the funds outright, which also guarantees an all-transparent transaction for exchange-bound payouts instead of silently topping up from shielded notes. A follow-up PR implements the minimal one-shot form of that method.

Issue #603, opened 2026-07-15
book: Render the full generated example config as a configuration reference


pacu:
Part of zcash/zallet#600.

**Current behavior:** `book/src/cli/example-config.md` includes only the first 25 lines of the generated example config (`{{#include ../../../backends/zebra/tests/cmd/example_config.out/zallet.toml::25}}`). The full generated file is 360 lines and documents every config option (generated from rustdoc on `zallet-core/src/config.rs`, kept accurate by a trycmd test). There is no page where an operator can read all options without running `zallet example-config`.

**Expected:** A "Configuration reference" page that includes the entire generated file (same `{{#include}}` mechanism, no truncation), separate from the `example-config` CLI page. Optionally split per-section with anchors for linkability.

**Why:** Operators evaluating Zallet read docs before installing anything; the config surface is where most setup questions land. Zero new prose to maintain — the content already exists and is CI-checked.

Related: zcash/zallet#187, zcash/zallet#124.

---

*Drafted with AI assistance (Claude); facts verified against 29bb94a.*

Issue #600, opened 2026-07-15
book: Documentation gap analysis for the 1.0.0 release (tracking)


pacu:
This issue tracked the documentation gaps identified in a code/book/issue-tracker analysis ahead of the Zallet 1.0.0 release. **All scoped work has merged; closing as complete.**

## Summary of the analysis

**Code documentation (rustdoc):** `missing_docs` is enforced across the workspaces and the tree compiles with zero warnings, so nominal coverage is complete. The doc-from-code pipeline is healthy: config rustdoc generates `zallet example-config` output (trycmd-tested), and the RPC trait rustdoc generates `zallet rpc help` and the `rpc.discover` OpenRPC document. Gaps addressed: rustdoc is now published alongside the book, and crate/module-level architecture docs were expanded.

**The Zallet Book:** Added a JSON-RPC method reference, a zcashd→Zallet status matrix, a full configuration reference, a Concepts section, backup & restore, operations, Docker install, an end-to-end tutorial, and a troubleshooting page; fixed the factual bugs.

**Principle followed:** fact-based documentation — generated from code where possible, citing [ZIPs](https://zips.z.cash) for protocol semantics — and concise.

## Scoped issues — all delivered

- [x] #601 — JSON-RPC method reference (generated from the RPC trait rustdoc, CI-pinned) — #627
- [x] #602 — zcashd→Zallet RPC status matrix — #612
- [x] #603 — full configuration reference from the generated example config — #625
- [x] #604 — Concepts section (accounts, encryption, notes/fees, async ops) — #629
- [x] #605 — backup & restore guide — #624
- [x] #606 — operations guide (logging, monitoring, deployment, shutdown) — #626
- [x] #607 — Docker-based installation — #630
- [x] #608 — end-to-end first-transaction tutorial — #631
- [x] #609 — publish rustdoc + expand architecture docs — #632
- [x] #610 — troubleshooting page — #628
- [x] #611 — fix backend name and stale URLs — #613

## Follow-on work that emerged during the initiative

Delivered:
- Feature experimental/deprecated lifecycle config docs (#187) — #646
- Indexer/validator connection config + broken `validator_cookie_auth` fix (#647) — #649
- Legacy transparent pool concept page (#648) — #650

Also closed via the merged migration-guidance PR (#612): #323, #324, #325, #326, #328, #329, #330, #332.

Remaining (tracked separately, not blocking this umbrella):
[...]

PR #646, opened 2026-07-17
config: Document the experimental/deprecated feature lifecycle

### LINKED ISSUE #187: Document experimental and deprecated feature processes on config options
#166 added the `[feature]` config section for #117. However the fields are not yet fully documented such that users will know how to use them and how to expect them to change. We need to finish the documentation.

pacu:
## Summary

Documents the process the `[features]` config section exists to support. The section already had per-field docs, but nothing explaining the *lifecycle* those fields participate in — which #187 asks for ("the fields are not yet fully documented such that users will know how to use them and how to expect them to change").

- Expands the rustdoc on `FeaturesSection` with the feature lifecycle (experimental → stabilised/removed; active → deprecated → removed), what `as_of_version` detects across upgrades, and why retired flags should be left in the config until acted on (that is how Zallet alerts you instead of silently changing behaviour).
- Adds matching orientation docs to `DeprecatedFeaturesSection` and `ExperimentalFeaturesSection`.

Because config rustdoc is the source for `zallet example-config`, this prose flows into the generated output (fixture regenerated with `TRYCMD=overwrite`) and therefore into the book's configuration reference (#603) — documented once, surfaced everywhere.

Verified against `config.rs` (the `as_of_version` compatibility-check behaviour) and the existing `legacy_pool_seed_fingerprint` background docs.

Closes #187. Part of #600.

## Test plan

- [x] `cargo test --manifest-path backends/zebra/Cargo.toml --features zcashd-import --test cli_tests` passes (fixture matches regenerated output)
- [x] `cargo fmt --all -- --check` clean; `cargo clippy -p zallet-core --all-targets` zero warnings; `cargo doc --no-deps -p zallet-core` zero warnings

## AI disclosure

Written with AI assistance (Claude Opus 4.8, via Claude Code); `Co-Authored-By` metadata is on the commit. Reviewed by the PR author.

### DIFF
--- a/backends/zebra/tests/cmd/example_config.out/zallet.toml
+++ b/backends/zebra/tests/cmd/example_config.out/zallet.toml
@@ -141,6 +141,31 @@ network = "main"
 #
 # Settings for Zallet features.
 #
+# Zallet's behaviour evolves over time: new functionality starts out as an
+# experimental feature that must be explicitly enabled, and functionality on its way
+# out becomes a deprecated feature that must be explicitly re-enabled. This section
+# records the non-default choices this wallet has made, along with the Zallet version
+# they were made against (`as_of_version`), which enables Zallet to detect when a
+# feature named in this config has changed state across an upgrade, rather than
+# silently changing the wallet's behaviour.
+#
+# The lifecycle of a feature:
+#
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
doc-only ACK

PR #649, opened 2026-07-17
book: Fix broken validator_cookie_auth reference and clarify indexer config

### LINKED ISSUE #647: book: Document the indexer/validator connection config (and fix the broken `validator_cookie_auth` reference)
Part of #600.

**Current behavior:** The book documents the `zebra` backend's `[indexer.read_state_service]` path thoroughly (Wallet setup), but the validator/JSON-RPC connection settings — which the `zaino` backend relies on entirely, and which the `zebra` backend still needs for mempool, transaction submission, and non-best-chain reads — are only mentioned in passing in the setup guide, and one mention is **wrong**:

- `book/src/guide/setup.md` tells users to set `validator_cookie_auth = true`. **No such config field exists.** `IndexerSection` (`zallet-core/src/config.rs`) has `validator_address`, `validator_cookie_path`, `validator_user`, `validator_password` — and cookie auth is enabled simply by setting `validator_cookie_path` ("If set, cookie file authorization will be used"). Because the section is `#[serde(deny_unknown_fields)]`, a config containing `validator_cookie_auth` fails to parse outright. So the documented setup is not just incomplete, it is broken.

**Expected:**
1. Fix the `validator_cookie_auth` reference in the setup guide (cookie auth = set `validator_cookie_path`; basic auth = set `validator_user`/`validator_password`).
2. Add a short "Connecting to your full node" section (or fold into the config reference) covering the four `[indexer]` validator settings, defaults (localhost + standard JSON-RPC port), and which the `zebra` vs `zaino` backends use — closing the asymmetry with the well-documented `read_state_service` path.

**Why:** This is the first thing every operator configures, the `zaino` backend depends on it completely, and the current instructions produce a parse error.

---
*Drafted with AI assistance (Claude); facts verified against `config.rs` at HEAD.*

pacu:
## Summary

Fixes a broken setup instruction and clarifies the indexer/validator connection config.

`book/src/guide/setup.md` told users to set `validator_cookie_auth = true`, but **no such field exists**. `IndexerSection` (`zallet-core/src/config.rs`) has `validator_address`, `validator_cookie_path`, `validator_user`, `validator_password`, and cookie auth is enabled just by setting `validator_cookie_path` ("If set, cookie file authorization will be used"). Because the section is `#[serde(deny_unknown_fields)]`, a config containing `validator_cookie_auth` fails to parse — so following the guide produced a broken config.

- Corrects the cookie-auth instruction.
- Adds a short note that both the `zebra` and `zaino` backends use these `[indexer]` JSON-RPC settings (mempool, tx submission, non-best-chain reads), and that `zaino` relies on them for all chain data unless `[indexer.read_state_service]` is also set — closing the documentation asymmetry with the well-covered read-state path.

Closes #647. Part of #600.

## Test plan

- [x] Grepped the book for remaining `validator_cookie_auth` references (none)
- [x] `mdbook build` clean; cross-reference anchor resolves

## AI disclosure

Written with AI assistance (Claude Opus 4.8, via Claude Code); `Co-Authored-By` metadata is on the commit. Reviewed by the PR author.

### DIFF
--- a/book/src/guide/setup.md
+++ b/book/src/guide/setup.md
@@ -49,9 +49,18 @@ In particular, you currently need to configure the `[indexer]` section to point
 at your full node's JSON-RPC endpoint. The relevant config options in that
 section are:
 - `validator_address` (if not running on localhost at the default port)
-- `validator_cookie_auth = true` and `validator_cookie_path` (if using cookie
-  auth)
+- `validator_cookie_path` (if using cookie authentication): set it to your
+  full node's cookie file. Setting this path is what enables cookie auth;
+  there is no separate on/off flag.
 - `validator_user` and `validator_password` (if using basic auth)
+
+Both the default `zebra` backend and the `zaino` backend use these `[indexer]`
+settings to reach the full node over JSON-RPC. The `zebra` backend reads chain
+state directly from a co-located `zebrad` (see below) — including non-best-chain
+(side-chain) blocks and transactions, which `zebrad`'s local state tracks — so
+it uses JSON-RPC only for the mempool and transaction submission. The `zaino`
+backend uses JSON-RPC for **all** chain data, unless you also configure
+[`[indexer.read_state_service]`](#reading-chain-state-from-a-local-zebrad).
 
 ### Reading chain state from a local zebrad
 
@@ -101,7 +110,7 @@ zebra_state_path = "/home/<username>/.cache/zebra"
 
 Notes:
 - The JSON-RPC `[indexer]` settings above are still required: they are used for the
-  mempool, transaction submission, and non-best-chain block reads.
+  mempool and transaction submission.
 - `zebrad` must be running on the **same machine** (Zallet reads its state files
   directly), built with the `indexer` feature, and configured with an
   `indexer_listen_addr`.


book/src/guide/setup.md
```diff
@@ -49,10 +49,17 @@ In particular, you currently need to configure the `[indexer]` section to point
...
 section are:
 - `validator_address` (if not running on localhost at the default port)
-- `validator_cookie_auth = true` and `validator_cookie_path` (if using cookie
-  auth)
+- `validator_cookie_path` (if using cookie authentication): set it to your
+  full node's cookie file. Setting this path is what enables cookie auth;
+  there is no separate on/off flag.
 - `validator_user` and `validator_password` (if using basic auth)
 
+Both the default `zebra` backend and the `zaino` backend use these `[indexer]`
```
Cosmos-Harry, line 57: Non-blocking but are we over generalizing the non-best-chain block reads? Should instead be:

Both the default `zebra` backend and the `zaino` backend use these `[indexer]` settings for mempool access and transaction submission. zaino also uses them for all other chain data, including non-best-chain reads, unless `indexer.read_state_service` is set; zebra requires that section unconditionally, so it only ever uses these settings for mempool and submission.

pacu: Good catch — adopted the clearer two-backend split. One deviation from the exact suggested wording: I kept non-best-chain block reads attributed to the `zebra` backend, because `ReadStateServiceSection`'s rustdoc in config.rs states the JSON-RPC settings are used for "mempool access, transaction submission, and any non-best-chain block reads" even when read_state_service is set (Zebra's read-state view is best-chain only, so side-chain/reorg blocks still come over JSON-RPC). The existing read_state_service note on this page says the same. Happy to drop it if that code comment is itself stale — let me know.

Cosmos-Harry: Checked the code again — `reader.rs` has `ReadRequest::AnyChainTransaction`/`AnyChainBlock`, both served locally via `ReadStateService` with no JSON-RPC fallback (`side_chain_transaction`, `raw_block_by_hash`). Zebra's local non-finalized state does track side chains, so non-best-chain reads for the `zebra` backend stay local. The `ReadStateServiceSection` rustdoc looks like it's describing `zaino`'s behavior, not `zebra`'s, worth double-checking that comment too.

pacu: Confirmed against reader.rs — you're right. AnyChainBlock and AnyChainTransaction (incl. AnyTx::Side) are both served locally via ReadStateService with no JSON-RPC fallback, so non-best-chain reads stay local for the zebra backend. Corrected both statements in setup.md (my paragraph + the existing read_state_service note): the zebra backend uses JSON-RPC only for the mempool and transaction submission. The stale ReadStateServiceSection rustdoc (which was my original source, and which as you note reads like it's describing zaino) is tracked as #654, including regenerating the example-config fixture once zaino's path is verified. Diff: https://github.com/zcash/zallet/compare/b83b210a643136095519ec6be835cbdb2a7e2f6d..c3a10c0e3b7b269b4a03d569c130d3c7d6ee1ed8

zcash-harry (approved):

PR #651, opened 2026-07-17
Replace the temporary rage git pin with the released age 0.12

nuttycom:
rage 0.12 has been released, carrying the APIs the keystore relies on (`EncryptedIdentity`, `Send + Sync` bounds on `IdentityFile::into_identities`) that motivated the temporary `[patch.crates-io]` pin to rage `main`. The pinned revision `d28f10ee` is an ancestor of the `v0.12.1` tag, so switching to the crates.io release requires no code changes.

- All three workspaces (root, `backends/zebra`, `backends/zaino`) now depend on `age 0.12` from crates.io; the git pins and their explanatory comments are removed, and the lockfiles resolve `age 0.12.1` / `age-core 0.12.0`. `utils/check-lockstep.sh` passes.
- cargo-vet: the `@git`-suffixed exemptions and `audit-as-crates-io` policies for `age`/`age-core` are replaced by `[[trusted.age]]` / `[[trusted.age-core]]` entries for str4d as publisher. `cargo vet --locked` passes in all three workspaces.
- `cargo check --all-features` passes in all three workspaces. `cargo deny` licenses/bans/sources pass; the two advisory failures (RUSTSEC-2026-0204, RUSTSEC-2026-0173) are pre-existing on `main` and unrelated to this change.

https://claude.ai/code/session_01DbEidoB2RKxwfWLNy2Ci5w

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,8 +103,9 @@ dependencies = [
 
 [[package]]
 name = "age"
-version = "0.11.3"
-source = "git+https://github.com/str4d/rage.git?rev=d28f10ee776ce7391f7065695997f998631b113a#d28f10ee776ce7391f7065695997f998631b113a"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


pacu (approved):
utACK

PR #652, opened 2026-07-18
migrate-zcashd-wallet: import all wallet transactions directly

### LINKED ISSUE #271: Migrate `wallet.dat` transactions directly, instead of restoring them from the chain.
We want to ensure that all transactions, even those that were never mined, are transferred across when migrating a wallet from a `wallet.dat` file. In #152, due to limitations in `zewif-zcashd`, we only brought across transactions that for which we could find the data on-chain.

nullcopy:
The migration used the wallet's saved transactions only to estimate birthday heights, then dropped all of them from the ZeWIF document unless --buffer-wallet-transactions was given, relying on the post-import chain scan to restore transaction history. That scan only re-derives transactions from the main-chain blocks it scans, so any transaction that was never mined (e.g. a still-unmined send) or was recorded only against a non-main-chain block (a conflicted or reorged transaction) had no on-chain record to recover and was lost.

Always carry every transaction through to be imported directly, and remove the now-redundant --buffer-wallet-transactions flag.

Closes https://github.com/zcash/zallet/issues/271
[COR-325](https://linear.app/zodl/issue/COR-325/migrate-walletdat-transactions-directly-instead-of-restoring-them-from)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ be considered breaking changes.
     an optional `ivk` argument that exports the account's unified incoming
     viewing key instead.
 
+### Removed
+
+- The `migrate-zcashd-wallet --buffer-wallet-transactions` flag has been removed.
+  All wallet transactions are now always imported directly, so the flag no longer
+  has any effect.
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


nuttycom (approved):
utACK

PR #658, opened 2026-07-20
Document database write atomicity requirements for agents

dannywillems:
## Motivation

Two atomicity defects were found by audit and fixed recently: `initialize_recipients` inserting age recipients in an untransacted loop behind a one-shot guard (#621), and `z_importkey` writing the keystore and the wallet account in two separate transactions (#622). This adds the check to `AGENTS.md` so the class is caught when the next such code is written rather than after.

## Approach

The section is a procedure rather than an exhortation. A rule saying "make writes atomic" would not have caught either defect; whoever wrote both already knew writes should be atomic in the abstract. What made them unrecoverable was a second-order property:

- in `initialize_recipients`, a partial insert plus a one-shot "if non-empty, refuse" guard, so no retry could ever repair it (the guard also read on a separate connection, making it a check-then-act race);
- in `z_importkey`, a partial write plus an asymmetric read path, so `z_exportkey` saw a key the scanner never did.

In both, the harm came from the retry not repairing, not from the partial write itself. So the section makes that an explicit question rather than an implied one. Four questions to answer in the PR description whenever an operation issues more than one write: how many writes, must they land together, what happens on retry after a partial commit, and does any read path see only one side. One-shot guards and asymmetric read paths are then named as hard triggers for a transaction rather than judgement calls.

It also records:

- which primitive to use, with a verified in-repo exemplar for each (`conn.transaction()` per `store_encrypted_standalone_transparent_keys`, `transactionally_with_extension` per `z_importkey`, `transactionally`);
- that a `rusqlite` transaction cannot be held across an `.await`, since that constraint shaped the `z_importkey` fix, and that a check hoisted out with the async work must be re-checked inside;
- that every fallible call inside a `transactionally` closure must propagate with `?`. The `zcash_client_sqlite` implementations take no savepoint per method, so a discarded error leaves partial writes in the enclosing transaction, which then commits while the operation reports success;
- that compensating logic is not a substitute, with the concrete reason (`delete_account` does not restore the scan-queue and note-commitment-tree state `add_account` rewinds);
- a test-evidence requirement: a unit test that errors inside the transaction and asserts nothing persisted.

## Known gap this does not close

Neither #621 nor #622 shipped a rollback test, and there are none in `zallet-core/src`. The requirement above applies to future changes, so both fixed sites remain uncovered and a later refactor could silently undo either. `transactionally_with_extension_rolls_back_on_error` in `zcash_client_sqlite` is a usable template. Happy to add both tests here or in a follow-up, whichever reviewers prefer.

## Testing

Documentation only; no code change.

## Related

Context, not closed by this PR: #615, #616. A companion PR adds the library-author version of this section to librustzcash's `AGENTS.md` (zcash/librustzcash#2672).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -233,6 +233,80 @@ Note that the Dockerfile's apt pins rot on their own, with no change on our side
 
 If the diff is not benign, or you cannot establish what changed, STOP and say so in the PR rather than merging a version you have not audited.
 
+## Database Write Atomicity
+
+`rusqlite` autocommits every statement executed outside an explicit transaction. A
+sequence of writes that must land together will therefore commit one at a time unless
+it is wrapped, and a failure partway through (an I/O error, a full disk, a crash, or an
+error returned by a later step) leaves a subset committed. This has produced real bugs
+in Zallet, so it is checked on every change that touches the wallet database.
+
+**Before writing or reviewing any code that writes to the wallet database, answer these
+four questions in order.** Answer them in the PR description when the answer to the
+first is "more than one".
+
+1. **How many writes does this operation issue?** Count every `INSERT`, `UPDATE`, and
+   `DELETE`, including those inside loops and inside the functions it calls. If the
+   answer is one, stop here. Otherwise continue.
+2. **Must they land together?** Is there any subset of these rows that, if committed
+   alone, the rest of the wallet would treat as valid state? If so, they must share one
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
ACK

PR #660, opened 2026-07-20
ci: scope integration dispatch by path and cancel superseded runs

nuttycom:
Two CI-efficiency changes, motivated by wasted integration-test compute:

**1. Scope the integration-test dispatch by path.** The dispatch now lives in
its own `integration.yml` workflow, gated by a `paths` include list, so
documentation, other CI/workflow, licensing, and build/supply-chain tooling
changes no longer kick off an integration run. A `paths` filter (rather than
`paths-ignore`) is used because GitHub only supports `!` negation in `paths`;
the trailing re-include means a change to the trigger workflow itself is always
integration-tested (which is why this PR triggers a run). The now-unused
`merge_group` trigger is dropped — the merge queue is not required here, and the
dispatch is a non-blocking, fire-and-forget call.

**2. Cancel superseded runs on rapid re-push.** A `concurrency` group is added
to the CI, lints, and audits workflows so a fresh push to a PR cancels that
PR's in-flight runs; `main` pushes are never cancelled. The dispatch payload now
carries a stable `source_ref` (PR number or branch) so that
zcash/integration-tests can likewise cancel a superseded interop run for the
same PR (see paired PR below).

### Paired change

- zcash/integration-tests#178 consumes `source_ref` as its concurrency key.

This PR pins that branch so the integration run for this PR exercises the paired
change end to end:

ZIT-Revision: ci/cancel-superseded-interop-runs

### DIFF
--- a/.github/workflows/audits.yml
+++ b/.github/workflows/audits.yml
@@ -9,6 +9,12 @@ on:
 permissions:
   contents: read
 
+# Cancel superseded runs: a fresh push to a PR aborts the PR's in-flight run.
+# `main` pushes are never cancelled, so every landed commit is fully tested.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 jobs:
   # Each chain backend lives in its own cargo workspace with its own lockfile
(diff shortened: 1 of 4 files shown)


dannywillems (approved):
utACK

PR #661, opened 2026-07-20
AGENTS: add a no-magic-numbers code convention

dannywillems:
Adds a "Code Conventions" section to `AGENTS.md` documenting a no-magic-numbers
rule: give a bare numeric or string literal a doc-commented `const` with a
rationale, and reuse the protocol's named constants (`COIN`, the ZIP-317
`MARGINAL_FEE`, and the migration engine's exported constants) rather than
re-deriving their values, in both production code and tests.

This convention was written alongside the Ironwood migration work; extracting it
into its own commit/PR so it can land independently of that feature.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,6 +120,21 @@ Key external dependencies from the Zcash ecosystem:
 - `zebra-chain`, `zebra-state`, `zebra-rpc` -- chain data types and node RPC
 - `zaino-*` -- indexer integration
 
+## Code Conventions
+
+- **Never use magic numbers.** Do not inline a bare numeric (or string) literal
+  whose meaning is not obvious from context. Give it a `const` with a
+  doc-commented rationale, and reuse the named constants the `librustzcash`
+  crates already export rather than re-deriving their values:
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -120,6 +120,14 @@ Key external dependencies from the Zcash ecosystem:
 - `zebra-chain`, `zebra-state`, `zebra-rpc` -- chain data types and node RPC
 - `zaino-*` -- indexer integration
 
+## Code Conventions
+
+- **Never use magic numbers.** Do not inline a bare numeric (or string) literal
```
pacu, line 125: May a reference to the librustzcash crates help here?

pacu (approved):

AGENTS.md
dannywillems: Addressed.

dannywillems: See https://github.com/zcash/zallet/compare/d087cb3eaa49384403def15d97dd28362a11c89b..0b8f59ea30c69eb5b84c219088b1004134311ac6

daira (commented):
Post-hoc ACK

PR #669, opened 2026-07-21
rpc: Mark panicked async operations as failed

### LINKED ISSUE #668: Panicked tasks are left in the 'Executing' state forever
A panic in an async operation task previously left the operation stuck in the Executing state forever, with no result ever recorded and no way to prune it (GHSA-582c-4hqf-53q3). Should catch the panic inside the task with catch_unwind and funnel it through the existing result-recording path, surfacing the panic message as a Misc RPC error.

nullcopy:
A panic in an async operation task previously left the operation stuck in the Executing state forever, with no result ever recorded and no way to prune it (GHSA-582c-4hqf-53q3). Catch the panic inside the task with catch_unwind and funnel it through the existing result-recording path, surfacing the panic message as a Misc RPC error.

Thanks @ouicate for the report.

Supersedes https://github.com/zcash/zallet/pull/466
Closes https://github.com/zcash/zallet/issues/668
[COR-1631](https://linear.app/zodl/issue/COR-1631)

### DIFF
--- a/zallet-core/src/components/json_rpc/asyncop.rs
+++ b/zallet-core/src/components/json_rpc/asyncop.rs
@@ -1,8 +1,10 @@
 use std::future::Future;
+use std::panic::AssertUnwindSafe;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use documented::Documented;
+use futures::FutureExt;
 use jsonrpsee::{
     core::{JsonValue, RpcResult},
     types::ErrorObjectOwned,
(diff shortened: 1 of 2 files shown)


zallet-core/src/components/json_rpc/asyncop.rs
```diff
@@ -135,18 +137,30 @@ impl AsyncOperation {
...
-                    )
-                    .expect("round trip should succeed")
+                // Run the async task, mapping the concrete result into a generic JSON
+                // blob. A panic is recorded as a failure of the operation, so it does
+                // not remain in the `Executing` state forever.
+                let res = AssertUnwindSafe(async {
+                    f.await.map(|ret| {
+                        serde_json::from_str(
+                            &serde_json::to_string(&ret)
+                                .expect("async return values should be serializable to JSON"),
```
nuttycom: Is this guaranteed structurally anywhere? Also, it seems really weird that there isn't a conversion that skips the string encoding here.

nullcopy: Good points.
1. No, it's not enforced, although everywhere seems to follow JSON serializability as a convention.
2. Yeah, I completely missed this in my self review 🤦.

Fixed both, but note: I couldn't think of a clean way to enforce the serializability invariant, so instead I removed the assertion and instead route a serialization failure through the normal result path.

nuttycom (approved):
utACK

PR #672, opened 2026-07-22
Lock spend proposal inputs while operations are in flight

dannywillems:
Wires the librustzcash note-locking facility (zcash/librustzcash#2726, itself building on zcash/librustzcash#2716) into Zallet's spend paths, so that concurrent operations on one account cannot select the same inputs during the selection-to-broadcast window (the conflict described in the Zallet audit).

Based directly on `main`; independent of the pool-migration integration branches.

## What this adds

- **Input locking on every spend path.** `z_sendmany` and `z_shieldcoinbase` generate one `LockOwner` per operation and pass `LockRequest::new(owner, note_lock_blocks)` to `propose_transfer` / `propose_shielding_coinbase`, locking the proposal's inputs until `target_height + note_lock_blocks`. Locks are keyed by the owner token: a concurrent operation cannot release or steal them, while the owning operation's retry is an idempotent re-acquire.
- **Config**: `builder.note_lock_blocks` (default **40 blocks**, roughly 50 minutes at 75-second blocks; comfortably above worst-case build+prove time and equal to the default `tx_expiry_delta`, so locks outlive an unmined transaction's validity window; `0` disables locking).
- **Retryable busy error.** A lock conflict (`ProposalError::InputsLocked`) maps to a clear "account is busy, retry shortly" RPC error (LegacyCode -4). In practice a second overlapping operation usually surfaces as insufficient funds instead, because selection excludes locked inputs; the busy error fires only on a genuine check-then-lock race.
- **Unlock on failure.** Every failure between successful proposal creation and transaction storage (privacy-policy rejection, orchard-action limit, seed decryption, build/prove/sign) releases the operation's locks via `unlock_proposal_inputs` under its owner, without masking the original error. If the process dies instead, locks self-expire as the chain advances past the lock window.
- **Recovery**: new `z_clearlockedoutputs(account)` RPC wrapping `clear_locked_outputs`, documented as safe only when no in-flight operation for the account remains.
- **Balance visibility**: `z_getbalanceforaccount` reports per-pool `lockedZat`, and `z_getbalances` populates its `locked` bucket. Locked funds leave the spendable totals but remain part of the account total.

## Dependency pin (temporary)

Note locking has merged to librustzcash `main` (zcash/librustzcash#2716, including zcash/librustzcash#2726), and the librustzcash crates are now pinned by git rev to `main` (`beec9b59`) until the next crates.io release; the pin also mirrors main's git `orchard` entry in `[patch.crates-io]` (its `zcash_primitives` uses the ZIP 374 deferred-anchor builder API). Lockfiles regenerated via `utils/sync-lockfiles.sh`; the lockstep check passes. The pin should be replaced with released crate versions before merging.

## Testing

- `cargo fmt --check`, clippy with zero warnings, and full `cargo test --release` across all three workspaces (root, zebra backend, zaino backend); book RPC reference and example config regenerated.
- End-to-end coverage lives in zcash/integration-tests#180: ten note-locking scenarios across five `qa/rpc-tests` suites (locked-balance visibility, concurrent-send double-spend prevention, unlock-on-broadcast, expiry restoring funds, locks surviving a SIGKILL restart, `z_clearlockedoutputs` recovery, account isolation, all-notes-locked errors, shielding with locking, and the short-lock-window self-heal trade-off). All ten pass live against this branch: `ZALLET_BACKEND=zebra uv run ./qa/rpc-tests/wallet_note_locking.py`.

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/z_get_balance_for_account.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_get_balance_for_account.rs
@@ -60,9 +60,18 @@ struct Pools {
 
 #[derive(Clone, Debug, Serialize, JsonSchema)]
 struct PoolBalance {
-    /// The balance in zatoshis.
+    /// The spendable balance in zatoshis.
     #[serde(rename = "valueZat")]
     value_zat: u64,
+
+    /// The value (in zatoshis) currently locked by in-flight spend operations.
+    ///
+    /// Locked value is excluded from `valueZat` (it cannot be selected by a new spend
+    /// while the operation that locked it is in flight) but still belongs to the account;
+    /// it returns to `valueZat` when the operation completes, fails, or its lock expires.
+    /// Omitted if zero.
+    #[serde(rename = "lockedZat", skip_serializing_if = "Option::is_none")]
+    locked_zat: Option<u64>,
 }
 
 pub(super) const PARAM_ACCOUNT_DESC: &str =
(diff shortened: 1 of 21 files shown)


zallet-core/src/components/json_rpc/methods/z_get_balance_for_account.rs
```diff
@@ -284,6 +307,42 @@ mod tests {
...
+    // `valueZat`, and a pool whose ENTIRE balance is locked still appears (spendable 0),
+    // so an in-flight operation does not make the funds appear to vanish.
+    #[test]
+    fn locked_value_is_rendered_and_keeps_the_pool_visible() {
+        let zat = Zatoshis::const_from_u64;
+        let mut balance = account_balance(0, 2 * COIN, 0, 0);
+        balance
+            .with_sapling_balance_mut::<_, BalanceError>(|b| b.add_locked_value(zat(3 * COIN)))
+            .unwrap();
+        balance
```
dannywillems, line 320: Probably missing ironwood.

zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -313,88 +332,120 @@ pub(crate) async fn call<C: Chain>(
...
+        _ => LegacyCode::Wallet.with_message(format!("Failed to propose transaction: {e}")),
+    })?;
+
+    // From here until `create_proposed_transactions` stores the built transactions, this
+    // function owns the proposal's input locks: every error on this path must release
+    // them, or the account's inputs would stay unspendable until the lock window expires.
+    let post_proposal = async {
+        enforce_privacy_policy(&proposal, privacy_policy)?;
+
+        let orchard_actions_limit = APP.config().builder.limits.orchard_actions().into();
```
dannywillems, line 354: Missing ironwood.

pacu:
Scheduling context.

Zallet has moved to a **weekly release train** — a cut every Tuesday, shipping whatever is
merged and green (#781). This PR is targeted at the **beta.8 cut, Tue 2026-09-29**, and it is
deliberately given its own review slot rather than being batched with others: at 21 files it is
the largest diff in the queue.

It also matters more than its size suggests. This is **COR-1664 / Least Authority Suggestion 2,
the last open item from the audit** — Issues A, B, I and N and Suggestions 1, 3, 4, 5, 6, 7, 8,
9 and 10 all shipped in beta.3 or earlier. When this lands, the audit is fully closed out.

Two asks whenever you next pick it up: it currently **conflicts with `main`** and would need a
rebase, and it is still marked draft — worth flipping to ready when you think it is reviewable.

No rush before 09-29; flagging the target now so it does not sit unscheduled.

pacu:
This conflicts with `main`, and is still marked draft — could you rebase and mark it ready when you think it's reviewable?

Issue #195, opened 2025-08-14
Support robust online backup of Zallet's state


daira:
Simply taking a copy of Zallet's database while it might be the process of being written to concurrently is incorrect, as [confirmed by sqlite docs](https://www.sqlite.org/backup.html). There are two correct ways to support robust database backup for a long-running Zallet process:

1. From a different process, obtain a shared (write-exclusive) lock on the database file, copy it, then relinquish the lock.
2. Have the long-running process make a backup of the database using the [Online Backup API](https://www.sqlite.org/c3ref/backup_finish.html#sqlite3backupinit).

Option 1 causes the Zallet process to block on database writes while the file is being copied. There is also the risk that the shared lock might not be released, blocking database writes by the Zallet process indefinitely. On the other hand, this approach works to back up the database while the Zallet process is unresponsive on its RPC API, as long as it is not holding a write-exclusive lock.

Option 2 has the lowest impact on latency. Zallet could have config options to do this periodically with backup rotation. (This could eat up a lot of disk space; we might want to have it monitor free disk space and complain if that gets low.)

The options are not mutually exclusive: for example you could have a command to trigger a manual backup, in addition to using the Online Backup API to take periodic backups. Also, it might be a good idea for the implementation of the `stop` RPC method to automatically take a backup before stopping (see also #184).

Issue #55, opened 2025-04-03
rpc: Implement `getwalletinfo`


str4d:

str4d:
Initial implementation was in #12 and #45. Most of the remaining work is blocked on #36 which is porting the necessary APIs to derive the various transparent balance fields in a way that matches the `zcashd` wallet.

nuttycom:
Closing as complete; we do not intend to add transparent balance data to getwalletinfo. Users should use a dedicated balance retrieval method instead.

PR #675, opened 2026-07-23
Add a global sync lock

### LINKED ISSUE #180: Add a global lock that prevents usage of the wallet when Zallet detects it is not in sync with the main chain
The zcashd wallet keeps itself as close to the chain tip as possible. Originally it was synchronous, but we disconnected them by around 1 second due to a remote key extraction attack. Currently Zallet does the same, and can reasonably maintain this because it bundles Zaino internally (and thus both are offline at the same time).

If we enable Zallet to connect to an external Zaino, then we will need to deal with the situation where Zallet might be offline for some time (e.g. for maintenance) but Zaino keeps running. When Zallet restarts, it may need to fetch state that Zaino no longer has (due to orphan chains being pruned once they entered the finalized part of Zebra's chain). The closest analogues zcashd has are:
- The initial re-validation of the last N blocks that zcashd is aware of.
- "Initial Block Download" mode: when first starting, most of the JSON-RPCs will not respond until zcashd believes it is within 24 hours of the current chain tip.
- "Safe Mode": if an alert is broadcast, affected zcashd nodes will go into a mode where any JSON-RPC that in any way relies on wallet balance (viewing balance, creating txs, etc) will error. (Zebra does not listen to alerts.)

Zallet could have a similar thing where if it detects that it is too out-of-sync with the Zaino index (and thus the main chain), then it can:
- Take the global lock to make the rest of Zallet aware of this state.
- Truncate its state back by 100 blocks (the max rollback state we should ever see, which zcashd probabilistically ensures and Zebra enforces).
- Let the sync engine scan until it has caught back up to Zaino.
- Release the global lock.

The rest of Zallet could then behave in an IBD- or Safe Mode-like way while the global lock is taken.

sellout:
Add a global sync lock that blocks wallet usage while Zallet’s view of the
chain is not trustworthy, mirroring `zcashd`’s two guards — initial block
download and safe mode. Closes #180.

_Supersedes #674 (closed; its branch was force-pushed to apply a `rustfmt`
fix, so it could not be reopened). Same branch, same content at the
fmt-clean head._

## What it does

The sync engine publishes a `SyncStatus` over a `watch` channel — it is the
only component that can observe a recovery in progress — and the JSON-RPC layer
reads it as a cheap, non-blocking borrow (no network round-trip):

- **`CatchingUp`** — the wallet has not yet scanned up to the chain tip, so its
  balance is incomplete (analogous to initial block download). Mapped to
  `ClientInInitialDownload (-10)`.
- **`Recovering`** — the engine is rolling back and rescanning after detecting
  that the wallet’s view diverged from the chain (analogous to safe mode).
  Mapped to `ForbiddenBySafeMode (-2)`.
- **`Synced`** — the wallet’s view is trustworthy.

The engine drives the transitions as it scans, reorgs, and recovers, so the
lock releases naturally once the wallet reaches the tip. Every gated method
takes the lock at the JSON-RPC dispatch layer, before any wallet work is done:

- the balance reads — `z_getbalances`, `z_getbalanceforaccount`,
  `z_gettotalbalance`, `z_listunspent`, `z_getnotescount`;
- the spends — `z_sendmany`, `z_shieldcoinbase`;
- the two PCZT roles that depend on the balance view — `pczt_create`, which
  selects inputs and computes change, and `pczt_extract`, which (for a PCZT
  this wallet created) records the extracted transaction and marks its inputs
  pending-spent.

Gating before the spends build their proposal matters for the error the caller
sees: a lagging wallet would otherwise fail during input selection and report
“insufficient funds”, because the proposal is built against an incomplete view
of the wallet’s notes. The check is also synchronous — it runs before the
execution future is queued — so a locked wallet fails the request immediately
rather than as a failed background operation.

`pczt_combine`, `pczt_prove`, `pczt_inspect`, and `pczt_sign` are deliberately
left ungated: none of them consults the balance view. `pczt_sign` touches the
database only for key material, and a PCZT assembled while the wallet was
trustworthy remains as valid to sign as it ever was.

`getwalletstatus` reports a `locked` field from the same status, and the new
`sync.lock_threshold` config option sets how far the wallet may lag the tip
before it is considered to be catching up (default 100).

## Relationship to the design questions in #180

**IBD vs. Safe Mode (what the lock is over).** Both are modeled explicitly:
`CatchingUp` (the wallet hasn’t scanned up to the tip — initial block download)
and `Recovering` (a detected chain divergence is being rolled back and rescanned
— safe mode), returning `ClientInInitialDownload (-10)` and
`ForbiddenBySafeMode (-2)` respectively. In `zcashd`, IBD blocks most RPCs until
near the tip, while Safe Mode disables a per-command set of wallet-balance-relying
RPCs. This PR gates the same set — the balance reads, the spends, and the
balance-dependent PCZT roles listed above — for *both* states, since in either
case the wallet’s balance view is not trustworthy. (`getwalletinfo` also reports
balances but is currently stubbed to zero — #55 — so it will join the gated set
when it is implemented.)

**“Require full sync” mode.** The lock implements the highest-reliability option
from the thread: while the wallet is not fully scanned to within
`sync.lock_threshold` blocks of the tip (using `block_fully_scanned`, the
in-order fully-scanned height), balance and spend RPCs error rather than operate
on an incomplete view. This deliberately sidesteps the spend-before-sync
balance-invariant problem for the gated methods.

**The on-restart reorg recovery** the issue describes (truncate back and resync)
is now handled by the sync engine’s own `locate_fork_point` → `truncate_to_height`
path, which rolls back to the actual fork point rather than a fixed depth; this
PR surfaces that as the `Recovering` state rather than reimplementing it.

**Deferred, with rationale:**

- *Selectable spendability modes* (pure spend-before-sync, or SBS + a spendability
  oracle, vs. require-full-sync). The oracle doesn’t exist and SBS is the
  underlying default; a mode selector is a larger feature, and this lock is
  forward-compatible as one such mode.
- *Configurable / multiple locks.* This ships a single lock over a fixed,
  security-relevant set with one knob (`sync.lock_threshold`). Per-method or
  multi-lock granularity adds config surface with no concrete use case yet.
- *Online backup on exiting safe mode* (@daira, #195). Blocked on #195; the
  natural hook is the `end_recovery` transition, where a `TODO(#195)` marker now
  records the intent.
- *The broader “when is Zallet ready for use” discussion* is out of scope here;
  this lock is a concrete step consistent with any eventual answer.

## Testing

- New unit tests in `sync::status`:
  - `ensure_available` maps each state to its error code (`CatchingUp` → `-10`,
    `Synced` → `Ok`, `Recovering` → `-2`).
  - writer methods flow through the `watch` channel to the reader, including a
    recovery overriding and then clearing the synced view.
  - the channel starts locked (`CatchingUp`), not `Synced`.
  - the `lock_threshold` tolerance boundary is exact (trailing by exactly the
    threshold is still `Synced`; one more is `CatchingUp`).
  - plus the existing derivation tests for `SyncStatus`.
- New `SyncSection` config tests: `lock_threshold` default (100), override, and
  that `0` is accepted (unlike `recover_batch_size`, which rejects it).
- `zallet-core`: `clippy --all-targets -- -D warnings` clean, and again with
  `--features zcashd-import`; `cargo fmt --all --check` clean; the test suite
  passes (293 tests), including `book_rpc_reference`.
- The `cli_tests` golden configs (incl. the regenerated `example_config`) pass.
- Not unit-tested (would need a full RPC harness with a live `Chain`/database):
  the RPC-dispatch gate wiring — a one-line `ensure_synced()?` whose logic is
  covered by the `ensure_available` test — and the engine transition points in
  `sync.rs`.
- Manual end-to-end against a live external Zaino (offline reorg → `Recovering`
  / `CatchingUp` → balance/spend rejected → `Synced`) still to be exercised.

## Review history

Rebased onto `main` after #541 (the PCZT RPC methods) landed. Two follow-up
commits sit on top of the original one:

- `rpc: Check the sync lock before building a spend proposal` — addresses
  @schell’s review comment; the check moved to the dispatch layer, which also
  drops the `SyncStatusReader` parameter that had been threaded into
  `z_sendmany` and `z_shieldcoinbase`.
- `rpc: Apply the sync lock to the PCZT spend methods` — closes the gap the
  rebase opened, since the PCZT methods were merged while this branch was in
  review and were never added to the gated set.

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/z_shieldcoinbase.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_shieldcoinbase.rs
@@ -44,6 +44,7 @@ use crate::{
             utils::{JsonZec, value_from_zatoshis},
         },
         keystore::KeyStore,
+        sync::SyncStatusReader,
     },
     prelude::*,
 };
@@ -149,6 +150,7 @@ pub(crate) async fn call<C: Chain>(
     mut wallet: DbHandle,
     keystore: KeyStore,
     chain: C,
+    sync_status: SyncStatusReader,
     fromaddress: String,
     toaddress: String,
     fee: Option<JsonValue>,
@@ -284,6 +286,13 @@ pub(crate) async fn call<C: Chain>(
         shielding_utxos,
         shielding_value: value_from_zatoshis(shielding_value_zats),
     };
+
+    // The proposal has been built. Before fetching key material and handing off the
+    // execution future, refuse the spend if the wallet is not synced enough to trust its
+    // view of the chain. This check is synchronous — it runs during the `call().await` in
+    // the dispatch layer, before the async operation is queued — so the caller sees the
+    // failure immediately rather than as a failed background operation.
+    sync_status.ensure_available()?;
 
     // Derive the spending key for the source account.
     let derivation = account.source().key_derivation().ok_or_else(|| {
--- a/zallet-core/src/components/json_rpc/methods/z_shieldcoinbase.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_shieldcoinbase.rs
@@ -44,7 +44,6 @@ use crate::{
             utils::{JsonZec, value_from_zatoshis},
         },
         keystore::KeyStore,
-        sync::SyncStatusReader,
     },
     prelude::*,
 };
@@ -150,7 +149,6 @@ pub(crate) async fn call<C: Chain>(
     mut wallet: DbHandle,
     keystore: KeyStore,
     chain: C,
-    sync_status: SyncStatusReader,
     fromaddress: String,
     toaddress: String,
     fee: Option<JsonValue>,
@@ -286,13 +284,6 @@ pub(crate) async fn call<C: Chain>(
         shielding_utxos,
         shielding_value: value_from_zatoshis(shielding_value_zats),
     };
-
-    // The proposal has been built. Before fetching key material and handing off the
-    // execution future, refuse the spend if the wallet is not synced enough to trust its
-    // view of the chain. This check is synchronous — it runs during the `call().await` in
-    // the dispatch layer, before the async operation is queued — so the caller sees the
-    // failure immediately rather than as a failed background operation.
-    sync_status.ensure_available()?;
 
     // Derive the spending key for the source account.
     let derivation = account.source().key_derivation().ok_or_else(|| {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,10 @@ be considered breaking changes.
 - A global sync lock that blocks wallet usage while Zallet’s view of the chain
   is not trustworthy. The balance and spend RPC methods (`z_getbalances`,
   `z_getbalanceforaccount`, `z_gettotalbalance`, `z_listunspent`,
-  `z_getnotescount`, `z_sendmany`, `z_shieldcoinbase`) return an error both while
-  the wallet is still catching up to the chain tip (analogous to `zcashd`’s
-  initial block download) and while it is recovering from a chain reorganization
-  (analogous to safe mode).
+  `z_getnotescount`, `z_sendmany`, `z_shieldcoinbase`, `pczt_create`,
+  `pczt_extract`) return an error both while the wallet is still catching up to
+  the chain tip (analogous to `zcashd`’s initial block download) and while it is
+  recovering from a chain reorganization (analogous to safe mode).
 - The `sync.lock_threshold` config option, controlling how far the wallet may
   lag the chain tip before it is considered to be catching up rather than synced
   (default 100).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,21 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Added
+
+- A global sync lock that blocks wallet usage while Zallet’s view of the chain
+  is not trustworthy. The balance and spend RPC methods (`z_getbalances`,
+  `z_getbalanceforaccount`, `z_gettotalbalance`, `z_listunspent`,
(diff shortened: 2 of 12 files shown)


zallet-core/src/components/json_rpc/methods/z_shieldcoinbase.rs
```diff
@@ -280,6 +282,13 @@ pub(crate) async fn call<C: Chain>(
         shielding_value: value_from_zatoshis(shielding_value_zats),
     };
 
+    // The proposal has been built. Before fetching key material and handing off the
+    // execution future, refuse the spend if the wallet is not synced enough to trust its
+    // view of the chain. This check is synchronous — it runs during the `call().await` in
+    // the dispatch layer, before the async operation is queued — so the caller sees the
+    // failure immediately rather than as a failed background operation.
+    sync_status.ensure_available()?;
```
schell: Here and above in `z_sendmany` - is there a reason not to do this check sooner, before constructing the proposal? Having this here means that proposal construction may fail because it's too far behind the chain state and the error would be misleading (something like "insufficient funds").

schell (approved):
Looks good! I had a nit/question. Might be a bit of an improvement but nothing to hold up for.

zallet-core/src/components/json_rpc/methods/z_shieldcoinbase.rs
sellout: Thanks – it should definitely be earlier. It’s already earlier for other methods, it simplifies the code, and it prevents misleading errors (like insufficient funds when searching for inputs that haven’t synced yet).

Just pushed a fix.

sellout:
I also added a check before some of the new PCZT methods that got merged yesterday.

sellout:
Looks like the main branch is failing, independent of this PR.

PR #676, opened 2026-07-24
ci: harden release workflow — pipe signing key, pin cargo-deb

y4ssi:
## Summary

Two hardening changes to the release workflow:

**Signing key never written to disk**

Both the matrix  step and the  job now pipe  directly into . The key is never materialized as a file on the runner filesystem.

Also adds  to the import step so an AWS CLI failure aborts the job rather than silently proceeding.

**cargo-deb pinned to an exact version**

Changed  to  so the installed build tool is a specific, reviewable version rather than any semver-compatible release.

## Test plan

- [ ] Confirm no  /  anywhere in either workflow
- [ ] Confirm cargo-deb installs exactly 
- [ ] Tag-push release works end-to-end
- [ ] Manual re-release () still works

### DIFF
--- a/.github/workflows/binaries-and-deb-release.yml
+++ b/.github/workflows/binaries-and-deb-release.yml
@@ -203,6 +203,7 @@ jobs:
 
       - name: Import GPG keys
         run: |
+          set -euo pipefail
           # Pipe directly from Secrets Manager to avoid writing the private key
           # to a plaintext file in the runner working directory.
           aws secretsmanager get-secret-value \
@@ -666,9 +667,7 @@ jobs:
         run: |
           set -euo pipefail
           aws secretsmanager get-secret-value --secret-id /release/gpg-signing-key \
-            --query SecretString --output text > /tmp/k.pgp
-          gpg --batch --import /tmp/k.pgp
-          shred -u /tmp/k.pgp
+            --query SecretString --output text | gpg --batch --import
       - name: Build distribution list
         run: |
           set -euo pipefail
--- a/.github/workflows/binaries-and-deb-release.yml
+++ b/.github/workflows/binaries-and-deb-release.yml
@@ -178,7 +178,7 @@ jobs:
           persist-credentials: false
 
       - name: Install Rust stable toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@4cda84d5c5c54efe2404f9d843567869ab1699d4 # stable
 
       - name: Install build tooling
         run: |
@@ -196,21 +196,19 @@ jobs:
             wget
 
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
         with:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
           aws-region: us-east-1
 
-      - name: Retrieve GPG signing key from Secrets Manager
+      - name: Import GPG keys
         run: |
+          # Pipe directly from Secrets Manager to avoid writing the private key
+          # to a plaintext file in the runner working directory.
           aws secretsmanager get-secret-value \
             --secret-id /release/gpg-signing-key \
             --query SecretString \
-            --output text > private.pgp
-
-      - name: Import GPG keys
-        run: |
-          gpg --import private.pgp
+            --output text | gpg --import
           wget -qO - https://apt.z.cash/zcash.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
 
       - name: Install cargo-deb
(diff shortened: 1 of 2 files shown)


pacu (commented):
## Overview

Reconciling the rendered diff against the branch topology turned out to matter here, so leading with that.

**Actual net change vs current `main`** — exactly three hunks in `.github/workflows/binaries-and-deb-release.yml`:

1. `binaries_deb_release`: merges "Retrieve GPG signing key" + "Import GPG keys" into one step that pipes `aws secretsmanager get-secret-value … | gpg --import`, plus `set -euo pipefail`.
2. `apt_publish`: same pipe treatment; drops the `> /tmp/k.pgp` … `shred -u` dance.
3. `cargo install cargo-deb --version "^2.10"` → `"=2.10.0"`.

`release.yml` nets to zero — the tag-verification commit (b6ae87d) is cleanly reverted by 15c5ac4, no residue.

---

## ⚠️ Main issue: stale branch, misleading diff

```
merge-base(main, HEAD) = 7b8ca04
commits in main not in this branch = 142
```

The "Files changed" view shows ~50 lines of **unrelated** churn as additions — the standalone-tarball restructuring, `dtolnay/rust-toolchain@4cda84d…` pinning, `aws-actions/configure-aws-credentials@7474bc4…` pinning, the SBOM/provenance rework. None of that is this PR's work. Commit fffd8a0 ("avoid writing GPG signing key to disk", 49+/54−) hand-carries a copy of the tarball work that `main` already has via bba50c6 — and bba50c6 is *not* an ancestor of this branch.

Why it matters:

- Reviewers can't distinguish the 3 real hunks from the 50 lines of duplicate.
- The merge base lacks the tarball code while *both* sides added it independently. Git will likely auto-resolve (identical content), but that's luck rather than design — if the copies ever drift, the merge silently picks a stale variant of the release-artifact logic.

**Please rebase onto `main` and force-push.** This should be a 3-hunk diff. The body's test-plan checklist also rendered with empty code spans and is worth fixing at the same time.

---

## Correctness

✅ **The pipe is fail-safe, and `pipefail` is why.** GitHub's default `run` shell is `bash -e {0}` — `-e` was already on, but `pipefail` was **not**. Without it, `aws … | gpg --import` reports only `gpg`'s status and an AWS failure passes silently. So `set -euo pipefail` is load-bearing here, not cosmetic. The PR body credits `set -e`; the meaningful addition is `pipefail`.

✅ **No leftovers.** Verified across all workflows: zero remaining references to `private.pgp`, `k.pgp`, or `shred`. Both key-import sites (lines 209 and 669) are converted — test-plan item #1 confirmed.

✅ **`cargo-deb 2.10.0` exists and is not yanked** (published 2025-01-07), and `=2.10.0` is valid `--version` syntax.

⚠️ **Inconsistent `gpg` invocation.** `apt_publish` uses `gpg --batch --import`; the matrix job uses bare `gpg --import`. Pre-existing, but the line is already being touched — worth adding `--batch` for guaranteed non-interactive behaviour and parity:

```yaml
    --output text | gpg --batch --import
```

---

## Security

Real improvement, but the framing overstates it.

**"The key is never materialized as a file on the runner filesystem" isn't accurate.** `gpg --import` writes the secret key straight to `~/.gnupg/private-keys-v1.d/` — it *is* on disk, just in the keyring rather than the workspace. Runners are ephemeral `ubuntu-latest`, so exposure is low either way.

What the change genuinely buys: `private.pgp` no longer sits in the **checkout working directory** for the whole build. I checked whether that was an active leak — it wasn't (`cp standalone/*`, `path: release/…/*`, and the `tar -C tarball` staging are all scoped; none glob the workspace root). So this is defense-in-depth against a future `upload-artifact: path: .` or a `cargo-deb` asset glob, not a fix for live exposure. Still worth doing — just worth describing as such.

Dropping `shred -u` is the right call: `shred` on ext4/overlayfs is unreliable anyway, and not creating the file beats trying to erase it.

**Optional:** neither job tears down the keyring. Cosmetic on ephemeral runners, but an `if: always()` cleanup step is cheap insurance if self-hosted runners ever enter the picture.

---

## The `cargo-deb` pin

Right direction, and `=2.10.0` + `--locked` is a strong reproducibility story. Two caveats:

- **2.10.0 is the *oldest* 2.x release** (2025-01-07). Newer 2.x exist — 2.11.x, 2.12.1 — so this forgoes bugfixes inside the range the existing comment says is compatible. If 2.12.1 still satisfies the MSRV constraint, that's the better pin. (Note 2.11.4 is yanked, so the old `^2.10` range could have resolved to it — another point for pinning.)
- **The comment no longer matches the code.** It explains why *2.x* rather than 3.x, but not why **2.10.0 specifically**. Exact pins rot silently — nothing will ever bump this. Either record the rationale or add a Renovate/Dependabot rule.

Suggested tweak:

```yaml
        # Pinned exactly (not a `^` range) so the build tool is a specific,
        # reviewable version. 3.x bumped its MSRV past the runner's stable
        # toolchain ("error[E0658]: `let` expressions in this position are
        # unstable"). Bump deliberately; nothing auto-updates this.
        run: cargo install cargo-deb --version "=2.10.0" --locked
```

---

## Verdict

The three substantive changes are correct and well-motivated — happy to approve them on their own. The blocking concern is process rather than code: **rebase onto `main`** so the diff shows 3 hunks instead of 50 lines of duplicated history, then adjust the two overstated claims in the description. The `--batch` flag and the pin-rationale comment are worth folding in while you're in there.

pacu:
Opened the two nits from my review as an opt-in follow-up against this branch rather than pushing to it: y4ssi/wallet-ci-test#1 (`gpg --batch --import` for parity with `apt_publish`, and a comment recording why the cargo-deb pin is exact). Nothing changes here unless you merge it.

The rebase-onto-`main` point from the review is the one that still needs you — that's what's inflating the diff on this PR.

pacu (approved):
utACK

PR #677, opened 2026-07-25
security hardening improvements

nullcopy:
## Summary

Closes security findings surfaced by a third-party audit. Each commit maps
to one finding; see the finding-to-commit map below. Rebased onto current
`main` (post-0.1.0-beta.2, post-MSRV-1.95).

### Privacy policy enforcement across all shielded pools
`e966780` — Replaces the pairwise Sapling↔Orchard checks in
`enforce_privacy_policy` and `required_privacy_policy` with a single
`shielded_pool_crossed_into(step)` helper that iterates
`all_shielded_pools()`. The pool list is an exhaustive match, so adding a
`ShieldedPool` variant fails to compile instead of silently bypassing the
gate. Ironwood and any future pool are covered by construction.

**Wire-compat note:** `RevealingSaplingAmount` / `RevealingOrchardAmount`
collapse into `RevealingShieldedAmount(ShieldedPool)`. The error *code* is
unchanged, but the error *message text* is now templated by
`PoolType::Shielded(pool).to_string()` rather than a fixed "…Sapling…" /
"…Orchard…" string. Clients that string-match on the message will see
different wording.

### Per-pool action limit
`c951703` — The `builder.limits.orchard_actions` transaction-size cap now
applies to each shielded pool's spends and outputs, not only Orchard's.

**Note:** The config key is intentionally retained as
`orchard_actions` for compatibility, but its semantics now cover Sapling
and Ironwood as well. The error message still references
`-orchardactionlimit=N`, which is the correct key but reads as a misnomer
for non-Orchard pools.

### Constant-time seed fingerprint comparison
`765da78` — `seed_matches_fingerprint` now uses `subtle::ConstantTimeEq`
instead of a short-circuiting `==`. Also adds a DFVK binding check for
decrypted standalone Sapling `extsk`: after decrypting, verifies
`extsk.to_diversifiable_full_viewing_key().to_bytes() == dfvk_array`
before returning, matching the checks already done for mnemonics, legacy
seeds, and transparent keys.

### Reject empty identity passphrases
`f7998db` — `generate-encryption-identity` now rejects empty passphrases
on both the `--passphrase-file` and interactive paths. An empty passphrase
previously produced an identity file that parsed as encrypted while being
trivially decryptable.

### Filesystem permissions
`92ca36c` — Datadir set to `0700` on lock; `wallet.db` created with
`create_new(true).mode(0o600)` or chmod'd to `0600` before SQLite opens it,
so journal/WAL/sidecar files inherit the database file's mode. Zero-length
file treated as missing to avoid a permissions window.

### Method-name-only RPC logging
`b0f79a4` — Replaces jsonrpsee's `rpc_logger(1024)` (which logged full call
parameters and responses) with `MethodLoggerMiddleware` that logs only the
method name and `is_error`. Compatibility middleware no longer logs full
replaced error objects — only error codes.

### Drop env-var passphrase, add `--passphrase-file`
`1477704` — `ZALLET_IDENTITY_PASSPHRASE` environment variable removed
(visible to sibling processes). `--passphrase-file PATH` reads the first
line from `-` (stdin) or a file. Buffer is `Zeroizing`.

### Reject empty RPC passwords
`fa29ee2` — `add-rpc-user` refuses an empty typed password; bare-password
`rpc.auth` config entries with an empty `password` fall through to the
existing error arm, so `AuthorizationLayer::new` returns
`err-init-rpc-auth-invalid`.

### Unit tests for privacy policy and action limit (review feedback)
`470fd2e` — Extracts the per-pool action limit check from `z_sendmany`
into `check_shielded_action_limits`, a pure function of the proposal, and
adds a `proposal_policy_tests` module that builds proposals in-memory from
dummy notes (the pattern used by `zcash_client_backend`'s own proposal
tests). Covers: every ordered pair of distinct shielded pools triggers
`RevealingShieldedAmount(pool)` under `FullPrivacy` (including the
Ironwood pairs the old pairwise checks missed) and passes under
`AllowRevealedAmounts`; same-pool steps satisfy `FullPrivacy`;
`required_privacy_policy` agreement; and the action limit for spends,
outputs, and combined actions in each pool, per-pool independence, and
the at-the-limit boundary.

## Finding → Commit Map

| Finding | Commit |
|---|---|
| Privacy policy misses Ironwood / future pools | `e966780` |
| Action limit only counts Orchard | `c951703` |
| Non-constant-time seed fingerprint compare; missing DFVK binding check | `765da78` |
| Empty passphrase produces trivially-decryptable "encrypted" identity | `f7998db` |
| Overly permissive datadir / wallet.db permissions | `92ca36c` |
| JSON-RPC logging leaks full parameters and responses | `b0f79a4` |
| `ZALLET_IDENTITY_PASSPHRASE` env var visible to sibling processes | `1477704` |
| Empty RPC password opens unauthenticated interface | `fa29ee2` |
| CHANGELOG documentation | `5807980` |
| CI: format imports + sync zaino lockfile | `d7d47ba` |
| Review feedback: unit tests for privacy policy + action limit | `470fd2e` |

## Known Follow-ups

- **Empty-RPC-password config-load path downgrades to a generic config
  parse error** (`err-init-rpc-auth-invalid`) rather than a targeted
  "empty password" message, unlike the `add-rpc-user` path which has a
  dedicated message.
- **`all_shielded_pools()` is `pub(super)` and duplicated in concept** —
  parallel pool enumeration exists in `view_transaction.rs`,
  `list_unspent.rs`, `get_notes_count.rs`. A single source of truth in
  `zcash_protocol` would prevent the next pool-addition bug.

### DIFF
--- a/zallet-core/src/components/json_rpc/payments.rs
+++ b/zallet-core/src/components/json_rpc/payments.rs
@@ -577,18 +577,17 @@ fn step_privacy_requirement<NoteRef>(
             PrivacyPolicy::AllowRevealedRecipients,
             IncompatiblePrivacyPolicy::TransparentChange,
         ))
-    } else if let Some(pool) = shielded_pool_crossed_into(step) {
-        // TODO: This should only trigger when there is a non-fee valueBalance.
-        // TODO: Determine whether this is due to the presence of an explicit
-        // recipient address in that pool, or having insufficient funds to pay a
-        // UA within a single pool.
-        Some((
-            PrivacyPolicy::AllowRevealedAmounts,
-            IncompatiblePrivacyPolicy::RevealingShieldedAmount(pool),
-        ))
     } else {
-        // Nothing is revealed by this step.
-        None
+        shielded_pool_crossed_into(step).map(|pool| {
+            // TODO: This should only trigger when there is a non-fee valueBalance.
+            // TODO: Determine whether this is due to the presence of an explicit
+            // recipient address in that pool, or having insufficient funds to pay a
+            // UA within a single pool.
+            (
+                PrivacyPolicy::AllowRevealedAmounts,
+                IncompatiblePrivacyPolicy::RevealingShieldedAmount(pool),
+            )
+        })
     }
 }
 
--- a/zallet-core/src/components/json_rpc/payments.rs
+++ b/zallet-core/src/components/json_rpc/payments.rs
@@ -678,6 +678,64 @@ pub(super) fn required_privacy_policy<FeeRuleT, NoteRef>(
         })
 }
 
+/// A shielded pool in which a proposal step exceeds the per-pool action limit.
+#[derive(Debug, PartialEq, Eq)]
+pub(super) struct ShieldedActionLimitExceeded {
+    pub(super) pool: ShieldedPool,
+    pub(super) count: usize,
+    /// Which side of the pool's bundle exceeds the limit: `"inputs"`, `"outputs"`, or
+    /// `"actions"` when both do.
+    pub(super) kind: &'static str,
+}
+
+/// Checks every step of the proposal against the per-pool action limit.
+///
+/// The limit applies per shielded pool: each pool's spends and outputs bound the memory
+/// needed to prove and construct that pool's part of the transaction.
+pub(super) fn check_shielded_action_limits<FeeRuleT, NoteRef>(
+    proposal: &Proposal<FeeRuleT, NoteRef>,
+    limit: usize,
+) -> Result<(), ShieldedActionLimitExceeded> {
+    for step in proposal.steps() {
+        for pool in all_shielded_pools() {
+            let spends = step
+                .shielded_inputs()
+                .iter()
+                .flat_map(|inputs| inputs.notes())
+                .filter(|note| note.note().pool() == pool)
+                .count();
+
+            let outputs = step
+                .payment_pools()
+                .values()
+                .filter(|payment_pool| **payment_pool == PoolType::Shielded(pool))
+                .count()
+                + step
+                    .balance()
+                    .proposed_change()
+                    .iter()
+                    .filter(|change| change.output_pool() == PoolType::Shielded(pool))
+                    .count();
+
+            let actions = spends.max(outputs);
+
+            if actions > limit {
+                let (count, kind) = if outputs <= limit {
+                    (spends, "inputs")
+                } else if spends <= limit {
+                    (outputs, "outputs")
+                } else {
+                    (actions, "actions")
+                };
+
+                return Err(ShieldedActionLimitExceeded { pool, count, kind });
+            }
+        }
+    }
+
+    Ok(())
+}
+
 /// Parses the optional `privacy_policy` JSON-RPC argument into a [`PrivacyPolicy`],
 /// defaulting to [`PrivacyPolicy::FullPrivacy`] when absent and rejecting the unsupported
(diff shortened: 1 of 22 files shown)


zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -326,46 +329,51 @@ pub(crate) async fn call<C: Chain>(
...
+
+                return Err(LegacyCode::Misc.with_message(fl!(
+                    "err-excess-shielded-actions",
+                    pool = PoolType::Shielded(pool).to_string(),
+                    count = count,
+                    kind = kind,
+                    limit = actions_limit,
+                    config = "-orchardactionlimit=N",
+                    bound = format!("N >= %u"),
+                )));
```
schell: Similarly to ~`zallet-core/src/components/json_rpc/payments.rs:319` - since we're adding this code to catch this case we should test this case. But yeah, building all the stuff is difficult.

https://github.com/zcash/zallet/issues/693

zallet-core/src/components/json_rpc/payments.rs
```diff
@@ -338,18 +367,13 @@ pub(super) fn enforce_privacy_policy<FeeRuleT, NoteRef>(
             if !privacy_policy.allow_revealed_recipients() {
                 return Err(IncompatiblePrivacyPolicy::TransparentChange);
             }
-        } else if step.input_in_pool(PoolType::ORCHARD) && has_sapling_recipient {
+        } else if let Some(pool) = shielded_pool_crossed_into(step) {
```
schell: It would be nice to be able to unit test that this returns `RevealingShieldedAmount` when crossing pools, but we need some better test helpers in this repo to accomplish that. 

https://github.com/zcash/zallet/issues/693

nullcopy: Added in 470fd2e: it turned out to be feasible without new repo-level test infrastructure. `zcash_client_backend`'s own proposal tests construct `Step`s in-memory from dummy notes via public constructors (`Note::from_parts` with fixed keys/randomness, `ReceivedNote::from_parts`, `Proposal::single_step`), so I followed that pattern in a new `proposal_policy_tests` module. `crossing_into_any_other_pool_requires_revealed_amounts` now asserts `Err(RevealingShieldedAmount(to))` under `FullPrivacy` for every ordered pair of distinct shielded pools (including the Ironwood pairs the old pairwise checks missed), plus `Ok` under `AllowRevealedAmounts`, the same-pool case, and `required_privacy_policy` agreement. Dedicated `Step`/`Proposal` builders per #693 would still make future tests nicer, but the specific coverage requested here is in.

zallet-core/src/components/json_rpc/methods/z_send_many.rs
nullcopy: Added in 470fd2e. To make this testable without standing up a wallet, I extracted the loop into `check_shielded_action_limits(&proposal, limit)` in `payments.rs` — a pure function returning a typed `ShieldedActionLimitExceeded { pool, count, kind }` — with `z_sendmany` keeping only the error-message formatting. Tests cover spends, outputs, and combined-actions overflow for each of the three shielded pools, the at-the-limit boundary, and that the limit is per pool (spends spread across pools may exceed it in aggregate).

schell: Nice solution :)

schell (approved):

Issue #256, opened 2025-09-10
Improve internal error types


str4d:
There are three kinds of errors we need to deal with:
- Errors returned to users from CLI commands.
- Errors returned to users via JSON-RPC.
- Errors produced by internal components that are consumed by other internal components (usually by being converted to one of the other two kind of errors).

Currently we are using an `ErrorKind` enum for the first type of errors and converting to the second kind as needed:
https://github.com/zcash/wallet/blob/2eb2d4b027195f664b4a0fe5ea3877402e00e37a/zallet/src/components/json_rpc/methods/z_send_many.rs#L193-L203

We should instead have comprehensive granular errors internally, and then map them to the appropriate user-facing errors at the necessary boundary.

PR #679, opened 2026-07-25
Implement z_sendfromaccount on the PCZT pipeline

### LINKED ISSUE #217: rpc: Design and implement `z_sendfromaccount`

nullcopy:
Implements z_sendfromaccount. This one-shot method is a composition of: build_pczt (shared with pczt_create, now also returning the proposal) → pczt_prove → pczt_sign (strict, acknowledging the caller's policy) → pczt_extract (which records the transaction in the wallet) → the shared verify-and-broadcast path used by z_sendmany.

The fund_source parameter is the isolating source selector ported with the fixups: a source that cannot cover the payment reports insufficient funds rather than drawing on the account's other pools, and "orchard" includes Ironwood, where an account's Orchard-receiver funds live once NU6.3 activates.

Closes https://github.com/zcash/zallet/issues/217
Subsumes https://github.com/zcash/zallet/pull/531
[COR-300](https://linear.app/zodl/issue/COR-300/rpc-design-and-implement-z-sendfromaccount)

### DIFF
--- a/book/src/rpc/methods.md
+++ b/book/src/rpc/methods.md
@@ -624,6 +624,39 @@ Use the `z_getaddressforaccount` RPC method to obtain addresses for an account.
   - `zip32_account_index` (numeric, required)
   - `birthday_height` (numeric, required)
 
+## `z_sendfromaccount`
+
+*Only available in wallet builds of Zallet.*
+
+Sends funds from an account in one shot, signing and broadcasting the
+transaction.
+
+Unlike `z_sendmany`, the source of funds is an account UUID together
+with an explicit fund source, rather than an address. The transaction is
+assembled through the same pipeline as the `pczt_*` methods (create,
+prove, sign, extract); use those methods directly to inspect a
+transaction before executing it. Unlike `z_sendmany`, the call completes
+synchronously, returning the txid rather than an operation id.
+
+#### Arguments
+- `account` (string, required) The UUID of the account to send the funds
+  from.
+- `fund_source` (string or array, required) Where funds may be drawn
+  from. One of the strings `"orchard"`, `"sapling"`,
+  `"any_transparent"`, or an array of transparent address strings. Each
+  source is isolating: a source that cannot cover the payment reports
+  insufficient funds rather than drawing on the account's other funds.
+  `"orchard"` includes the Ironwood pool, where an account's
+  Orchard-receiver funds are held once NU6.3 activates.
+- `recipients` (array, required) An array of JSON objects representing
+  the amounts to send, with the same shape as `z_sendmany`'s `amounts`.
+- `minconf` (numeric, optional) Only use funds confirmed at least this
+  many times.
+- `privacy_policy` (string, required) Policy for what information
+  leakage is acceptable, using the same values as `z_sendmany`.
+  Required: the send is not proposed for review first, so the caller
+  must acknowledge its privacy implications up front.
+
 ## `z_sendmany`
 
 *Only available in wallet builds of Zallet.*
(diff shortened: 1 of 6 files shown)


zallet-core/src/components/json_rpc/methods/z_send_from_account.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    let seed = keystore
+        .decrypt_seed(derivation.seed_fingerprint())
+        .await
+        .map_err(|e| match e.kind() {
+            // TODO: Improve internal error types.
+            //       https://github.com/zcash/zallet/issues/256
+            crate::error::ErrorKind::Generic if e.to_string() == "Wallet is locked" => {
+                LegacyCode::WalletUnlockNeeded.with_message(e.to_string())
+            }
```
schell, line 119: Oof, yeah #256 should get dealt with.

zallet-core/src/components/json_rpc/methods/z_send_from_account.rs
```diff
@@ -0,0 +1,173 @@
...
+    // broadcast.
+    let extracted = pczt_extract::call(wallet_handle(&wallet).await?, &signed.pczt).await?;
+
+    // The user-facing txid string is byte-reversed hex, per Bitcoin convention.
+    let txid = {
+        let mut bytes = [0u8; 32];
+        hex::decode_to_slice(&extracted.txid, &mut bytes)
+            .map_err(|e| LegacyCode::Misc.with_message(e.to_string()))?;
+        bytes.reverse();
+        TxId::from_bytes(bytes)
```
schell: I think you can use `ReverseHex::decode` here. That would also make the error handling there go away.

zallet-core/src/components/json_rpc/methods/z_send_from_account.rs
schell: I think right now since the only translation we have is `en-US` this is fine, but it's a bit of a time bomb. The problem is #256, though.

schell (dismissed):
I think this is good. 

One thing to think about is that there's no guard for the maximum recipients. `z_sendmany` has a TODO for it, and `pczt_create::call` does it, but if `build_pczt` _also_ did it, that would solve the `z_sendmany` TODO and help `z_sendfromaccount`.

book/src/rpc/methods.md
```diff
@@ -624,6 +624,39 @@ Use the `z_getaddressforaccount` RPC method to obtain addresses for an account.
   - `zip32_account_index` (numeric, required)
   - `birthday_height` (numeric, required)
 
+## `z_sendfromaccount`
+
+*Only available in wallet builds of Zallet.*
```
pacu, line 629: are there non wallet builds of Zallet? 🤔

pacu (approved):
non blocking comment  about confusing docs

book/src/rpc/methods.md
nullcopy: Ha, I guess it depends how you define that. That caveat is listed on every other RPC, and I see a few other confusing docs. We should probably give these docs a review/clean-up.

PR #680, opened 2026-07-25
feat: dedicated-port /ready and /healthy health probes (#366)

### LINKED ISSUE #366: Add readiness and liveness endpoints
HTTP endpoints used to assess the readiness and liveness state of daemons running across the Z3 stack should be available and consistent. Readiness indicates "should this service receive traffic?" while liveness indicates answers "is this service broken? (and potentially needs its orchestrator to restart it?)"

Zallet should expose GET /ready and GET /healthy, ideally running on a dedicated port for security. (["readyz" and "healthz" are even more standard](https://github.com/ZcashFoundation/zebra/pull/9895#discussion_r2361125183) but I would make this match Zebra, and eventually Zaino when it has readiness and liveness)

This is a common ops pattern for container probes and load balancers, and it also keeps us aligned with [what Zebra already does.](https://zebra.zfnd.org/user/health.html)

User stories:
- As an infrastructure engineer responsible for ensuring that zallet is online and reliable for my users, I need to point my orchestration stack at readiness and liveness endpoints in order to know if zallets is running in a healthy state, and to know when it is ready to accept end-user connections via a load balancer.
- As an infrastructure engineer I want to configure my observability tools to notify the team when our zallet deployment is unhealthy or not ready to receive user traffic.
- As a privacy-conscious person wanting to self-host zallet, I want a simple "docker compose up" way to launch an entire Z3 stack, and know that each service within the stack only communicates once it is ready, using readiness and liveness endpoints behind the scenes to define healthy behavior in its docker-compose.yml or similar.
- As an infrastructure engineer exposing zallet RPCs to other teams within my organization, I want to ensure that readiness and liveness endpoints are only accessible on a dedicated port so that I have full control over my zallet deployment's security.

Related: 
- https://github.com/zcash/wallet/issues/316
- https://github.com/ZcashFoundation/zebra/pull/9895
- https://github.com/zingolabs/zaino/issues/643

AnubisQuantumCipher:
## Summary

Implements #366: HTTP readiness and liveness endpoints for Zallet, on a **dedicated port** so ops probes do not share the auth-gated JSON-RPC surface.

Matches the Zebra probe path names (`GET /ready`, `GET /healthy`) and also accepts common k8s aliases (`/readyz`, `/healthz`).

### Config (opt-in; disabled by default)

```toml
[health]
listen_addr = "127.0.0.1:8232"
```

### Semantics

| Endpoint | Meaning | Status |
|----------|---------|--------|
| `GET /healthy` | Process + health server up | `200 ok` |
| `GET /ready` | Wallet DB pool can hand out a handle (usable for traffic) | `200 ready` / `503 not ready` |

### Design notes

- Unauthenticated by design (same security model as Zebra health). Bind loopback / private interfaces only.
- Readiness is intentionally DB-level first: if the pool cannot open a handle, orchestrators should not route traffic. Sync-depth readiness can be layered later if maintainers want Zebra-style tip lag.

## Test plan

- [x] `cargo check -p zallet-core` clean on this branch
- [ ] Config with `health.listen_addr` set: `curl -i http://127.0.0.1:8232/healthy` → 200
- [ ] Same: `curl -i http://127.0.0.1:8232/ready` → 200 when wallet DB is open
- [ ] Health unset: no listener (existing configs unchanged)

Happy to adjust default port, readiness criteria, or docs path to match Zebra/Zaino conventions.


zallet-core/src/components/health.rs
```diff
@@ -0,0 +1,140 @@
...
+    let state = Arc::new(HealthState { db });
+
+    Ok(crate::spawn!("health server", async move {
+        loop {
+            let (stream, peer) = match listener.accept().await {
+                Ok(conn) => conn,
+                Err(e) => {
+                    warn!(?e, "health server accept failed");
+                    continue;
+                }
```
pacu, line 62: **The accept loop can busy-spin at 100% CPU.**

`accept()` errors aren't all transient. On fd exhaustion (`EMFILE`/`ENFILE`/`ENOBUFS`) — which is reachable on this very listener, see the comment on the `tokio::spawn` below — `accept()` returns an error *immediately and continuously*, so this loop spins as fast as the CPU allows and writes a warning line per iteration. That starves the wallet sync and JSON-RPC tasks of the runtime and fills the log at disk-write speed.

Worth splitting transient errors (`ECONNABORTED`, `EINTR`) from resource errors, and sleeping with a backoff before retrying the latter.

zallet-core/src/components/health.rs
```diff
@@ -0,0 +1,140 @@
...
+            tokio::spawn(async move {
+                let io = TokioIo::new(stream);
+                let service = hyper::service::service_fn(move |req| {
+                    let state = state.clone();
+                    async move { handle(req, &state).await }
+                });
+                if let Err(e) = http1::Builder::new().serve_connection(io, service).await {
+                    warn!(?peer, ?e, "health connection error");
+                }
+            });
```
pacu, line 74: **Unbounded connections with no timeouts, on an unauthenticated port.**

Every accepted connection gets a bare `tokio::spawn`, and `http1::Builder::new()` sets no header-read, idle, or request timeout. Since this port is unauthenticated by design, anyone who can reach it can open connections that never send a request line and hold one fd plus one task each indefinitely (slowloris). The process eventually hits its fd limit — which then breaks the wallet database and the JSON-RPC listener too, not just this probe.

These handles are also untracked, so the `health_task_handle.abort()` on shutdown stops the accept loop but leaves live connections running.

Suggestions: bound concurrent connections (e.g. a `Semaphore`), set `http1::Builder::header_read_timeout(..)`, and consider `hyper_util::server::graceful` (feature `server-graceful`) so in-flight connections are shut down with the task.

zallet-core/src/components/health.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// Liveness: the process is up and the health server itself is serving.
+fn healthy() -> Response<Full<Bytes>> {
+    plain(StatusCode::OK, "ok\n")
+}
+
+/// Readiness: wallet database is openable (sync / RPC traffic can be served
+/// against a working DB). Returns 503 when the pool cannot hand out a handle.
+async fn ready(state: &HealthState) -> Response<Full<Bytes>> {
+    match state.db.handle().await {
```
pacu, line 107: **This makes `/ready` hang forever rather than return 503 — the opposite of the documented behaviour.**

`Database::handle()` is `deadpool` `Pool::get()`, and the pool is built with `PoolConfig::default()` (`components/database/connection.rs`). In deadpool 0.12 that is `Timeouts::default()`, documented as *"Creates an empty `Timeouts` config (no timeouts set)"* — so `wait` is `None`.

When the pool is saturated, `get()` waits indefinitely instead of erroring, so the `Err` arm below never runs and the probe never answers. The doc comment above promises "Returns 503 when the pool cannot hand out a handle", which is exactly the case that doesn't work.

There's a second-order problem: the pool queue is FIFO, so unauthenticated `/ready` traffic queues *ahead of* the sync engine and RPC handlers for wallet DB connections — an unauthenticated port can degrade the wallet.

Wrapping the call in `tokio::time::timeout(..)` and mapping elapsed → 503 fixes both. `Pool::status()` may be a better fit than actually taking a connection.

zallet-core/src/config.rs
```diff
@@ -896,6 +927,12 @@ impl ZalletConfig {
             ),
             rpc("bind", &conf.rpc.bind),
             rpc("timeout", conf.rpc.timeout().as_secs()),
+            health(
+                "listen_addr",
+                conf.health
+                    .listen_addr
+                    .unwrap_or_else(|| "127.0.0.1:8232".parse().expect("valid")),
+            ),
```
pacu, line 935: **This renders a default that doesn't exist, and the consequence is an unauthenticated listener.**

The generated example config's legend tells operators that a commented-out field shows "the current default value that Zallet will use for it (or `UNSET` if the field has no default)". This field has no default — unset means the health server never starts — but `unwrap_or_else(..)` makes it render as `#listen_addr = "127.0.0.1:8232"`.

So the example config states that Zallet is already listening there. Someone who uncomments it believing it's a no-op silently opens an unauthenticated HTTP port.

Passing `conf.health.listen_addr` through directly lets `field()` map `None` → `UNSET`, which is how `external("export_dir", ..)` and `indexer("validator_address", ..)` already behave.

zallet-core/src/config.rs
```diff
@@ -746,6 +752,31 @@ impl NoteManagementSection {
     }
 }
 
+/// Settings for HTTP readiness and liveness endpoints.
+///
+/// Disabled by default. Enable by setting a listen address:
+/// ```toml
+/// [health]
+/// listen_addr = "127.0.0.1:8232"
```
pacu, line 760: **8232 is the mainnet RPC port for zcashd/Zebra**, as this repo notes at `backends/zaino/src/chain.rs:143`. Zallet's own examples deliberately use `28232` (see `config.rs` further down, and `book/src/troubleshooting.md`).

`zallet-zebra` typically runs on the same host as `zebrad`, so an operator copying this line gets `Address already in use` from `TcpListener::bind` — and because that's mapped to `ErrorKind::Init`, it aborts `zallet start` outright. A probe misconfiguration taking down the wallet is a rough failure mode.

Something in Zallet's own range, e.g. `28233`, would avoid it.

pacu (commented):
Thanks for this, and apologies it sat unreviewed for so long — the shape is right and it matches Zebra's conventions as the issue asked.

I've left specific comments inline. Three are worth calling out as blocking, all in `health.rs`: the accept loop can busy-spin on fd exhaustion, connections are unbounded and untimed on an unauthenticated port, and `/ready` hangs instead of returning 503 when the DB pool is saturated (deadpool's default config sets no wait timeout). Those three compound — the second causes the fd exhaustion that triggers the first.

There's also one config-generation issue that's security-relevant rather than cosmetic: the example config would advertise a default health address that doesn't exist, so uncommenting it silently opens an unauthenticated port.

**The one thing worth deciding before you write more code** is the readiness contract — see the comment on the `ready` doc comment. You offered tip-lag checks on the issue and I think that's the right instinct; agreeing the criteria now saves you a round.

Two things I couldn't anchor inline: the branch needs a rebase (`main` has since refactored `start.rs` task supervision — details in the comment there), and `CHANGELOG.md` wants an `Unreleased` entry, since a new config section plus two HTTP endpoints is exactly the user-facing surface it covers.

_(review prepared with Claude)_

zallet-core/src/config.rs
```diff
@@ -1216,6 +1260,7 @@ impl ZalletConfig {
                     write_section::<NoteManagementSection>(&mut config, field_name, &sec_def)
                 }
                 RPC => write_section::<RpcSection>(&mut config, field_name, &sec_def),
+                HEALTH => write_section::<HealthSection>(&mut config, field_name, &sec_def),
```
pacu, line 1263: **CI will be red until the snapshots are regenerated** — four cases, not one.

Adding a section changes generated-config output, and `backends/zebra/tests/cli_tests.rs` runs `trycmd` over `tests/cmd/*.toml`:

- `example_config.out/zallet.toml` gains a whole `[health]` section
- `migrate_zcash_conf_{mainnet,testnet,regtest}.toml` each gain an empty `[health]` table in expected stdout (those snapshots are a serialized `ZalletConfig`)

zallet-core/src/commands/start.rs
```diff
@@ -71,6 +72,9 @@ impl StartCmd {
         )
         .await?;
 
+        // Launch readiness / liveness HTTP probes (dedicated port; opt-in).
+        let health_task_handle = health::spawn(&config.health, db.clone()).await?;
```
pacu, line 76: **Heads-up for the rebase: this hunk won't apply, and the fix isn't textual.**

Since this branch's base, `main` replaced the inline `pin!`/`abort()` block with `StartupTaskOwner` + `supervise_zallet_tasks`. The health handle now needs `task_owner.include(&health_task_handle)` (plus a parameter and a `select!` arm in `supervise_zallet_tasks`) — otherwise the health task escapes cancellation when a later `?` (e.g. `WalletSync::spawn`) returns early.

Also worth matching the new arms' behaviour: they propagate the task's inner result, whereas the arm added here discards it as `Ok(())`, so a failing health task wouldn't make the process exit non-zero.

zallet-core/src/components/health.rs
```diff
@@ -0,0 +1,140 @@
...
+        _ => Ok(plain(StatusCode::NOT_FOUND, "not found\n")),
+    }
+}
+
+/// Liveness: the process is up and the health server itself is serving.
+fn healthy() -> Response<Full<Bytes>> {
+    plain(StatusCode::OK, "ok\n")
+}
+
+/// Readiness: wallet database is openable (sync / RPC traffic can be served
```
pacu, line 104: **Readiness contract question** (this is the one design decision I'd settle before another round).

As written, `/ready` returns 200 on a working DB alone. That means it reports ready when `rpc.bind` is empty and no RPC listener exists at all, and while the sync engine is far enough behind that `sync_status` gates balance and spend methods into errors. An orchestrator would route user traffic to an instance that answers nothing, or rejects everything.

Given the issue frames readiness as "should this service receive traffic?", including "the RPC server is actually bound" seems like the minimum. You offered tip-lag checks on the issue — that seems right too, and it'd be good to agree the criteria now so you only do one more pass.

zallet-core/src/components/health.rs
```diff
@@ -0,0 +1,140 @@
...
+            };
+            let state = state.clone();
+            tokio::spawn(async move {
+                let io = TokioIo::new(stream);
+                let service = hyper::service::service_fn(move |req| {
+                    let state = state.clone();
+                    async move { handle(req, &state).await }
+                });
+                if let Err(e) = http1::Builder::new().serve_connection(io, service).await {
+                    warn!(?peer, ?e, "health connection error");
```
pacu, line 72: This warns for every connection that ends in a hyper error, including a peer that connects and closes without sending a request — which is exactly what a Kubernetes `tcpSocket` probe, an L4 load-balancer check, or a port scanner does. At probe frequency that's a warning line every few seconds, forever.

`debug!` is probably the right level here, or filter the "closed before message completed" class specifically.

zallet-core/src/components/health.rs
```diff
@@ -0,0 +1,140 @@
...
+        let res = healthy();
+        assert_eq!(res.status(), StatusCode::OK);
+    }
+
+    #[test]
+    fn method_not_allowed_body() {
+        // path coverage is integration-level; unit-test response helper shape.
+        let res = plain(StatusCode::NOT_FOUND, "not found\n");
+        assert_eq!(res.status(), StatusCode::NOT_FOUND);
+    }
```
pacu, line 139: This doesn't test what its name says — it calls `plain(NOT_FOUND, ..)` and asserts 404, so the 405 branch is never exercised. Between this and `healthy_is_200`, the four route strings and the `/ready` 503 path have no coverage.

`handle()` is directly unit-testable with `Request::builder()`, no server needed. Worth covering the routing table — a typo in one of those four literals is precisely the bug that would cause a silent probe outage.

Cargo.toml
```diff
@@ -90,7 +90,9 @@ rpassword = "7"
 
 # RPC
 http-body-util = "0.1"
-hyper = "1"
+bytes = "1"
+hyper = { version = "1", features = ["server", "http1"] }
+hyper-util = { version = "0.1", features = ["tokio", "server-auto", "http1"] }
```
pacu, line 95: `server-auto` expands to `["server", "http1", "http2"]`, so this enables `hyper/http2` and pulls `h2` into the build — but the code only uses `hyper_util::rt::TokioIo` (feature `tokio`) and drives `hyper::server::conn::http1` directly.

Dropping `server-auto` should be a no-op here. `server-graceful` is likely the feature this component actually wants, for connection shutdown.

pacu (changes requested):
Marking this as request-changes so the state reflects where it stands — but to be clear about intent: **we'd like you to land this, and the PR is yours.** The design is right, it matches Zebra's conventions as #366 asked, and the dedicated-port/opt-in shape is exactly what the security user story wanted. Nothing below is a reason to start over.

I left detailed inline comments in my previous review. Consolidating what actually blocks:

**Blocking — all in `health.rs`, and they compound:**

1. Connections are unbounded and untimed on an unauthenticated port, so an idle-connection flood exhausts file descriptors — which takes down the wallet DB and the RPC listener, not just the probe.
2. Fd exhaustion then makes the accept loop busy-spin at 100% CPU, starving the sync and RPC tasks.
3. `/ready` hangs instead of returning 503 when the DB pool is saturated, because deadpool's default config sets no wait timeout. That's the one condition the endpoint exists to report.

**Blocking — config generation:** the example config would advertise a `health.listen_addr` default that doesn't exist, so an operator uncommenting it believing it's a no-op silently opens an unauthenticated port. That one is security-relevant rather than cosmetic.

**Needed to go green:** four `trycmd` snapshots to regenerate, and a rebase — `main` has since refactored task supervision in `start.rs`, so that hunk needs the new `StartupTaskOwner` API rather than a textual fixup. Happy to help with the rebase if it's fiddly.

**One decision before you write code.** The readiness contract is a product question, not a code problem, and it's worth settling first so you do one pass instead of two. As written, `/ready` returns 200 on a working DB alone — so it reports ready when no RPC listener is bound at all, and while the sync engine is far enough behind that balance and spend methods return errors. Our view: readiness should require at minimum that the RPC server is bound, and you were right to offer tip-lag checks on the issue — we'd take those. If you'd rather propose the exact criteria first and get a 👍 before implementing, that's welcome; a comment here is enough, no design doc needed.

**On timing:** we've moved #366 to the beta.4 milestone, so there's no deadline pressure on you. That's a scheduling decision on our side, not a comment on this PR.

Sorry again for the long silence after you did what you'd offered to do. Ping me here when you've pushed, or if you'd like to talk through any of the above — I'll turn re-reviews around quickly this time.

_(review prepared with Claude)_

PR #683, opened 2026-07-27
rpc: Apply both `z_listunspent` confirmation bounds to every pool

### LINKED ISSUE #682: z_listunspent ignores `minconf` for unmined shielded notes, and `maxconf` for transparent outputs
`z_listunspent` does not correctly enforce either of its documented confirmation bounds.

### `minconf` is not applied to notes whose transaction is not mined

The shielded filter in `zallet-core/src/components/json_rpc/methods/list_unspent.rs` tests the mined height of the note's transaction with `Option::iter().any(..)`:

```rust
let tx_mined_height = get_mined_height(*note.txid())?;
// ...
if tx_mined_height
    .iter()
    .any(|h| *h > target_height.saturating_sub(minconf))
    || maxconf.iter().any(|c| confirmations > *c)
{
    continue;
}
```

`Option::iter().any(..)` is equivalent to `is_some_and(..)`, and so is vacuously false when the transaction is not mined in the main chain. Such a note is therefore reported at *every* `minconf`, with a `confirmations` count of zero.

This filter is the only enforcement on the path. `InputSource::select_unspent_notes` takes no `ConfirmationsPolicy`, and the `zcash_client_sqlite` implementation returns the note even when its confirmations check fails, because the `NoteRequest::Unspent` arm of the eligibility match is:

```rust
(NoteRequest::Unspent, false) | (_, true) => Ok(Some(note)),
```

The transparent path, by contrast, is gated in SQL: a NULL mined height is admitted only when the query's `:min_confirmations` is zero.

#### Reachability

Notes received in mempool transactions are *not* affected: the selection query requires `nf IS NOT NULL`, and a non-NULL commitment tree position is required to reconstruct the note. Both are recorded only by block scanning, whereas a mempool transaction is stored via `decrypt_and_store_transaction`, whose `DecryptedOutput` supplies neither. (That mempool notes are consequently never listed at all, even at `minconf = 0`, is a separate divergence from `zcashd` and is not addressed here.)

The reachable case is a reorg. `truncate_to_height` clears `transactions.block` and `transactions.mined_height` but leaves each note's `nf` and `commitment_tree_position` intact, so a note that had been scanned in a mined block remains selectable, now with no mined height, until its containing block is re-scanned and while its transaction remains unexpired.

### `maxconf` is not applied to transparent outputs

The transparent loop computes `confirmations` but never compares it against `maxconf`; only the three shielded loops do. A transparent output with more than `maxconf` confirmations is reported regardless.

### Impact

This is a reporting defect; I do not believe it is exploitable beyond that:

- Input selection is unaffected. Spends go through `select_spendable_notes` under the configured `ConfirmationsPolicy`, and independently of that a shielded note cannot be spent until it is mined and its witness established, so an over-reported note cannot propagate into a transaction.
- The `confirmations` field is itself reported correctly as `0`, so a consumer that reads it is not misled; only one that relies on `minconf` as its gate is.

What remains is that a consumer using `minconf` as a confirmation gate — crediting deposits at N confirmations, say — can be handed a zero-confirmation output during a reorg, which is precisely when that gate is load-bearing. Worth fixing, but with no path to loss of funds.

### Suggested fix

Note that swapping `any` for `all` is not correct on its own: it would exclude unmined outputs at `minconf = 0`, which this RPC permits whenever `asOfHeight` is absent. Testing the already-computed confirmation count instead handles both bounds and both cases uniformly:

```rust
confirmations >= minconf && maxconf.is_none_or(|c| confirmations <= c)
```

where an output of an unmined transaction has zero confirmations.

---

Reported by a security researcher reviewing the beta; filed publicly with their agreement, as it carries no disclosure sensitivity. Issue drafted with Claude Code assistance.

nuttycom:
Closes #682

`z_listunspent` enforced neither of its documented confirmation bounds correctly.

The `minconf` filter tested the mined height of a note's transaction with `Option::iter().any(..)`, which is equivalent to `is_some_and(..)` and so vacuously false when that transaction is not mined in the main chain. Such a note was reported at every `minconf`, with a `confirmations` count of zero. `maxconf` was in turn never applied to transparent outputs at all.

Both bounds now come from a single predicate over the confirmation count, which each pool's loop consults:

```rust
fn confirmations_in_range(confirmations: u32, minconf: u32, maxconf: Option<u32>) -> bool {
    confirmations >= minconf && maxconf.is_none_or(|c| confirmations <= c)
}
```

Expressing the test in terms of the count rather than the mined height is what makes `minconf = 0` fall out correctly. Simply swapping `any` for `all` would have excluded unmined outputs at `minconf = 0`, which this RPC permits whenever `asOfHeight` is absent; against the count, an unmined transaction's outputs have zero confirmations, so they are in range for `minconf = 0` and out of range for any greater value, with no special case. This also removes the copy of the condition that each of the three shielded loops carried.

Since the RPC-level filter is the only enforcement on this path — `InputSource::select_unspent_notes` takes no `ConfirmationsPolicy`, and the `zcash_client_sqlite` implementation returns notes failing its confirmations check for `NoteRequest::Unspent` — the fix belongs here rather than in the data layer.

See #682 for the reachability analysis (mempool notes are excluded from this path by `nf IS NOT NULL`; the live case is a reorg un-mining a scanned transaction) and for why the impact stops short of a spend path.

## Testing

Eight unit tests over the two extracted functions cover the decision table: zero confirmations when unmined, the mining block counted, saturation at or above the target height (the `asOfHeight` case), exclusion at `minconf = 1` and inclusion at `minconf = 0`, and both bounds inclusive.

I confirmed the regression tests bite rather than merely restating the new code: substituting the pre-fix semantics back into the predicate makes `unmined_output_is_excluded_at_minconf_1` fail while the rest still pass.

One limit worth flagging for review: `maxconf_bound_is_inclusive` pins the predicate, but nothing exercises the transparent loop end to end, as the RPC methods have no in-crate database harness. That half of the fix is verified by construction — one predicate, four call sites — rather than by test. Genuine coverage there would want an integration test.

## Notes

No dependency changes; this builds against `main`'s existing `zcash_client_backend` 0.24.0-rc.1 pin, so it can land ahead of the librustzcash 0.24.0-rc.4 update.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,12 @@ be considered breaking changes.
 
 ### Fixed
 
+- `z_listunspent` now applies both confirmation bounds to every pool. Shielded
+  notes whose transaction is not mined in the main chain (which can occur when a
+  reorg un-mines a previously scanned transaction) were reported at every
+  `minconf`, and `maxconf` was never applied to transparent outputs at all. An
+  output of an unmined transaction is now reported only for `minconf = 0`, and
+  both bounds remain inclusive.
 - Async RPC operations whose task panics are now marked as failed, with the
   panic message reported via `z_getoperationstatus` and `z_getoperationresult`.
   Previously such operations remained in the `executing` state forever and
--- a/zallet-core/src/components/json_rpc/methods/list_unspent.rs
+++ b/zallet-core/src/components/json_rpc/methods/list_unspent.rs
@@ -20,7 +20,7 @@ use zcash_client_backend::{
     wallet::NoteId,
 };
 use zcash_keys::address::Address;
-use zcash_protocol::{ShieldedPool, value::Zatoshis};
+use zcash_protocol::{ShieldedPool, consensus::BlockHeight, value::Zatoshis};
 use zip32::Scope;
 
 use crate::components::{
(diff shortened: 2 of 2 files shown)


nullcopy (approved):
utACK e0973624dd5107796c1773d42381bbe16a9db9e6

PR #684, opened 2026-07-27
build: raise MSRV to Rust 1.95

gustavovalverde:
## Summary

- Raise the root and backend workspace MSRV to a coordinated Rust 1.95 compiler floor.
- Update the pinned rustup, Docker, StageX, and Nix toolchains together, plus the developer documentation and release note.
- Use the upstream post-0.24 `librocksdb-sys` standard-library selection fix so the StageX builder can select libc++ while the Nix builder retains its native libstdc++ toolchain.

## Compatibility

Zallet now requires Rust 1.95 for source builds. The previous Rust 1.88 floor was introduced during the NU6.3 update because `zewif-zcashd` adopted let-chain syntax stabilized in Rust 1.88; it was a real minimum at the time, not an upper-bound compatibility constraint. The current dependency graph does not require a higher compiler, and Rust 1.96 adds no capability needed by this change.

Builders using the repository's pinned Docker, StageX, or Nix inputs receive the matching compiler automatically. The RocksDB source override can return to crates.io after a release containing the upstream C++ standard-library selection fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,7 +74,7 @@ Co-Authored-By: Claude <noreply@anthropic.com>
 Zallet is a Zcash full node wallet, designed to replace the legacy wallet that was included within zcashd.
 
 - **Rust edition**: 2024
-- **MSRV**: 1.88 (pinned in `rust-toolchain.toml`)
+- **MSRV**: 1.95 (pinned in `rust-toolchain.toml`)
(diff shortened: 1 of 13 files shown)


schell (approved):

schell:
Thanks @gustavovalverde 🙇

PR #711, opened 2026-07-30
fix(start): own admitted runtime tasks

### LINKED ISSUE #136: sync: If one of the other tasks fails, `steady_state` loops instead of aborting

### LINKED ISSUE #702: zallet-core: Own admitted runtime tasks through startup
## Problem

`zallet start` can open or migrate the wallet database before a configured chain backend has completed admission. Tasks created during backend, RPC, or wallet-sync construction can also detach when a later startup step fails, the supervisor is cancelled, or a supervised task returns an error.

## Intended outcome

Treat successful chain construction as the admission boundary, keep every runtime task under abort-on-drop ownership until supervision takes over, and preserve the first task failure as the process result.

## Acceptance criteria

- Chain construction and consensus validation complete before the wallet database is opened or migrated.
- Backend rejection does not create or alter a wallet database.
- Later database, keystore, RPC, sync, cancellation, and supervision failures cancel every task already started.
- A supervised task error is returned instead of being converted to a successful process exit.
- Focused tests cover rejection, partial construction, cancellation, and failure propagation.

gustavovalverde:
Closes #702. Also closes #136 (see the final commit).

Backend admission and partial startup failures could mutate the wallet database or detach runtime tasks before supervision took ownership. This change makes chain construction the admission boundary and retains abort-on-drop ownership through startup and supervision.

## What changes

- Constructs and validates the selected chain backend before opening or migrating the wallet database.
- Keeps chain, RPC, and sync task ownership across startup failures and cancellation.
- Propagates the first supervised task failure instead of leaving sibling tasks detached.

## #136 follow-up (final commit)

The task-ownership rework above ensures `abort()` reaches `steady_state`, but the loop's `None if tip_changed` arm returned without an intervening await. Against a `ChainView` whose `snapshot`, `tip`, and `get_mempool_stream` all return `Poll::Ready` from cached state (the in-tree `MockChain`, and the Zaino `FetchServiceSubscriber` as reported in #136), the task could complete a full iteration without polling its abort status and spin indefinitely. The final commit inserts `tokio::task::yield_now().await` in that arm and adds a regression test.

### DIFF
--- a/zallet-core/src/commands/start.rs
+++ b/zallet-core/src/commands/start.rs
@@ -57,8 +57,6 @@ async fn supervise_zallet_tasks(
     wallet_sync_data_requests_task_handle: TaskHandle,
 ) -> Result<(), Error> {
     // Retain abort-on-drop ownership while the supervisor itself is cancellable.
-    let _task_owner = task_owner;
-
     info!("Spawned Zallet tasks");
 
     // ongoing tasks.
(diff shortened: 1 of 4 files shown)


zallet-core/src/commands/start.rs
```diff
@@ -21,12 +22,121 @@ use crate::{
 #[cfg(zallet_build = "wallet")]
 use crate::components::keystore::KeyStore;
 
+/// Owns cancellation for every task spawned by `zallet start`.
+struct StartupTaskOwner {
```
pacu, line 26: Nit: `StartupTaskOwner` and `PendingWalletSyncTasks` (in `sync.rs`) are structurally identical abort-on-drop guards — same fields, same `new`/`include`/`Drop`, differing only in that one has `transfer_to_caller`. A single shared guard type in `components` (with the disarm method) would avoid the duplication and keep the two call sites from drifting apart. Not blocking.

zallet-core/src/components/sync.rs
```diff
@@ -706,7 +747,15 @@ async fn steady_state_iteration<C: Chain>(
...
-        None if tip_changed => (),
+        //
+        // Yield to the runtime before re-iterating. `snapshot`, `tip`, and
+        // `get_mempool_stream` can all return `Poll::Ready` from cached state
+        // (MockChain does, and the Zaino `FetchServiceSubscriber` was observed
+        // doing so in #136), so without this yield an aborted `steady_state`
+        // task can complete a full iteration without ever polling its abort
+        // status, spinning until the backend's view changes. Yielding lets
+        // tokio observe the abort and end the task at this await point.
+        None if tip_changed => tokio::task::yield_now().await,
```
pacu: The `yield_now` is the correct fix for abortability, and the comment explains why the coop budget can't help here. One follow-up worth an issue: a *non-aborted* task in this arm still busy-spins at full CPU (yielding but immediately rescheduled) for as long as the backend keeps serving a stale cached view where `tip_changed` stays true and `stream_blocks_to_tip` yields nothing. That's arguably a backend contract violation, but given the Zaino cached-state behavior observed in #136, it might be worth a short bounded sleep or a tracked follow-up rather than relying on the backend to always make progress.

pacu (commented):
The substance here is good and I'd like to see it land — all four fixes are real, the abort-on-drop ownership handoffs are genuinely await-free where the comments claim they are, and the #136 regression test targets exactly the right failure mode (a `ChainView` serving cached state never touches an instrumented tokio resource, so the coop budget can't force a yield in that arm).

However, there's a structural problem with the base branch:

**This PR targets `build/pin-wallet-rc4-cohort`, which is now superseded by main.** Main adopted the rc.6 cohort in c652b6d using the same `=` exact-pin style, so merging the rc.4 pin branch would downgrade the pins, and it conflicts with main in all three `Cargo.toml`s plus `CHANGELOG.md`. I'd suggest closing `build/pin-wallet-rc4-cohort` as superseded and retargeting this PR to main.

**Retarget catch:** `start.rs`/`sync.rs` auto-merge textually onto main, but the result won't compile — the global sync lock (c56add3) added a `SyncStatusWriter` parameter to `WalletSync::spawn` and `steady_state`, and the new regression tests here call both with the old arity. The `status` handle needs threading through the new tests after rebase. The good news: main's `WalletSync::spawn` introduced no new awaits between the task spawns, so the `PendingWalletSyncTasks` insertion points and the ownership invariant hold unchanged after rebase.

One behavioral tradeoff worth stating explicitly for the record: wallet-DB configuration errors now surface only after chain-backend construction, which for the Zaino backend can be slow. That's the intended consequence of making backend construction the admission boundary (#702), just worth being aware of operationally.

A few smaller notes inline.

zallet-core/src/components/sync.rs
```diff
@@ -1042,6 +1136,176 @@ mod tests {
...
+            database,
+            MockChain::reporting(Vec::new(), 0),
+            None,
+            decryptor,
+            decryptor_engine,
+        )
+        .await;
+
+        assert!(result.is_err(), "mock chain rejects wallet sync startup");
+        if let Some(reload_finished) = decryptor_observer.reload_keys().await {
```
pacu, line 1275: This assertion is conditional: if `reload_keys()` returns `None`, the test silently passes without verifying anything about the decryptor. Consider asserting on the `None` case too (or documenting why `None` also implies the engine is down), so a regression that makes `reload_keys` return `None` early doesn't turn this test into a no-op.

schell:
Good catches @pacu. I didn't even notice that it wasn't targeting `main`. The issues should be resolved now.

pacu (dismissed):
Thanks for the changes! I think there's one nit that we could actually fix or punt for another PR 
 StartupTaskOwner and PendingWalletSyncTasks. Also we should close #720 since it has been superseeded by this PR

pacu (approved):

Issue #690, opened 2026-07-28
Surface per-item mempool failures from existing backends


gustavovalverde:
## Problem

The shared `ChainView` mempool stream can represent per-item `ChainError` values, but the existing Zaino and Zebra implementations retain their prior behavior of logging or dropping source, fetch, hex, and transaction-decoding failures. A later clean stream completion can therefore remain indistinguishable from a captured-view tip change.

## Intended outcome

Use the shared fallible item contract to surface recoverable mempool failures without changing the meaning of `None` or clean stream completion.

## Acceptance criteria

- Source, fetch, hex, and transaction-decoding failures are classified and yielded as `ChainError` items where the stream has already been acquired.
- The outer result remains limited to stream acquisition failure.
- `None` means that the captured view is already stale.
- Clean stream completion means that the captured view changed.
- Focused tests cover each backend error path and clean tip-change completion.
- No transport, dependency, retry-policy, or process-architecture redesign is introduced.

PR #713, opened 2026-07-30
fix(sync): propagate mempool stream failures

### LINKED ISSUE #709: zallet-core: Make mempool stream items fallible
## Problem

`ChainView::mempool_stream` cannot distinguish an item-level source or decoding failure from normal stream completion, which wallet sync interprets as a chain-tip change.

## Intended outcome

Separate stream acquisition failure from fallible transaction items so wallet sync can stop on an item error while preserving the existing meanings of `ChainError::ViewExpired` and clean completion.

## Acceptance criteria

- The outer result reports stream acquisition failure.
- Acquired streams yield `Result<RawTransaction, ChainError>` items.
- Wallet sync propagates the first item error.
- `ChainError::ViewExpired` continues to mean that the captured view is already stale.
- Clean stream completion continues to signal a later tip change.
- Zebra and Zaino preserve their current error-classification behavior; surfacing their currently dropped item errors remains tracked by #690.

gustavovalverde:
Closes #709.

A mempool stream could report acquisition failure, but it could not distinguish an item error from clean completion after acquisition. Wallet sync therefore risked interpreting a source or decoding failure as an ordinary chain-tip transition.

## What changes

- Makes acquired mempool streams yield fallible transaction items.
- Propagates the first item error through the sync task.
- Preserves the existing meanings of fixed-view expiry and clean stream completion.
- Updates Zebra and Zaino without broadening their current error classification; #690 tracks surfacing backend errors they still discard before the stream boundary.

### DIFF
--- a/backends/zaino/src/chain.rs
+++ b/backends/zaino/src/chain.rs
@@ -577,7 +577,9 @@ impl ChainView for ZainoChainView {
         self.stream_blocks_inner(range.start, Some(range.end - 1))
     }
 
-    async fn get_mempool_stream(&self) -> Result<Option<BoxStream<'_, Transaction>>, ChainError> {
+    async fn get_mempool_stream(
+        &self,
+    ) -> Result<Option<BoxStream<'_, Result<Transaction, ChainError>>>, ChainError> {
         let mempool_height = self.tip().await?.height() + 1;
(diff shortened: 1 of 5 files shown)


nuttycom (dismissed):
utACK 24c0fca392f1a7ef5777cd3fa1fc5de9285d46de

PR #714, opened 2026-07-30
fix(sync): serialize wallet reconfiguration

gustavovalverde:
Closes #707.

Account and key mutations could race active block-scan commits, while hot imports and rescans near the current tip could remain idle until another block or restart. This change gives wallet reconfiguration one serialized admission path and wakes every recovery owner after commit.

## What changes

- Serializes shielded wallet reconfiguration with block scanning.
- Retains admission until committed key changes are reloaded.
- Supports atomic rescans for existing viewing keys.
- Wakes historical recovery and near-tip scanning after account creation, recovery, import, or rescan.

PR #712, opened 2026-07-30
fix(sync): reacquire expired chain views

gustavovalverde:
Closes #706.

Wallet sync previously treated general source unavailability as fixed-view expiry and could begin wallet work before a complete range had been proven coherent. This change gives view invalidation an exact error category and makes range acquisition atomic from the wallet's perspective.

## What changes

- Adds `ChainError::ViewExpired` as a distinct backend contract.
- Buffers the complete block range before decryption or database mutation.
- Reacquires and retries only after concrete fixed-view expiry.
- Maps pinned-view loss precisely in the Zebra and Zaino backends.

PR #718, opened 2026-07-30
fix(sync): preserve transparent history watermarks

### LINKED ISSUE #704: zallet-core: Preserve transparent-history watermarks
## Problem

Address-based transparent-history recovery can receive a transaction identifier that the same fixed chain view cannot resolve. Treating that absence as a skippable source failure allows the address watermark to advance past incomplete spend history.

## Intended outcome

Fail closed on a history transaction missing from the selected view, and leave the request watermark unchanged so a later retry cannot silently omit the spend.

## Acceptance criteria

- Every transaction identifier returned by address history must resolve in the same chain view.
- A missing transaction is classified as invalid source data.
- The transaction-data worker stops rather than logging and continuing.
- The address scan watermark remains unchanged after failure.
- Focused tests cover the pure classification and persisted request state.

gustavovalverde:
Closes #704.

Address-based transparent-history recovery could advance its watermark after the selected fixed view failed to resolve a transaction identifier it had returned. That allowed later retries to skip incomplete spend history.

## What changes

- Requires every history transaction identifier to resolve in the same chain view.
- Classifies a missing transaction as invalid source data and stops the worker.
- Leaves the persisted address watermark unchanged on failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,8 @@ be considered breaking changes.
 
 ### Fixed
 
+- Address-based transparent-history recovery now fails before advancing its
+  scan watermark when the selected chain view omits a referenced transaction.
 - A note commitment tree conflict during sync no longer shuts the wallet down
   permanently. Zallet now rolls the wallet back to a progressively older point
   and rescans, up to a bounded number of attempts and never below the wallet's
--- a/zallet-core/src/components/chain.rs
+++ b/zallet-core/src/components/chain.rs
@@ -873,6 +873,8 @@ mod tests {
     #[derive(Clone)]
     pub(crate) struct MockChainView {
         tip: ChainBlock,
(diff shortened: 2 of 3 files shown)


pacu:
Scheduling context, and a rebase request.

Zallet is now on a **weekly release train** — a cut every Tuesday, shipping whatever is merged
and green (#781). This PR is targeted at the **beta.5 cut, Tue 2026-09-08**, owned by @schell.

It currently **conflicts with `main`** — could you rebase? Unlike #712 and #714 this one is not
part of a stack, so it can go in independently once it is mergeable; it is grouped into the
same week only because it touches `sync.rs`, and doing all the sync work in one week means one
rebase wave rather than several.

pacu:
This conflicts with `main` — could you rebase? It can't be reviewed until it's mergeable.

PR #685, opened 2026-07-27
Update to librustzcash 0.30/0.10 cohort and zaino 0.5 backend

schell:
## What

Moves Zallet onto the librustzcash 0.30/0.10 dependency cohort
(`zcash_primitives 0.30.0`, `zcash_transparent 0.10.0`, `zcash_keys 0.16.0`,
`zcash_client_backend 0.24.0-rc.4`, `zcash_client_sqlite 0.22.0-rc.4`,
`orchard 0.15.3`, ...), and updates the **zaino backend** to zaino 0.5 in
lockstep.

## zaino backend (the main change in this PR)

`backends/zaino` previously pinned the zaino crates to `zingolabs/zaino` `dev`.
That pin carries the older librustzcash 0.29/0.9 cohort, so this PR repoints it
to the zodl-inc/zaino branch
[`update/librustzcash-0.24.0-rc.4`](https://github.com/zingolabs/zaino/pull/1428)
(**zingolabs/zaino#1428**), which bumps zaino to the same 0.30/0.10 versions and
carries the Ironwood (NU6.3) note-commitment treestate required for spendable
Ironwood notes. The zaino crates remain a git pin because zaino 0.5 is not yet
released on crates.io.

The zaino `0.3 → 0.5` upgrade brings zaino's ADR-0003 network redesign:

- `ChainIndexConfig::new` is now a struct-literal config whose `network` is a
  `zebra_chain::parameters::Network`; the old payload-carrying
  `zaino_common::Network` is gone. Both the indexer config and the `State`
  validator connector now take the runtime network from the existing
  `network_to_zebra` helper, so the `network_to_zaino` converter is removed.
- `get_spendable_transparent_outputs` (zcash_client_backend 0.24.0-rc.4) gained
  a `lock_filter` argument; the sync spend-detection path passes
  `LockFilter::Unfiltered` so that locked outputs are not mistaken for spent.

## Zebra dependency

The zebra crates are consumed from **crates.io** at the first librustzcash
0.30/0.10 releases, which Zebra has now published (`zebra-chain 11.3`,
`zebra-state 12.0.1`, `zebra-rpc 15.0`). Earlier published versions
(11.2 / 12.0 / 14.0) are still the older librustzcash 0.29/0.9 builds. Using the
released crates unifies on a single copy of each zebra crate and removes the
earlier dependency on a draft Zebra PR.

## Verification

- `cargo check --all-targets` — clean for root, `backends/zebra`, and
  `backends/zaino`.
[...]

PR #720, opened 2026-07-30
build(deps): pin incompatible wallet API release candidates

### LINKED ISSUE #710: workspace: Pin the wallet release candidate cohort exactly
## Problem

The root, Zebra, and Zaino workspaces declare prerelease wallet dependencies with non-exact requirements. A fresh manifest resolution can therefore select a later release candidate with a different API even while checked-in lockfiles and certified builds use RC4.

## Intended outcome

Require the audited RC4 wallet cohort exactly in every independent workspace until the repository deliberately upgrades the cohort together.

## Acceptance criteria

- Root, Zebra, and Zaino manifests use the same exact `zcash_client_backend = "=0.24.0-rc.4"` requirement.
- The root manifest uses the matching exact `zcash_client_sqlite = "=0.22.0-rc.4"` requirement.
- Existing lockfiles remain semantically unchanged.
- `utils/check-lockstep.sh` passes across all three workspaces.
- A later cohort upgrade remains one explicit, audited change.

gustavovalverde:
Closes #710.

#685 replaced the shared librustzcash git revision with the published `zcash_client_backend 0.24.0-rc.4` and `zcash_client_sqlite 0.22.0-rc.4` crates. That change adapted and certified Zallet against the RC4 APIs, but the manifests used non-exact prerelease requirements.

Cargo permits those requirements to resolve to later release candidates. The lockfile-free latest-dependencies job consequently selected RC5 after it was published, where incompatible wallet API changes caused `zallet-core` to fail compilation. RC6 is now selected by a fresh resolution, while the checked-in lockfiles remain on the audited RC4 versions.

## What changes

- Requires `zcash_client_backend = "=0.24.0-rc.4"` in the root, Zebra, and Zaino workspaces, which each have an independent dependency graph.
- Requires the shared root `zcash_client_sqlite = "=0.22.0-rc.4"` dependency exactly.
- Leaves the checked-in lockfiles and stable wallet dependencies unchanged.

This is a narrow guard for the two prerelease wallet API roots introduced by #685. Stable wallet dependencies remain semver-flexible so the latest-dependencies job continues exercising compatible updates, while the existing lockstep check keeps wallet-critical versions aligned across all three workspaces.


schell:
Subsumed by #711

PR #721, opened 2026-07-30
fix(db): implement OutputLockStore and the new InputSource methods for DbConnection (rc.6 cohort)

pacu:
## Summary

`zcash_client_backend` `0.24.0-rc.5`/`rc.6` and `zcash_client_sqlite` `0.22.0-rc.5`/`rc.6` (published 2026-07-29/30) reshaped the data-access traits in two ways that break `zallet-core`'s `DbConnection`, which hand-implements these traits directly rather than only through `zcash_client_sqlite::WalletDb` (which received the corresponding impls upstream):

1. **rc.5** extracted `lock_outputs`/`unlock_output`/`clear_locked_outputs`/`get_locked_outputs` off `WalletWrite` into a new `OutputLockStore` supertrait ([`zcash_client_backend`'s `3b388cda12`](https://github.com/zcash/librustzcash/commit/3b388cda127dd5901ce85da90ea932a8e69f718f)), so `DbConnection` stopped satisfying `WalletWrite`'s new supertrait bound.
2. **rc.6** added a required `InputSource::anchor_computable` method with **no default implementation** (used by `propose_transfer` to abandon a ZIP 318 canonical crossing when the boundary anchor cannot be materialized, instead of proposing a transaction whose build would fail with `AnchorNotFound`).

The rc.5 break surfaced first in the `build-latest` CI canary (the job that deletes `Cargo.lock` to catch exactly this kind of floating-dependency break) on the push for #686 — a CI-only zizmor lint fix that just happened to be the first push after rc.5 hit crates.io.

## Changes

- Adds `impl OutputLockStore for DbConnection` in `zallet-core/src/components/database/connection.rs`, moving the three lock methods out of the `WalletWrite` impl and adding the new `get_locked_outputs`; disambiguates the now-ambiguous `Self::Error`/`Self::AccountId` references inside the `WalletWrite` impl, matching the `<Self as WalletRead>::...` convention used upstream.
- Adds `InputSource::anchor_computable` to `DbConnection`'s `InputSource` impl, delegating to `WalletDb` (backed by `zcash_client_sqlite`'s `wallet::anchor_computable` query).
- Also delegates the new `InputSource::select_single_spendable_note`. It has a best-effort trait default, but `WalletDb` overrides it with a direct oldest-covering-note query; without delegation `DbConnection` would silently degrade ZIP 318 canonical-crossing note selection to the default.
- Pins `zcash_client_backend`/`zcash_client_sqlite` to exactly `0.24.0-rc.6`/`0.22.0-rc.6` across the root manifest and both backend manifests, with surgical `cargo update --precise` updates of all three lockfiles rather than a full regeneration (`zcash_pool_migration` `0.1.0-rc.4` → `rc.5` is the one required transitive bump).
- Refreshes cargo-vet audit state so `cargo vet` passes in the root and both backends against the new cohort.

## Test plan

- [x] `cargo build --workspace --all-targets --features zcashd-import,rpc-cli,tokio-console`
- [x] `cargo build --manifest-path backends/zaino/Cargo.toml --all-targets --features zcashd-import,rpc-cli,tokio-console`
- [x] `cargo build --manifest-path backends/zebra/Cargo.toml --all-targets --features zcashd-import,rpc-cli,tokio-console`
- [x] `cargo clippy -p zallet-core --all-targets --features zcashd-import,rpc-cli,tokio-console`
- [x] `cargo fmt --check -p zallet-core`
- [x] `cargo test -p zallet-core --features zcashd-import,rpc-cli,tokio-console` (301 passed)
- [x] `cargo vet` in the root and both backends
- [x] Reproduced the `build-latest` canary (deleted all three lockfiles, fresh `cargo build` of the root workspace and both backends) against the rc.6 pins to confirm the floating-dependency break is resolved end-to-end

https://claude.ai/code/session_019Yrqzkq1tPkyyhmhJoT51U


schell (approved):
👍

Issue #255, opened 2025-09-10
Add transaction size estimation logic


str4d:
https://github.com/zcash/wallet/blob/2eb2d4b027195f664b4a0fe5ea3877402e00e37a/zallet/src/components/json_rpc/methods/z_send_many.rs#L177-L178

PR #722, opened 2026-07-30
Close remaining gaps in the hardening changes from #677

### LINKED ISSUE #662: generate-encryption-identity --passphrase accepts an empty passphrase on both input paths.
generate-encryption-identity --passphrase accepts an empty passphrase on both input paths. A present-but-empty ZALLET_IDENTITY_PASSPHRASE is returned as Ok("") with no validation (generate_encryption_identity.rs:215), and two empty interactive entries pass the only check, which is equality (:231); the value then flows to age::Encryptor::with_user_passphrase("") (:191), producing an ASCII-armored, mode-0600 identity file whose secret is the empty string, which KeyStore::new subsequently recognizes and unlocks with "" (keystore.rs:294). The file looks encrypted but its protection is zero-entropy and publicly guessable, so anyone who obtains the identity file together with wallet.db decrypts every mnemonic and spending key and spends offline, and the command reports success with no weak-passphrase warning. To fix this, reject empty (and whitespace-only) passphrases fail-closed on both paths through shared validation, and add regression tests for the empty-input cases.

pacu:
## Summary

Follow-ups to #677, targeting that PR's branch so they land together.

Each commit closes a gap in a hardening change #677 already makes: either the
change covers one location of a problem but not another, or it stops one step
short of the property it is meant to guarantee. One commit per fix, as in #677.

### `c06bb54` — constant-time DFVK comparison

The binding check for the decrypted standalone Sapling `extsk` used a
short-circuiting `!=`, while `seed_matches_fingerprint` in the same commit moved
to `subtle::ConstantTimeEq`. Both operands here derive from a secret spending
key, so use the same comparison.

### `9f55387` — reject a pwhash that authenticates the empty password

#677 rejects an empty **bare** `password`, but the `pwhash` arm still accepts
any parseable hash — including one generated from the empty string.
`PasswordHash::from_bare("")` looks like any other hash, so `add-rpc-user`
output generated before that check, or a hand-written entry, still authenticates
the secretless credential `<username>:`.

This matters most for `__cookie__`: the username is defined in the source, and a
configured entry using it suppresses generated-cookie creation, so the resulting
credential is both publicly derivable and the only one in play. Each configured
pwhash is now checked against the empty string at load time.

### `d94dde5` — owner-only identity parent directory

`generate-encryption-identity` created missing parent directories at the process
umask (commonly 0755). It does not take the data directory lock, so it cannot
rely on #677's `lock_datadir` chmod: a datadir first initialized by this command
stays world-readable until some later command runs, and a custom `-o` path
outside the datadir is never restricted at all.

The acceptance test asserts mode 0700 so it holds regardless of the umask CI
runs under.

### `d1abb2a` — protected parameter input for `zallet rpc`

`zallet rpc` required every JSON-RPC parameter to be a command-line argument, so
`z_importkey` put a Sapling spending key in the process argument vector —
readable through process listings, and recorded in shell history. #677 removed
the `ZALLET_IDENTITY_PASSPHRASE` env-var channel; this is the argv side.

A parameter written `@PATH` is now read from the first line of `PATH` and sent
as a JSON string:

```
$ get-key-from-vault | zallet rpc z_importkey @- '"whenkeyisnew"'
$ zallet rpc z_importkey @/dev/fd/3 3<<<"$KEY"
$ zallet rpc z_importkey @-      # prompts without echo on a terminal
```

`@` cannot collide with an ordinary parameter, as no JSON value begins with it.
The read buffer is `Zeroizing`. This also fills in the `zcash-cli -stdin` row of
the comparison table in the book, previously "Not implemented".

### `dc4d23a` — bound the transaction before the expensive work

Two gaps remained after #677 made the action limit per-pool:

1. `z_sendmany` only bounded the transaction *after* proposing it, so an
   unreasonable `amounts` array still drove input selection first. Every
   recipient needs at least one output in its pool, so a request with more
   recipients than the limit cannot produce a passing proposal — reject it up
   front. The serialized-size bound (which also depends on memo lengths and
   selected inputs) remains #255.
2. `z_shieldcoinbase` had **no** action limit at all, though it is the method
   most likely to produce an oversized transaction: without `limit` it selects
   every eligible coinbase UTXO, as its own docstring warns. #677's extracted
   `check_shielded_action_limits` made this a few lines.

### `928c756` — CHANGELOG

Also drops a duplicated `zallet rpc help` entry introduced earlier on this
branch.

## Not included: exact `cargo-deb` pin

One more commit is ready but **not pushed**: the deb job installs
`cargo-deb --version "^2.10"` while the GPG signing key is already on the
runner, so a semver-compatible release from a compromised upstream would execute
next to the key without review. `--locked` pins cargo-deb's dependencies, not
cargo-deb itself. The fix pins `=2.12.1`.

My token lacks the `workflow` scope needed to push `.github/workflows/`. Happy
to add it here once that is sorted, or someone with the scope can apply the
one-line change directly.

## Testing

- `cargo test -p zallet-core --lib --all-features` — 295 passed, 0 failed.
- `cargo fmt` clean on all three manifests (workspace, zebra backend, zaino
  backend), and `cargo clippy` clean.
- Compiles under both `zallet_build = "wallet"` and
  `zallet_build = "merchant_terminal"`.
- New tests: `pwhash_of_empty_password_is_rejected`,
  `reads_first_line_without_terminator`, `missing_file_is_an_error`, plus a mode
  assertion in the existing `generate_encryption_identity_creates_missing_parent_dir`
  acceptance test.

Note: the `Latest build` CI job is failing on `main` as well — it deletes the
lockfiles and builds against newest dependencies, so it is unrelated to this
branch.

https://claude.ai/code/session_01L2qGvpzxEM1Vk1XNEz7Xg1

Closes https://github.com/zcash/zallet/issues/662

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/z_send_many.rs
+++ b/zallet-core/src/components/json_rpc/methods/z_send_many.rs
@@ -1,5 +1,7 @@
 use std::convert::Infallible;
 
+use abscissa_core::Application;
+
 use jsonrpsee::core::{JsonValue, RpcResult};
 use secrecy::ExposeSecret;
 use serde_json::json;
@@ -39,6 +41,7 @@ use crate::{
         keystore::KeyStore,
     },
     fl,
+    prelude::APP,
 };
 
 #[cfg(feature = "zcashd-import")]
@@ -126,8 +129,22 @@ pub(crate) async fn call<C: Chain>(
 
     let privacy_policy = parse_privacy_policy(privacy_policy.as_deref())?;
 
-    // Sanity check for transaction size
-    // TODO: https://github.com/zcash/zallet/issues/255
+    // Reject an oversized recipient set before input selection and proving, which both
+    // scale with it. Every recipient needs at least one output in its pool, so this only
+    // rejects requests the post-proposal per-pool check would reject anyway.
+    //
+    // TODO: also bound the serialized transaction size.
+    // https://github.com/zcash/zallet/issues/255
+    let actions_limit: usize = APP.config().builder.limits.orchard_actions().into();
+    if amounts.len() > actions_limit {
+        return Err(LegacyCode::Misc.with_message(fl!(
+            "err-excess-recipients",
+            count = amounts.len(),
+            limit = actions_limit,
+            config = "-orchardactionlimit=N",
+            bound = format!("N >= %u"),
+        )));
+    }
 
     let confirmations_policy = confirmations_policy_for_minconf(minconf)?;
 
--- a/zallet-core/src/components/json_rpc/server/authorization.rs
+++ b/zallet-core/src/components/json_rpc/server/authorization.rs
@@ -170,7 +170,12 @@ impl AuthorizationLayer {
                 }
                 (None, Some(pwhash)) => {
                     using_pwhash = true;
-                    Ok((a.user, pwhash.parse()?))
+                    let hash: PasswordHash = pwhash.parse()?;
+                    // Reject empty passwords.
+                    if hash.check("") {
+                        return Err(());
+                    }
+                    Ok((a.user, hash))
                 }
                 _ => Err(()),
             })
@@ -272,6 +277,27 @@ mod tests {
             pwhash: None,
         }];
         assert!(AuthorizationLayer::new(auth, None).is_err());
+    }
+
+    /// An empty password pre-hashed into a `pwhash` entry is just as unauthenticated as a
+    /// bare empty password, and must be rejected on the same grounds.
+    #[test]
+    fn pwhash_of_empty_password_is_rejected() {
+        use crate::config::RpcAuthSection;
+
+        let auth = vec![RpcAuthSection {
+            user: "__cookie__".to_string(),
+            password: None,
+            pwhash: Some(PasswordHash::from_bare("").to_string()),
+        }];
+        assert!(AuthorizationLayer::new(auth, None).is_err());
+
+        let auth = vec![RpcAuthSection {
+            user: "user1".to_string(),
+            password: None,
+            pwhash: Some(PasswordHash::from_bare("abadpassword").to_string()),
+        }];
+        assert!(AuthorizationLayer::new(auth, None).is_ok());
     }
 
     #[test]
--- a/zallet-core/src/components/keystore.rs
+++ b/zallet-core/src/components/keystore.rs
@@ -118,6 +118,7 @@ use std::time::{Duration, SystemTime};
 use bip0039::{English, Mnemonic};
 use rusqlite::named_params;
 use secrecy::{ExposeSecret, SecretString, SecretVec, Zeroize};
+use subtle::ConstantTimeEq;
 use tokio::{
     sync::{Mutex, RwLock},
     task::JoinHandle,
@@ -890,8 +891,13 @@ impl KeyStore {
                 let extsk = decrypt_standalone_sapling_extsk(&identities, &encrypted_extsk)?;
 
                 // The ciphertext is not bound to the DFVK the row is keyed by, so verify
-                // that the decrypted key reproduces the DFVK used for the lookup.
-                if extsk.to_diversifiable_full_viewing_key().to_bytes() != dfvk_array {
+                // that the decrypted key reproduces it.
+                if !bool::from(
+                    extsk
+                        .to_diversifiable_full_viewing_key()
+                        .to_bytes()
+                        .ct_eq(&dfvk_array),
+                ) {
                     return Err(ErrorKind::Generic
                         .context(fl!("err-keystore-key-material-mismatch"))
                         .into());
@@ -1113,7 +1119,6 @@ fn encrypt_string(
 ///
 /// [ZIP 32]: https://zips.z.cash/zip-0032#seed-fingerprints
 fn seed_matches_fingerprint(seed: &[u8], seed_fp: &SeedFingerprint) -> bool {
-    use subtle::ConstantTimeEq;
     SeedFingerprint::from_seed(seed)
         .is_some_and(|fp| bool::from(fp.to_bytes().ct_eq(&seed_fp.to_bytes())))
 }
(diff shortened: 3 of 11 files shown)


zallet-core/src/components/keystore.rs
```diff
@@ -890,8 +890,16 @@ impl KeyStore {
                 let extsk = decrypt_standalone_sapling_extsk(&identities, &encrypted_extsk)?;
 
                 // The ciphertext is not bound to the DFVK the row is keyed by, so verify
-                // that the decrypted key reproduces the DFVK used for the lookup.
-                if extsk.to_diversifiable_full_viewing_key().to_bytes() != dfvk_array {
+                // that the decrypted key reproduces the DFVK used for the lookup. The
+                // comparison is constant-time, as `seed_matches_fingerprint` is: the
+                // operand is derived from a secret spending key.
+                use subtle::ConstantTimeEq;
```
nullcopy: Please move imports to the top of the file

zallet-core/src/components/json_rpc/server/authorization.rs
```diff
@@ -105,6 +105,14 @@ impl PasswordHash {
     fn check(&self, password: &str) -> bool {
         hash_password(password, &self.salt) == self.hash
     }
+
+    /// Whether this hash authenticates the empty password.
+    ///
+    /// A configured `pwhash` is opaque, so an entry generated from an empty password is
+    /// indistinguishable from a real one until it is checked against `""`.
+    fn accepts_empty_password(&self) -> bool {
```
nullcopy: This helper doesn't feel useful. Remove it and inline the check.

zallet-core/src/components/json_rpc/server/authorization.rs
```diff
@@ -170,7 +178,15 @@ impl AuthorizationLayer {
                 }
                 (None, Some(pwhash)) => {
                     using_pwhash = true;
-                    Ok((a.user, pwhash.parse()?))
+                    let hash: PasswordHash = pwhash.parse()?;
+                    // A pwhash generated from an empty password authenticates the
+                    // credential `<username>:`, which carries no secret. Reject it here
+                    // as well as in the bare-password arm, so an empty password cannot
+                    // be smuggled past that check by pre-hashing it.
```
nullcopy: This entire comment should be reduced to just: "reject empty passwords"

zallet-core/src/components/json_rpc/methods/z_send_many.rs
```diff
@@ -126,8 +127,30 @@ pub(crate) async fn call<C: Chain>(
 
     let privacy_policy = parse_privacy_policy(privacy_policy.as_deref())?;
 
-    // Sanity check for transaction size
-    // TODO: https://github.com/zcash/zallet/issues/255
+    // Sanity check for transaction size, before any work proportional to it.
```
nullcopy: Can this comment be made more concise?

nullcopy (changes requested):
See comments
NACK cfa1b06243d2b4b8d340ced061eeae260b9a8069

zallet-core/src/components/json_rpc/server/authorization.rs
```diff
@@ -170,7 +170,12 @@
                 }
                 (None, Some(pwhash)) => {
                     using_pwhash = true;
-                    Ok((a.user, pwhash.parse()?))
+                    let hash: PasswordHash = pwhash.parse()?;
+                    // Reject empty passwords.
+                    if hash.check("") {
```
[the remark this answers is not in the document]
nullcopy: Its just testing the empty case

zallet-core/src/components/keystore.rs
pacu: Done — hoisted to a top-level `use subtle::ConstantTimeEq;`. I also removed the inline one in `seed_matches_fingerprint` (from the earlier commit on this branch), so the file now has a single import rather than a top-level one plus a leftover inline. Shortened the comment at the call site too.

zallet-core/src/components/json_rpc/server/authorization.rs
pacu: Done — helper removed, now just `if hash.check("")` inline.

zallet-core/src/components/json_rpc/server/authorization.rs
pacu: Done — reduced to `// Reject empty passwords.`

zallet-core/src/components/json_rpc/methods/z_send_many.rs
pacu: Done — cut from 13 lines to 3 plus the TODO. Note this hunk also moved: the base branch's `propose_and_check` refactor now runs the post-proposal check for both `z_sendmany` and `pczt_create`, so I dropped my duplicate of it here and kept only the pre-proposal recipient bound.

pacu:
Addressed all four comments and rebased onto the current branch head (`5fb80e4`).

**Review fixes**
- `keystore.rs`: hoisted `use subtle::ConstantTimeEq;` to the top. Also removed the inline one in `seed_matches_fingerprint` so there is a single import, not a top-level plus a leftover inline.
- `authorization.rs`: dropped the `accepts_empty_password` helper, inlined as `hash.check("")`; comment reduced to `// Reject empty passwords.`
- `z_send_many.rs`: comment cut from 13 lines to 3 plus the TODO.

**Rebase note worth a look:** the `propose_and_check` refactor now runs `enforce_privacy_policy` and `check_shielded_action_limits` for both `z_sendmany` and `pczt_create`, so my duplicate post-proposal check in `z_send_many.rs` is gone — only the pre-proposal recipient bound remains there. `z_shieldcoinbase` does not go through `propose_and_check`, so it keeps its own call.

On the CodeQL hard-coded-password alert: agreed it is just the empty case being tested. Left as is — happy to dismiss it or add a suppression if you would rather it not sit in the alert list.

Verified: 312 tests pass, `cargo fmt` clean on all three manifests, clippy clean. The `Latest build` job is failing on `main` too (it drops the lockfiles and builds against newest deps), so it is not from this branch.

nullcopy (approved):
utACK 8f8abcc6f8e6bd730420313c524ff4ee67d41647

PR #725, opened 2026-07-31
fix(rpc): bound the async operation queue without discarding results

nullcopy:
Takes over #465 (the original author appears to be inactive; thanks @ouicate for the report and initial fix). This is a reimplementation that addresses @nuttycom's review feedback there.

## Summary

Fixes GHSA-3wr9-v982-59fj: `WalletRpcImpl::async_ops` was an unbounded `Vec<AsyncOperation>`, so an authenticated JSON-RPC caller could exhaust wallet memory by calling `z_sendmany`/`z_shieldcoinbase` repeatedly without ever collecting results.

The queue is now a dedicated `OperationQueue` type in `asyncop.rs` that owns the capacity and retention invariants:

- **Retention contract preserved:** finished operations are retained until explicitly collected with `z_getoperationresult`, matching `zcashd`'s `asyncrpcqueue` semantics. Uncollected txids are never silently discarded below the limit.
- **Bounded memory:** inserting into a full queue evicts only the *oldest finished* operations, one-in-one-out. If the queue is entirely unfinished operations, the insertion is rejected.
- **Configurable limit:** `rpc.async_operation_limit` (default 1000), joining the existing configurable anti-memory-exhaustion surface. The example config snapshot is regenerated.
- **Cheap rejection:** `z_sendmany` and `z_shieldcoinbase` call `check_capacity()` *before* running `call()`, so a rejected flood request does not pay for input selection or a keystore decryption. The check is advisory (small TOCTOU window); `insert` re-enforces the bound authoritatively.
- **Lock discipline:** eviction candidates are computed from a snapshot taken under a read lock, with per-operation state reads done while no queue lock is held. The write-lock section is synchronous (`retain` + push), so status/result/list RPCs are not blocked across awaits.
- **Distinct error code:** queue-full returns `LegacyCode::OutOfMemory` (-7) rather than `Misc` (-1), so clients can distinguish the transient, retriable condition. Happy to change if there's a better precedent.

Unit tests cover the retention contract (the regression #465 would have introduced), oldest-finished-first eviction, and rejection/recovery when the queue is full of unfinished operations.

## Review feedback from #465 → how it's addressed

1. *Result loss* → finished ops are only evicted at capacity, oldest first; never merely because a new RPC started.
2. *Liveness wedge* → rejection now requires the entire (larger, configurable) queue to be unfinished, and capacity is reclaimed as soon as any operation finishes. A true wedge still requires `limit` permanently-stuck futures; an execution timeout or cancellation RPC is left as a follow-up (see below).
3. *Arbitrary constant* → hardcoded 64 replaced by `rpc.async_operation_limit` with an honest rationale; no claimed `zcashd` lineage.
4. *Unbounded work on rejection* → capacity pre-check before proposal construction.
5. *Lock held across awaits* → see lock discipline above.

## Remaining follow-ups (out of scope here)

- In-flight operations have no execution timeout and `OperationState::Cancelled` is still unreachable (pre-existing). Worth its own issue/design discussion, since aborting an in-flight send interacts badly with broadcast.
- Worst-case retained memory is `limit × max params size`. Dropping `ContextInfo.params` from finished operations (keeping the result) would shrink this substantially, at the cost of changing `z_getoperationstatus`/`z_getoperationresult` output for finished ops.

https://claude.ai/code/session_01Fkm7VcvwCkGCy2y8AagzGZ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,17 @@ be considered breaking changes.
 - `getwalletstatus` now reports a `locked` field indicating whether the wallet
   is currently usable.
 
+### Fixed
+
+- The queue of asynchronous RPC operations (`z_sendmany`, `z_shieldcoinbase`)
+  is now bounded (GHSA-3wr9-v982-59fj). Finished operations are still retained
+  until collected with `z_getoperationresult`, but once the queue reaches its
+  limit (configurable via `rpc.async_operation_limit`, default 1000), the
+  oldest finished operations are evicted to make room, and new operations are
+  rejected while the queue is full of unfinished operations. Previously the
+  queue was unbounded, allowing an authenticated caller to exhaust wallet
+  memory by starting operations without collecting their results.
+
 ## [0.1.0-beta.2] - 2026-07-28
 
 ### Added
--- a/backends/zebra/tests/cmd/example_config.out/zallet.toml
+++ b/backends/zebra/tests/cmd/example_config.out/zallet.toml
@@ -337,6 +337,15 @@ as_of_version = "0.1.0-beta.2"
 #
 [rpc]
 
+# The maximum number of asynchronous RPC operations (e.g. `z_sendmany`) that can
+# be queued at once.
+#
+# Finished operations are retained in the queue until their results are collected
+# with `z_getoperationresult`, or until they are evicted (oldest first) to make
+# room for a new operation. New operations are rejected while the queue is full
+# of unfinished operations.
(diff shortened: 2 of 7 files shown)


nullcopy (commented):
Self-ack e26a7c41e8d958b818a5b8c3abeee639cc2bfc3a

nullcopy:
Moved to PR from local branch, so that integration tests can run: https://github.com/zcash/zallet/pull/726

PR #726, opened 2026-08-03
fix(rpc): bound the async operation queue without discarding results

nullcopy:
https://github.com/zcash/zallet/pull/725

Fixes GHSA-3wr9-v982-59fj: the queue of async operations backing z_sendmany and z_shieldcoinbase was unbounded, so an authenticated caller could exhaust wallet memory by starting operations without ever collecting their results.

Queue maintenance is now owned by a dedicated OperationQueue type. Finished operations are retained until collected via z_getoperationresult (matching zcashd), but inserting into a full queue evicts the oldest finished operations, and fails if the queue is full of unfinished operations. The bound is configurable via rpc.async_operation_limit (default 1000). Async RPC methods check capacity before constructing a proposal, so rejected requests are rejected cheaply, and eviction candidates are computed without holding the queue lock across per-operation state reads.

### DIFF
--- a/zallet-core/src/components/json_rpc/asyncop.rs
+++ b/zallet-core/src/components/json_rpc/asyncop.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::future::Future;
 use std::panic::AssertUnwindSafe;
 use std::sync::Arc;
@@ -12,7 +13,7 @@ use jsonrpsee::{
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use tokio::sync::RwLock;
+use tokio::sync::{RwLock, RwLockReadGuard};
 use uuid::Uuid;
 
 use super::server::LegacyCode;
@@ -73,6 +74,11 @@ impl OperationState {
             "success" => Some(Self::Success),
             _ => None,
         }
+    }
+
+    /// Returns whether an operation in this state has finished executing.
+    pub(super) fn is_finished(self) -> bool {
+        matches!(self, Self::Cancelled | Self::Failed | Self::Success)
     }
 }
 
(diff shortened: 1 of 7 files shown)


schell:
Hey @nullcopy - the claude code session requires an Anthropic account. I'm not sure that was your intent?

zallet-core/src/components/json_rpc/asyncop.rs
```diff
@@ -243,6 +249,158 @@ impl AsyncOperation {
...
+    pub(super) async fn check_capacity(&self) -> RpcResult<()> {
+        if self.ops.read().await.len() < self.limit || self.any_finished().await {
+            Ok(())
+        } else {
+            Err(Self::full_error())
+        }
+    }
+
+    /// Returns a read guard over the operations in the queue.
+    pub(super) async fn read(&self) -> RwLockReadGuard<'_, Vec<AsyncOperation>> {
```
schell, line 326: nit: In these cases I like to return `impl Deref<Vec<AsyncOperation>> + '_` just to keep the guard's type out of the picture.

zallet-core/src/components/json_rpc/asyncop.rs
```diff
@@ -243,6 +249,158 @@ impl AsyncOperation {
...
+    /// their statuses.
+    ///
+    /// The queue's write lock is held for the duration, so concurrent calls cannot
+    /// collect the same result twice. Insertions must go through [`Self::insert`] to
+    /// maintain the queue bound.
+    pub(super) async fn collect_finished(
+        &self,
+        filter: impl Fn(&AsyncOperation) -> bool,
+    ) -> Vec<OperationStatus> {
+        let mut ops = self.ops.write().await;
```
schell, line 340: It seems holding the lock would be good and fine so long as `op.state().await` and `op.to_status()` are "quick".

schell (approved):
utACK [b0d4f9c](https://github.com/zcash/zallet/pull/726/commits/b0d4f9c282d8af5b72e09048ebc508388b9234f7)

zallet-core/src/components/json_rpc/asyncop.rs
nullcopy: Hmm, I actually prefer to return the guard type, so that guard-specific methods (like `RwLockReadGuard::map`) are still accessible. That said, I don't think it makes much of a difference here, and I can come back and change it if you have a strong opinion.

PR #728, opened 2026-08-04
Require confirmed mnemonic backups before deriving new spend authority

### LINKED ISSUE #201: Consider how `walletrequirebackup` setting from `zcash.conf` should be used in Zallet
`zcash.conf` has a `walletrequirebackup` option, added as part of migrating `zcashd` wallets to mnemonic-based HD seeds (https://github.com/zcash/zcash/pull/5306). Due to never having had mnemonics before (and not having a central seed that could be backed up at Zcash launch), we requires `zcashd` operators to use `zcashd-wallet-tool` to confirm that they had correctly backed up the seed phrase.

Currently, we migrate this option over to Zallet, but don't do anything with it. We need to decide what to do with it.

Note that Zallet will be somewhat different from `zcashd` wallet: not only will new wallets be using seed phrases from the get-go, but:
- Wallets do not generate mnemonics without explicit user interaction (currently via `zallet generate-mnemonic`), so we have that opportunity to interact with them instead for encouraging or checking backups.
- Zallet supports multiple seed phrases.

sellout:
`zcash.conf`'s `walletrequirebackup` is migrated into `zallet.toml` as
`keystore.require_backup`, but Zallet did nothing with it beyond warning at
startup that it was unimplemented. This implements it.

The design departs from `zcashd` in one significant way, so the rationale leads
and the code follows.

## What `zcashd` did

`-walletrequirebackup` gated six RPC methods through one helper
(`src/wallet/rpcwallet.cpp:105`), which refused with
`RPC_WALLET_BACKUP_REQUIRED` (-18) until `CWallet::MnemonicVerified()` returned
true. That flag was set only by `walletconfirmbackup`
(`rpcwallet.cpp:2210`), which takes **the full recovery phrase** and compares it
against the wallet's own (`CWallet::VerifyMnemonicSeed`, `wallet.cpp:4271`)
before persisting the confirmation. It defaulted on everywhere except regtest
(`chainparams.cpp:941`, `fRequireWalletBackup = false`).

`zcashd-wallet-tool` drove that: it called `z_exportwallet` (which writes every
wallet secret to a file in plain text), displayed the phrase, cleared the
screen, quizzed the operator on up to three randomly chosen words, and piped the
phrase into `walletconfirmbackup`. The load-bearing part is the round trip — the
word quiz is client-side, and the wallet only ever sees the full phrase come
back.

## Why `confirm-backup` never displays the phrase

`zcashd-wallet-tool` printed the phrase to the terminal. Zallet does not, and
this is the deliberate departure.

Everything else in the keystore is built the other way — `export-mnemonic`
emits ciphertext and warns at length that it is not plaintext; `generate-mnemonic`
prints only the fingerprint. Displaying a phrase here would make this the one
place Zallet emits key material in the clear, and since Zallet is typically run
headless that means writing 24 words into an SSH session.

Instead, the operator obtains their own copy (`export-mnemonic`, then `age -d`)
and reads three words back. The quiz mechanism is unchanged; only the display is
gone. That also strengthens what confirmation attests to:

- Display-and-quiz would prove the operator saw the phrase on screen just now.
- Quizzing against their own decrypted export proves they hold the export *and*
  a working age identity *and* its passphrase.

The second is a superset, and the identity is the failure mode that actually
destroys a Zallet wallet. Neither proves the copy is durable or off-machine;
nothing can.

This is also why `export-mnemonic` gained `--seedfp`. Obtaining the phrase now
runs through it, and it was keyed on an account UUID — so a freshly generated
phrase, which by definition has no accounts derived from it yet, could not be
exported at all. Without that, the gate would have been a deadlock: no export,
so no confirmation, so no first account.

## Why account creation is gated, not just address derivation

`zcashd` gated both. It would be tempting to gate only address derivation, on
the theory that an account with no addresses is harmless — but that is false
here. `WalletWrite::create_account` materializes addresses before it returns
(`zcash_client_sqlite-0.22.0-rc.6/src/wallet.rs:655`): the default unified
address always, plus external transparent addresses up to the default index and
the full gap limit for the external, internal, and ephemeral scopes. Those are
readable via `listaddresses` and `z_listaccounts` without calling any
address-deriving method, so an account can receive funds the moment it exists.

Gating address derivation alone would leave all of that reachable. Gating
account creation is also the narrower surface: two RPC methods plus one regtest
command, versus a reveal surface that is larger and easier to grow by accident.

Of `zcashd`'s six gated methods Zallet implements two, and both are gated, plus
Zallet's own `z_recoveraccounts`. `getnewaddress`, `z_getnewaddress`,
`getrawchangeaddress`, and `keypoolrefill` do not exist here.

`z_recoveraccounts` looks like an odd fit — recovery implies the operator has
the phrase — but it takes a fingerprint, not a phrase, and the seed must already
be in the keystore, so the call itself proves nothing. Whether the operator
holds the phrase was settled when the seed was *stored*. The documented recovery
flow (`import-mnemonic` then `z_recoveraccounts`) is confirmed by construction
and never hits the gate; what the gate catches is `z_recoveraccounts` used for
what it also is — creating accounts at chosen ZIP 32 indices — on a seed nobody
has backed up.

## Which phrases need confirming

| Origin | Status | Reason |
|---|---|---|
| `import-mnemonic` | Confirmed | The operator typed it in, so they necessarily had it |
| `generate-mnemonic` | Unconfirmed | Exists nowhere but this wallet |
| `migrate-zcashd-wallet` | Unconfirmed | ZeWIF does not carry `zcashd`'s own backup flag, so we cannot tell an operator who wrote the phrase down from one who never saw it |
| Imported viewing key | n/a | No mnemonic covers it, so there is nothing to have backed up |

Confirmation is monotonic: storing a phrase again never withdraws it, and
re-importing a phrase Zallet generated *confirms* it — re-supplying it
demonstrates possession exactly as reading it back does.

The default is network-dependent, matching `zcashd`: off on regtest, where
wallets are disposable and confirmation is interactive (requiring it would leave
automated tests no way through), on everywhere else.

## Passphrase-encrypted identity files

`export-mnemonic` could not work at all on a wallet whose age identity file is
passphrase-encrypted. Exporting decrypts the stored phrase, which needs the age
identities; `KeyStore::new` leaves those unloaded when the identity file is
encrypted, and nothing asked for the passphrase, so the command failed with
“Wallet is locked”.

There was no way around that. Zallet has no unlock command; `walletpassphrase`
unlocks a running server's in-memory identities rather than a one-shot command's,
and the two cannot run concurrently because both take the data directory lock;
and no RPC method exports a mnemonic. Such a wallet could not produce a mnemonic
backup at all, short of pulling the ciphertext out of `wallet.db` by hand.

That predates this work, but the backup requirement puts `export-mnemonic` on the
critical path — it is how an operator obtains the phrase to confirm — so leaving
it would have meant those wallets could not derive accounts either, with
`keystore.require_backup = false` as their only escape. It is fixed here, in its
own commit ahead of the rest, by prompting on the terminal exactly as
`confirm-backup` does. The regtest account command reads the seed the same way
and is fixed with it.

## Database write atomicity

Every operation added here issues exactly **one** write, so per AGENTS.MD the
remaining three questions do not arise:

- `encrypt_and_store_mnemonic` — one `INSERT … ON CONFLICT DO UPDATE`.
- `confirm_backup` — one `UPDATE`.
- The migration — one `ALTER TABLE`, inside the migration framework's
  transaction.

No new multi-statement write, and no compensating logic.

## Upgrade impact

The wallet database gains a `backup_confirmed` column; existing databases are
migrated on open. **They carry no record of any prior confirmation**, so an
existing wallet whose phrase Zallet generated will refuse to derive further
accounts or addresses until `confirm-backup` is run against it, or
`keystore.require_backup` is set to `false`. Spending is unaffected, as is
receiving to addresses already derived.

That is the honest default: we cannot know whether an operator backed up a
phrase before this existed, and the safe assumption is the one that asks.

## Testing

18 new unit tests (305 → 323), covering the storage semantics (including the
upsert's confirm-on-reimport and never-unconfirm behaviour), the gate's three
branches including the `-18` code and the `require_backup = false` bypass, the
network-dependent default in both directions, seed selection, and the quiz's
word choice and answer matching. The migrated schema is pinned by the existing
`verify_schema` test.

Each commit independently passes `cargo fmt --check` and
`cargo clippy --all-targets -- -D warnings`, on the root workspace and both
backends.

**Not covered:** the interactive passphrase prompt described above. It needs a
passphrase-encrypted identity file and a terminal, and this repository is
unit-tests-only. I verified the mechanism by reading `KeyStore::new`,
`decrypt_mnemonic`, and age's `read_secret`, but it has not been exercised
end-to-end.

## Follow-ups, not in this PR

- **Seed discoverability.** A seed with no accounts derived from it appears in
  no listing: `z_listaccounts` derives `seedfp` from accounts, and
  `getwalletinfo`'s `mnemonic_seedfp` is a `"TODO"` stub. The only place a fresh
  fingerprint is shown is `generate-mnemonic` stdout. On a single-seed wallet
  every command falls back on the sole seed so this is invisible; on a
  multi-seed wallet, losing that output leaves the seed unnameable. This PR puts
  the fingerprint on the critical path for `export-mnemonic` and
  `confirm-backup`, which makes it worth fixing — probably alongside
  human-readable names for seeds.

Closes #201.


sellout:
I put up a related change in #730 and further work proposed in #731.

I think #731 would be good to do all the way through (give names & timestamps to seed fingerprints, and a way to list them).

zallet-core/src/components/keystore/db/migrations/mnemonic_backup_confirmation.rs
```diff
@@ -0,0 +1,44 @@
...
+        // that asks: `zallet confirm-backup` clears it.
+        transaction.execute_batch(
+            "ALTER TABLE ext_zallet_keystore_mnemonics
+            ADD COLUMN backup_confirmed INTEGER NOT NULL DEFAULT FALSE",
+        )?;
+        Ok(())
+    }
+
+    fn down(&self, _transaction: &rusqlite::Transaction<'_>) -> Result<(), Self::Error> {
+        Ok(())
```
schell, line 49: I think this should have a `DROP COLUMN` here, no?

schell (changes requested):
I think the migrations need to drop the newly added `backup_confirmed` on _down_. Otherwise, looks great to me!

zallet-core/src/components/keystore/db/migrations/mnemonic_backup_confirmation.rs
sellout: Whoops, it does. Also added a test for the migration.

schell:
@sellout I'm re-running CI jobs as it looks like some of them failed for... ...no reason.

schell (dismissed):

sellout:
@schell Rebased & re-running CI. There were definitely some GH Actions issues when this first went up.

schell (approved):

schell:
Thanks @sellout 🙇 !

PR #729, opened 2026-08-04
feat(rpc): Implement signmessage

### LINKED ISSUE #68: rpc: Implement `signmessage`

nullcopy:
Moving https://github.com/zcash/zallet/pull/359 to a local branch to run integration tests
Resolves #68 
Resolves COR-253

### DIFF
--- a/zallet-core/src/components/json_rpc/methods/sign_message.rs
+++ b/zallet-core/src/components/json_rpc/methods/sign_message.rs
@@ -1,0 +1,239 @@
+use base64ct::{Base64, Encoding};
+use documented::Documented;
+use jsonrpsee::core::RpcResult;
+use schemars::JsonSchema;
+use secp256k1::{Message, SECP256K1};
+use secrecy::ExposeSecret;
+use serde::Serialize;
+use transparent::{address::TransparentAddress, keys::TransparentKeyScope};
+use zcash_client_backend::{
+    data_api::{Account as _, WalletRead},
+    keys::UnifiedSpendingKey,
+};
+use zcash_keys::encoding::AddressCodec;
+
+use super::verify_message;
+use crate::components::{database::DbConnection, json_rpc::server::LegacyCode, keystore::KeyStore};
+
+/// Response to a `signmessage` RPC request.
+pub(crate) type Response = RpcResult<ResultType>;
+
+/// The result of signing a message.
+#[derive(Clone, Debug, Serialize, Documented, JsonSchema)]
+#[serde(transparent)]
+pub(crate) struct ResultType(String);
+
+pub(super) const PARAM_T_ADDR_DESC: &str = "The transparent address to use to look up the private key that will be used to sign the message.";
+pub(super) const PARAM_MESSAGE_DESC: &str = "The message to create a signature of.";
+
+/// Signs a message with the private key of a transparent address.
+pub(crate) async fn call(
+    wallet: &DbConnection,
+    keystore: &KeyStore,
+    t_addr: &str,
+    message: &str,
+) -> Response {
+    let transparent_addr = TransparentAddress::decode(wallet.params(), t_addr)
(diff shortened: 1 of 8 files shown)


zallet-core/src/components/json_rpc/methods/sign_message.rs
```diff
@@ -0,0 +1,239 @@
...
+        // Invalid state: scope and address_index should both be Some or both be None.
+        _ => Err(LegacyCode::Wallet.with_static("Private key not available")),
+    }
+}
+
+/// Maps keystore errors, converting "Wallet is locked" to the appropriate RPC error.
+fn map_wallet_locked_error(e: crate::error::Error) -> jsonrpsee::types::ErrorObject<'static> {
+    match e.kind() {
+        // TODO: Improve internal error types.
+        //       https://github.com/zcash/zallet/issues/256
```
schell, line 132: Here it is again! :)

schell (dismissed):
Looks good to me, but I think we want to add the `signmessage` JSON-RPC call to the changelog?

pacu (dismissed):
Reviewed and verified locally:

- **Signing format matches zcashd**: magic prefix + CompactSize framing + double SHA-256, compact recoverable signature, `31 + recovery_id` header for compressed pubkeys. The roundtrip test anchors against `verifymessage`, which itself has a real-world fixture signature.
- **Error codes match zcashd**: `-5 "Invalid Zcash address"`, `-3 "Address does not refer to key"`, `-13` unlock-needed, `-5 "Sign failed"`.
- **The keystore feature-gate swap is correct**: encrypt-side items are only used by the zewif import path (`zcashd-import`), while decrypt-side is needed by `signmessage` under the broader `transparent-key-import`. Confirmed `cargo check` passes with `--all-features`, with only `transparent-key-import`, and with no features.
- **Tests pass** (11 message-related tests), and the `PARAM_*_DESC` constants match the build.rs OpenRPC convention so `help`/`rpc.discover` pick the method up.
- Good defense-in-depth: pubkey-vs-address consistency check before signing, ephemeral (ZIP 320) refusal, view-only accounts correctly rejected.

### Possible follow-up tasks (non-blocking)

1. **Uncompressed imported keys always fail with an opaque `"Sign failed"`.** The zewif import path supports uncompressed transparent keys (`keystore/zewif.rs` verifies against `serialize_uncompressed`), so a migrated zcashd wallet can contain an address hashed from an uncompressed pubkey. The consistency check uses compressed serialization, so signing for such an address can never succeed. Refusing is defensible (our `verifymessage` rejects uncompressed signatures anyway), but worth confirming it's intentional and returning a clearer error or documenting the limitation.
2. **zcashd parity nits**: zcashd checks `EnsureWalletIsUnlocked` before parsing the address, and returns `"Private key not available"` (same `-4` code) for addresses not in the wallet where we return `"Unknown address"`. Strict-compat consumers might notice the message/ordering differences.
3. **Perf**: `UnifiedSpendingKey::from_seed` derives all pools just to obtain one transparent key; `transparent::keys::AccountPrivKey::from_seed` would be cheaper per call.
4. **Integration tests** for `sign_message::call()` once a wallet testing framework exists (already noted as a TODO in the code).

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

schell (dismissed):
utACK [171530b](https://github.com/zcash/zallet/pull/729/commits/171530bdd1f0f5c307369f26c6bc3f31ad16f079)

pacu (approved):
Re-approving after the rebase onto current main: `git range-diff` between the previously-approved `cf7a40a` and the new head `62dc811` shows no changes to the patch itself, only rebase context drift (CHANGELOG neighbors and a hunk header in `utils.rs`).

My verification notes and suggested follow-up tasks from the [earlier, now-dismissed review](https://github.com/zcash/zallet/pull/729#pullrequestreview-4878675252) still apply.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

PR #730, opened 2026-08-04
getwalletinfo: report the wallet's seed fingerprint

### LINKED ISSUE #620: rpc: Align `getwalletinfo` response with the #55 decision (placeholder fields serialize as real zeros)
**Current behavior:** `getwalletinfo` responds with placeholder values: the handler logs `TODO: Implement getwalletinfo` and returns zero for `balance`, `unconfirmed_balance`, `immature_balance`, `shielded_balance`, `shielded_unconfirmed_balance`, `txcount`, `keypoololdest`, `keypoolsize`, `walletversion: 0`, and `mnemonic_seedfp: "TODO"`. Only `unlocked_until` is computed (`zallet-core/src/components/json_rpc/methods/get_wallet_info.rs`).

#55 was closed (2026-07-13) with the decision that transparent balance data will **not** be added to `getwalletinfo` — users should use a dedicated balance retrieval method instead.

**Problem:** a caller today cannot distinguish "decided: use another method" from "balance is actually zero" — the placeholder fields serialize as real values. This is misleading for zcashd migrators whose monitoring reads `getwalletinfo.balance`.

**Expected:** align the response with the #55 decision:
1. Drop (or stop serializing) the balance fields that will never be populated, so callers fail loudly instead of reading zeros — or document their permanent absence.
2. Decide the fate of the remaining placeholder fields (`walletversion`, `txcount`, `keypoololdest`/`keypoolsize` — presumably 0 forever given no key pool, `mnemonic_seedfp`), and implement or remove them.
3. Record the resulting differences on the book's [altered semantics](https://zcash.github.io/zallet/zcashd/json_rpc.html) page (the status matrix from #602 already flags the method as partial).

Related: #55, #602, #600.

---
*Drafted with AI assistance (Claude); behavior verified against `get_wallet_info.rs` at a25cd85.*

sellout:
`getwalletinfo`'s `mnemonic_seedfp` field returned the literal string `"TODO"`,
so a caller reading it got a value that was neither a fingerprint nor an error.
It now reports the wallet's real ZIP 32 seed fingerprint.

Part of #620, which lists this among the method's placeholder fields.

## How this surfaced, and why it only half helps

This came out of the `walletrequirebackup` work (#201, #728), where I went
looking for a way to discover a wallet's seed fingerprints and found there
isn't one. A seed with no accounts derived from it appears in no listing:
`z_listaccounts` derives each account's `seedfp` from its key derivation, so a
seed with no accounts cannot appear there, and `getwalletinfo` — the one place
that looks like it should answer the question — said `"TODO"`. The only place a
fingerprint is ever shown is the stdout of `generate-mnemonic` /
`import-mnemonic`, at the moment the phrase is created.

That matters because #728 puts the fingerprint on the critical path:
`export-mnemonic` and `confirm-backup` both need to name a seed. An operator
who has lost that terminal output cannot name it, and cannot derive accounts.

**This change does not fix that**, and it is worth being explicit about why. A
fingerprint is only *needed* when a wallet holds more than one phrase — with a
single phrase, every command falls back on it and nobody has to name anything.
But this field can only report a fingerprint when there is exactly one phrase;
with several it reports how many, because the `zcashd`-inherited field has room
for one value. So it gives you the fingerprint precisely when you do not need
it, and withholds it precisely when you do.

What it does fix is the narrower bug that the field lied: `"TODO"` is not a
fingerprint, and a caller had no way to know that. Closing the discoverability
gap properly needs somewhere that can carry a list — a `list-seeds` command, or
a field that reports every fingerprint — most likely alongside human-readable
names, since fingerprints are not what an operator wants to tell wallets apart
by. A `TODO` in the code marks the multi-phrase case as waiting on that.

## Behaviour

| Wallet holds | `mnemonic_seedfp` |
|---|---|
| one mnemonic phrase | the ZIP 32 seed fingerprint |
| no mnemonic phrases | the field is omitted |
| several phrases | `"This wallet holds N mnemonic phrases"` |

Omitting the field when there is no mnemonic matches `zcashd`, which pushed
`mnemonic_seedfp` only when `GetMnemonicHDChain()` had a value rather than
emitting a placeholder — the same treatment it gave the optional
`legacy_seedfp` beside it.

**One deliberate divergence from `zcashd`:** the fingerprint is reported in its
ZIP 32 bech32m encoding (`zip32seedfp1…`), matching Zallet's own
`z_listaccounts`, whereas `zcashd` reported the `uint256` hex form.
`parse_seedfp` accepts both on input, so this only affects what a caller reads.
Say if you would rather match `zcashd` exactly.

## Scope

This does not close #620. `walletversion`, `txcount`, `keypoololdest`,
`keypoolsize` and the balance fields are still placeholders, and item 1 of that
issue — dropping the balance fields so callers fail loudly rather than reading
zeros — is a design decision rather than a mechanical change.

Also updates the `getwalletinfo` row in the `zcashd` status matrix, which
claimed only `unlocked_until` was meaningful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,20 @@ be considered breaking changes.
   through an authenticated, encrypted tunnel such as SSH port forwarding or a
   VPN. Previously any bind address was accepted without a warning.
 
+- `getwalletinfo`’s `mnemonic_seedfp` field now reports the wallet’s ZIP 32 seed
+  fingerprint, instead of the placeholder string `"TODO"`. It is omitted when
+  the wallet holds no mnemonic phrases, matching `zcashd`, which emitted the
+  field only when its wallet had a mnemonic.
+
+  A Zallet wallet may hold any number of phrases, which this `zcashd`-inherited
+  field cannot represent, so a wallet holding several gets a sentence saying how
+  many instead of a fingerprint. Use `z_listaccounts` to learn which phrase an
+  individual account derives from.
+
+  Note that Zallet reports the fingerprint in its ZIP 32 bech32m encoding
+  (`zip32seedfp1…`), as `z_listaccounts` already does, whereas `zcashd` reported
+  the `uint256` hex form.
+
 ### Fixed
 
 - `migrate-zcashd-wallet` now passes a fallback network to the wallet parser
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,14 +131,18 @@ be considered breaking changes.
   VPN. Previously any bind address was accepted without a warning.
 
 - `getwalletinfo`’s `mnemonic_seedfp` field now reports the wallet’s ZIP 32 seed
-  fingerprint, instead of the placeholder string `"TODO"`. It is omitted when
-  the wallet holds no mnemonic phrases, matching `zcashd`, which emitted the
-  field only when its wallet had a mnemonic.
+  fingerprint, instead of the placeholder string `"TODO"`. It is present only
+  when the wallet holds exactly one mnemonic phrase: `zcashd` had at most one per
+  wallet and omitted the field when it had none, while a Zallet wallet may hold
+  any number, and there is no one fingerprint to report when it holds several.
 
-  A Zallet wallet may hold any number of phrases, which this `zcashd`-inherited
-  field cannot represent, so a wallet holding several gets a sentence saying how
-  many instead of a fingerprint. Use `z_listaccounts` to learn which phrase an
-  individual account derives from.
+  A new `mnemonic_seedfps` field lists every ZIP 32 seed fingerprint the wallet
+  holds, in lexicographic order, so a wallet with several phrases is no longer
+  unreportable. It is empty when the wallet holds no mnemonic phrases, and is
+  omitted — along with `mnemonic_seedfp` — only when the key store could not be
+  queried, which is what distinguishes a wallet with no phrases from a failed
+  lookup. Use `z_listaccounts` to learn which phrase an individual account
+  derives from.
 
   Note that Zallet reports the fingerprint in its ZIP 32 bech32m encoding
   (`zip32seedfp1…`), as `z_listaccounts` already does, whereas `zcashd` reported
--- a/book/src/zcashd/rpc_status.md
+++ b/book/src/zcashd/rpc_status.md
@@ -31,7 +31,7 @@ Statuses:
 | `getreceivedbyaddress` | Not yet implemented | [#52](https://github.com/zcash/zallet/issues/52) |
 | `gettransaction` | Not planned | Superseded by `z_viewtransaction`, which now includes its top-level fields ([altered semantics](json_rpc.md#z_viewtransaction)); `gettransaction` cannot represent partially-shielded transactions correctly |
 | `getunconfirmedbalance` | Not yet implemented | [#54](https://github.com/zcash/zallet/issues/54) |
-| `getwalletinfo` | Implemented (partial) | Balance fields will not be populated — use dedicated balance methods ([#55](https://github.com/zcash/zallet/issues/55)). `unlocked_until` and `mnemonic_seedfp` are meaningful; the remaining fields are still placeholders ([#620](https://github.com/zcash/zallet/issues/620)) |
+| `getwalletinfo` | Implemented (partial) | Balance fields will not be populated — use dedicated balance methods ([#55](https://github.com/zcash/zallet/issues/55)). `unlocked_until` and the seed fingerprints are meaningful, the latter with [altered semantics](json_rpc.md#getwalletinfo) and a new `mnemonic_seedfps` field; the remaining fields are still placeholders ([#620](https://github.com/zcash/zallet/issues/620)) |
 | `importaddress` | Omitted | Use the [`zallet import-address`](../cli/import-address.md) CLI command ([#56](https://github.com/zcash/zallet/issues/56)), which accepts a bare address, a public key, or a redeem script, and imports into the legacy transparent account by default; `z_importaddress` covers the public-key and redeem-script forms over RPC |
 | `importprivkey` | Not yet implemented | [#57](https://github.com/zcash/zallet/issues/57) |
 | `importpubkey` | Omitted | Use `z_importaddress` |
--- a/book/src/zcashd/rpc_status.md
+++ b/book/src/zcashd/rpc_status.md
@@ -31,7 +31,7 @@ Statuses:
 | `getreceivedbyaddress` | Not yet implemented | [#52](https://github.com/zcash/zallet/issues/52) |
 | `gettransaction` | Not planned | Superseded by `z_viewtransaction`, which now includes its top-level fields ([altered semantics](json_rpc.md#z_viewtransaction)); `gettransaction` cannot represent partially-shielded transactions correctly |
 | `getunconfirmedbalance` | Not yet implemented | [#54](https://github.com/zcash/zallet/issues/54) |
-| `getwalletinfo` | Implemented (partial) | Balance fields will not be populated — use dedicated balance methods ([#55](https://github.com/zcash/zallet/issues/55)); most other fields are currently placeholders, and only `unlocked_until` is meaningful |
+| `getwalletinfo` | Implemented (partial) | Balance fields will not be populated — use dedicated balance methods ([#55](https://github.com/zcash/zallet/issues/55)). `unlocked_until` and `mnemonic_seedfp` are meaningful; the remaining fields are still placeholders ([#620](https://github.com/zcash/zallet/issues/620)) |
 | `importaddress` | Omitted | Use the [`zallet import-address`](../cli/import-address.md) CLI command ([#56](https://github.com/zcash/zallet/issues/56)), which accepts a bare address, a public key, or a redeem script, and imports into the legacy transparent account by default; `z_importaddress` covers the public-key and redeem-script forms over RPC |
 | `importprivkey` | Not yet implemented | [#57](https://github.com/zcash/zallet/issues/57) |
 | `importpubkey` | Omitted | Use `z_importaddress` |
--- a/zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
+++ b/zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
@@ -7,10 +7,7 @@ use serde::Serialize;
 use zcash_protocol::value::Zatoshis;
 
 use crate::components::{
-    json_rpc::{
-        server::LegacyCode,
-        utils::{JsonZec, value_from_zatoshis},
-    },
+    json_rpc::utils::{JsonZec, value_from_zatoshis},
     keystore::KeyStore,
 };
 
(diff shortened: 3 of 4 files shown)


zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -55,14 +58,45 @@ pub(crate) struct GetWalletInfo {
...
+    /// A Zallet wallet may hold any number of phrases, which this `zcashd`-inherited
+    /// field cannot represent. It is the fingerprint when the wallet holds exactly one,
+    /// and otherwise a sentence saying how many it holds. Use `z_listaccounts` to learn
+    /// which phrase an individual account derives from.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    mnemonic_seedfp: Option<String>,
 }
 
 pub(crate) async fn call(keystore: &KeyStore) -> Response {
     // https://github.com/zcash/zallet/issues/55
```
pacu: this issue es closed.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -55,14 +58,45 @@ pub(crate) struct GetWalletInfo {
...
     warn!("TODO: Implement getwalletinfo");
```
pacu, line 75: does this warning still hold? Then reopen issues #55 and document the remaining bits

CHANGELOG.md
```diff
@@ -40,6 +40,22 @@ be considered breaking changes.
...
   is currently usable.
 
+### Changed
+
+- `getwalletinfo`’s `mnemonic_seedfp` field now reports the wallet’s ZIP 32 seed
+  fingerprint, instead of the placeholder string `"TODO"`. It is omitted when
+  the wallet holds no mnemonic phrases, matching `zcashd`, which emitted the
+  field only when its wallet had a mnemonic.
+
+  A Zallet wallet may hold any number of phrases, which this `zcashd`-inherited
```
pacu, line 50: is this desirable or even useful? Given that partners are already migrating to new methods anyway this change does not seem terrible to implement on their side. Maybe we should document this change in the Zallet book and make a `getwalletinfo` that actually gives all the info instead of partial information that needs follow-up calls for users that store more than one seed phrase. 

Thoughts?

pacu (changes requested):

CHANGELOG.md
sellout: I should have cross-referenced here, but I opened #731 to cover what you’re describing here.

Some other context around this PR:

When I was working on #728, I made the changes in this PR, because there’s currently no way to find the seed fingerprints a wallet contains, but for #728, you only need to know a fingerprint when a wallet has more than one[^1], so this change doesn’t actually help. I figured this is still a useful change for most users & re: zcashd compatibility, so I kept it, but separated it from #728.

[^1]: Before #728, you needed to explicitly pass a fingerprint to create an account (and mirroring that, would have needed to pass a fingerprint to confirm the backup), but I changed it to default to the sole fingerprint in that case, which is what makes this change unhelpful for #728.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
sellout: It still holds, but #620 is the relevant issue now – I changed it to point there.

pacu (dismissed):

pacu (commented):
Automated review (Claude Code, posted by @pacu). Verification confirmed the PR's core claims — the bech32m encoding matches `z_listaccounts`' path through `zip32::SeedFingerprint`'s `Display`, `list_seed_fingerprints` counts only mnemonic-table seeds, and the zero-phrase omission matches `zcashd`'s contract. Five findings below, ranked by severity; the first two are behavioral, the rest are docs/cleanup.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -55,14 +58,45 @@ pub(crate) struct GetWalletInfo {
...
+    //       prose. `zcashd` had at most one mnemonic phrase per wallet, so this field is
+    //       shaped to hold one fingerprint; Zallet supports any number, and a caller
+    //       whose wallet holds several currently gets a sentence where it expects a
+    //       fingerprint. Replacing that case needs a field that can carry a list.
+    let mnemonic_seedfp = match seed_fps.as_slice() {
+        [seed_fp] => Some(seed_fp.to_string()),
+        // `zcashd` omitted this field when the wallet had no mnemonic, rather than
+        // reporting a placeholder; a wallet with no phrases has no fingerprint to give.
+        [] => None,
+        several => Some(format!(
```
pacu: A multi-mnemonic wallet returns `"This wallet holds 2 mnemonic phrases"` in a field whose contract is a seed fingerprint, and a machine caller has no structural way to tell the two apart: feeding the value to `parse_seedfp` fails with a bech32 parse error, and comparing it against `z_listaccounts` `seedfp` values silently never matches. The PR body documents this deliberately, but omitting the field in the multi-phrase case (as the `[]` arm already does) would be *distinguishable* where prose is not — and it avoids a second breaking change to this same field when the promised list-capable field ships.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -55,14 +58,45 @@ pub(crate) struct GetWalletInfo {
...
+    mnemonic_seedfp: Option<String>,
 }
 
 pub(crate) async fn call(keystore: &KeyStore) -> Response {
-    // https://github.com/zcash/zallet/issues/55
+    // https://github.com/zcash/zallet/issues/620
     warn!("TODO: Implement getwalletinfo");
 
+    let seed_fps = keystore
+        .list_seed_fingerprints()
```
pacu: `getwalletinfo` was previously infallible (in-memory reads only); this awaited DB call now fails the **entire** method with `LegacyCode::Database` (-20) on any keystore DB error. Since `getwalletinfo` is a common liveness/health poll inherited from `zcashd`, transient SQLite contention or pool-acquisition failure now costs the caller `unlocked_until` and every other computable field, and can trip a false "wallet down" alert. Consider degrading instead — log the error and omit `mnemonic_seedfp` — so the rest of the response survives.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -55,14 +58,45 @@ pub(crate) struct GetWalletInfo {
...
+    /// and otherwise a sentence saying how many it holds. Use `z_listaccounts` to learn
+    /// which phrase an individual account derives from.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    mnemonic_seedfp: Option<String>,
 }
 
 pub(crate) async fn call(keystore: &KeyStore) -> Response {
-    // https://github.com/zcash/zallet/issues/55
+    // https://github.com/zcash/zallet/issues/620
     warn!("TODO: Implement getwalletinfo");
```
pacu: This blanket `warn!("TODO: Implement getwalletinfo")` still fires on every call, and now contradicts this PR's own `rpc_status.md` edit saying `unlocked_until` and `mnemonic_seedfp` are meaningful. Narrowing the message to name the still-placeholder fields (balances, `txcount`, `keypool*`, `walletversion`) would keep it accurate and stop operators distrusting the fields that do work.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
```diff
@@ -55,14 +58,45 @@ pub(crate) struct GetWalletInfo {
...
 
+    let seed_fps = keystore
+        .list_seed_fingerprints()
+        .await
+        .map_err(|e| LegacyCode::Database.with_message(e.to_string()))?
+        .into_iter()
+        .collect::<Vec<_>>();
```
pacu: Minor: the `HashSet` is collected into a `Vec` solely to enable slice pattern matching, and the catch-all arm binds the whole slice just to call `.len()`. The sibling call site (`get_new_account.rs:87-96`) matches on `seed_fps.len()` and extracts the singleton with `into_iter().next()` — same job, no intermediate allocation, and keeps the two call sites consistent.

book/src/zcashd/rpc_status.md
```diff
@@ -31,7 +31,7 @@ Statuses:
 | `getreceivedbyaddress` | Not yet implemented | [#52](https://github.com/zcash/zallet/issues/52) |
 | `gettransaction` | Not planned | Superseded by `z_viewtransaction`, which now includes its top-level fields ([altered semantics](json_rpc.md#z_viewtransaction)); `gettransaction` cannot represent partially-shielded transactions correctly |
 | `getunconfirmedbalance` | Not yet implemented | [#54](https://github.com/zcash/zallet/issues/54) |
-| `getwalletinfo` | Implemented (partial) | Balance fields will not be populated — use dedicated balance methods ([#55](https://github.com/zcash/zallet/issues/55)); most other fields are currently placeholders, and only `unlocked_until` is meaningful |
+| `getwalletinfo` | Implemented (partial) | Balance fields will not be populated — use dedicated balance methods ([#55](https://github.com/zcash/zallet/issues/55)). `unlocked_until` and `mnemonic_seedfp` are meaningful; the remaining fields are still placeholders ([#620](https://github.com/zcash/zallet/issues/620)) |
```
pacu: The two semantic divergences from `zcashd` (bech32m `zip32seedfp1…` instead of `uint256` hex, and the possible English-sentence value) are documented only in CHANGELOG.md. The repo's established registry for exactly this is `book/src/zcashd/json_rpc.md` ("altered semantics"), which many rows in this table link to for smaller divergences (e.g. `z_listunspent`'s field rename, `z_exportviewingkey`'s encoding change). Suggest adding a `getwalletinfo` entry there and linking it from this row, so migrating operators find the format change in the compat docs rather than one release's changelog.

sellout: Done.

I used bech32m here instead of uint256 because it’s what other Zallet responses use, including in `listaddresses`, which existed in zcashd and thus has the same divergence. However, that divergence isn’t documented, so I added a commit to document it for `listaddresses`.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
sellout: Done.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
sellout: Previously, this field just always printed “TODO”, so I figured prose wasn’t off the table. Omitting the field wouldn’t be distinguishable because an omitted `mnemonic_seedfp` already means “no mnemonic”. The prose was at least distinguishable in that it’s not a valid fingerprint. It seemed the best option without further changes. Although if we’re expecting zcashd-era callers to blindly handle the bech32m encoding instead of uint32, I guess any prose (including “TODO”) is problematic.

But … to make things more well-defined, I instead added a `mnemonic_seedfps` field that returns an array of fingerprints, which I had previously considered future work. So, the cases are:

- no mnemonic
    ``` json
    { ...
      "mnemonic_seedfps": []
    }
    ```
- one mnemonic
    ``` json
    { ...
      "mnemonic_seedfp": "fingerprint1",
      "mnemonic_seedfps": ["fingerprint1"]
    }
    ```
- multiple mnemonics
    ``` json
    { ...
      "mnemonic_seedfps": ["fingerprint1", "fingerprint2"]
    }
    ```
- and when there’s a keystore error (to avoid failure)
    ``` json
    { ...
    }
    ```

which should all maintain compatibility with zcashd’s response without conflating the various outcomes for a Zallet user.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
sellout: This is no longer true – the `Vec` is now used, due to `mnemonic_seedfps`.

zallet-core/src/components/json_rpc/methods/get_wallet_info.rs
sellout: The problem is that the failure (which couldn’t happen in zcashd at all) would then be indistinguishable from the no-seed case (which omits).

That is, zcashd only had two possible outcomes (mnemonic or no mnemonic) while Zallet has four (adding multiple mnemonics and keystore error). And zcashd’s response structure has no wiggle room for indicating an unexpected outcome.

The addition of `mnemonic_seedfps` makes all four outcomes distinguishable, but a caller has to be Zallet-aware to make the distinction. To a zcashd-era caller, a missing `mnemonic_seedfp` _only_ means “no mnemonic phrase”, but in Zallet it also means “multiple phrases” or “keystore error”.

So, I‘m not sure if that’s good enough or not. Using other JSON values (e.g., `null` for keystore error and `3` if there’s three phrases) could be more distinguishable, but could also cause a caller to fail when just claiming that there’s no mnemonic phrase is a safe fallback.

pacu (approved):

PR #734, opened 2026-08-05
init-wallet-encryption: canonicalize and validate age recipients before storing

pacu:
The recipients stored by `init-wallet-encryption` were obtained by writing an age recipients file to a buffer and splitting it on newlines, storing **every** line as a recipient row. The age recipients-file format permits `#` comments and blank lines, so this was an unvalidated assumption about `IdentityFile::write_recipients_file`'s output. It holds for the identities we generate today (so this is latent rather than live), but nothing guarantees it.

Compounding it, `initialize_recipients` inserted whatever it was handed without parsing it, so any bad entry would make `init-wallet-encryption` report success and the wallet would only break at the first attempt to encrypt key material (e.g. `generate-mnemonic`), with an error pointing at the keystore rather than at initialization.

## What changes

- A `canonicalize_recipients_file` helper in the keystore turns recipients-file contents into bare recipient strings: lines are trimmed, blank lines and `#` comments are dropped, and `@`-prefixed entries are rejected (they denote indirection through an external source, which would make the effective recipient set depend on state outside the stored set itself). `init-wallet-encryption` now stores its output instead of the raw lines.
- `initialize_recipients` validates up front that the recipient set is non-empty and parses via the same path the encryptor uses, so a bad set fails initialization with a clear error instead of deferring the failure to first use.
- New `KeystoreError::{EmptyRecipients, RecipientIndirection}` variants with Fluent messages; the exhaustive match in `walletpassphrase` maps them faithfully even though `unlock` cannot produce them.

## Test evidence

- New unit tests over the canonicalization helper: comments, blank lines, leading/trailing whitespace, CRLF, comment-only input, and `@`-prefixed entries.
- All three workspace legs pass locally: root (`cargo fmt`/`clippy -D warnings`/`test`, 319 tests), zebra backend (including the `generate_encryption_identity_drives_setup` acceptance test, which exercises `init-wallet-encryption` → `generate-mnemonic` end to end), and zaino backend.
- `cargo +beta clippy --tests --all-features --all-targets` introduces no new warnings (the three `useless_format` warnings it reports are preexisting, in files this PR does not touch).

No manifests or lockfiles change, so librustzcash lockstep is unaffected.

https://claude.ai/code/session_01AdVA42x8NdPH5mnFVehqJQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,15 @@ be considered breaking changes.
 
 ### Fixed
 
+- `init-wallet-encryption` now canonicalizes the age recipients derived from
+  the identity file before storing them, and validates that the resulting set
+  is non-empty and parseable. Previously every line of the internally generated
+  recipients file was stored verbatim as a recipient; the format permits
+  comments and blank lines, and a stray entry would only surface as an error
+  the first time the wallet tried to encrypt key material (e.g. during
+  `generate-mnemonic`) rather than failing initialization. `@`-prefixed
+  recipient entries are rejected, as indirection through external sources
+  cannot be part of the wallet's stored recipient set.
 - The queue of asynchronous RPC operations (`z_sendmany`, `z_shieldcoinbase`)
   is now bounded (GHSA-3wr9-v982-59fj). Finished operations are still retained
   until collected with `z_getoperationresult`, but once the queue reaches its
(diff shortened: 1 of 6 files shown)


nullcopy (approved):
Good catch!
utACK 583de84c573c8d471293f2121db16c4db886ff3b

PR #735, opened 2026-08-05
rpc: Bind the signing key to the account record before signing (#642 item 2)

pacu:
## Summary

Implements **item 2 of #642**, which was closed by #643 after landing only items 1 and 3.

Input selection runs against the account record in `wallet.db`; signing uses a key derived from the decrypted seed. Those are the same key only if that record is intact, and it is not integrity-protected — a modified `accounts.ufvk` would have the wallet select one key's notes and sign with another's, and nothing checks.

`WalletRead::validate_seed` exists for precisely this and was never wired up: on `main` it is reachable only as a trait-forwarding shim on `DbConnection` (`connection.rs:223-228`) called from nowhere. This calls it between decrypting the seed and deriving the spending key, and fails closed.

## What changed

All three account-based spend paths repeated the same block — resolve the derivation, decrypt the seed, derive the USK — with slightly different error mapping. That is now `payments::spending_key_for_account`, with the check in one place:

- `z_sendmany` (`z_send_many.rs`)
- `z_shieldcoinbase` (`z_shieldcoinbase.rs`)
- `z_sendfromaccount` (`z_send_from_account.rs`)

Net effect on those three files is −16, −18 and −15 lines respectively; the shared helper is the only place the check can be forgotten.

## How a mismatch is reported

Upstream `validate_seed` signals a mismatch in two different shapes, and getting this wrong makes the check silently useless:

- `Ok(false)` — neither the recorded seed fingerprint nor the recorded viewing key matches the seed.
- `Err(CorruptedData)` — **exactly one** of them matches.

The substitution attack always lands in the second. `derivation` is read from the same account row being validated, so the seed is decrypted under that row's own `hd_seed_fingerprint`; the fingerprint therefore always agrees, leaving a swapped `ufvk` as the sole disagreement. An earlier revision of this PR only handled `Ok(false)` and mapped the error arm to `LegacyCode::Database`, which meant `err-account-seed-mismatch` was effectively unreachable on the real attack path — the operator saw `Data DB is corrupted: Seed fingerprint match: true, uivk match: false` instead.

Both shapes now report `err-account-seed-mismatch`, phrased like the existing tamper-detection errors: *"The wallet database is corrupted or has been tampered with. No transaction has been created."*

Every other error — a busy or unreadable database, a seed of unusable length — is **not** treated as tampering, since that message opens by asserting the recorded key disagrees, which in those cases is untrue. Those surface as `LegacyCode::Database` and are logged. The check still fails closed either way: `CorruptedData` is the only error meaning "the recorded key disagrees", so anything else leaves the binding unverified and no spending key is derived.

## Why `pczt_sign` is excluded

`pczt_sign` resolves its key from the PCZT's signing hints (`hints.seed_fp`, `hints.account_idx`), not from an account — there is no account record to bind the seed to, so `validate_seed` has no applicable argument. It also deliberately tolerates "this wallet does not hold that seed" as the multi-party co-signing case, which a naive check would break.

`z_sendfromaccount` drives `pczt_sign` over its own PCZT and binds before doing so, so the one-shot path is covered. A comment at the hint-resolution site records this.

## Cost

None at any call site. The seed is already decrypted at each one, so no operation loads a secret it did not already load — the constraint #642 set for these checks. A locked wallet still fails at the decryption step, not at this one. View-only and imported accounts are unaffected: they hard-fail earlier at `key_derivation()`.

## Testing

Two tests, pinning the two halves of the behaviour the helper depends on:

- `validate_seed_accepts_only_the_accounts_own_seed` (`components/database/tests.rs`) pins the `Ok(false)` semantics for an unrelated seed, so the fail-closed branch is the one that runs.
- `spending_key_for_account_rejects_tampered_uivk` (`components/json_rpc/payments.rs`) exercises the actual attack: it overwrites an account's `ufvk` with a key derived from an unrelated seed at the same account index, leaving `hd_seed_fingerprint` intact. It asserts both that upstream still classifies this as `Err(CorruptedData)` and that the spend path surfaces it as the tamper message rather than a generic database error.

Together these catch an upstream change of meaning rather than letting it silently turn the check into a no-op.

- `cargo test -p zallet-core --lib --all-features` — 422 passed, 0 failed
- `cargo fmt --check` and `cargo clippy --all-targets` clean
- Test build clean under both `zallet_build = "wallet"` and `zallet_build = "merchant_terminal"`

## Note on #642

#642 is closed but items **2** (this PR) and **4** (the age recipient set) were never implemented. Item 4 remains open after this lands; the issue is worth reopening or splitting so the remaining item is tracked somewhere visible.

https://claude.ai/code/session_01Qug9tBVf49c97BX6JbQwk4

### DIFF
--- a/zallet-core/src/components/json_rpc/payments.rs
+++ b/zallet-core/src/components/json_rpc/payments.rs
@@ -25,7 +25,7 @@ use zcash_client_backend::{
     wallet::TransparentAddressSource,
     zip321::{Payment, TransactionRequest},
 };
-use zcash_client_sqlite::{AccountUuid, ReceivedNoteId, wallet::Account};
+use zcash_client_sqlite::{AccountUuid, ReceivedNoteId, error::SqliteClientError, wallet::Account};
 use zcash_keys::{
     address::Address,
     keys::{UnifiedFullViewingKey, UnifiedSpendingKey},
@@ -168,11 +168,18 @@ pub(super) fn confirmations_policy_for_minconf(
 /// that gap by checking the recorded viewing key against the one this seed derives, and
 /// fails closed.
 ///
-/// `validate_seed` returns `Ok(false)` for a seed that does not derive the recorded key,
-/// and `Err(CorruptedData)` when the seed fingerprint matches but the viewing key does
-/// not (the real substitution attack: same seed, swapped ufvk row). Both are treated as
-/// corruption or tampering: the operation is refused with `err-account-seed-mismatch`
-/// and no spending key is derived.
+/// `validate_seed` reports a mismatch in two different ways and both must be caught.
+/// `Ok(false)` is neither the recorded seed fingerprint nor the recorded viewing key
+/// matching this seed; `Err(CorruptedData)` is exactly one of them matching. The
+/// substitution attack lands in the latter: `derivation` is read from the same account
+/// row being checked, so the seed is decrypted under that row's own fingerprint and the
+/// fingerprint always agrees, leaving a swapped `ufvk` as the sole disagreement. Both
+/// refuse the operation with `err-account-seed-mismatch`.
+///
+/// Any other error means the check could not be performed — a busy or unreadable
+/// database, a seed of unusable length — which is not evidence of tampering and is not
+/// reported as such. It still fails closed: a spending key is derived only if the check
+/// affirmatively passed.
 ///
 /// The check is free at every call site: the seed is decrypted here anyway, so no
 /// operation loads a secret it did not already load, and a locked wallet still fails at
@@ -197,11 +204,22 @@ pub(super) async fn spending_key_for_account(
 
     match wallet.validate_seed(account_id, &seed) {
         Ok(true) => {}
-        Ok(false) | Err(_) => {
+        Ok(false) | Err(SqliteClientError::CorruptedData(_)) => {
             return Err(LegacyCode::Wallet.with_message(fl!(
                 "err-account-seed-mismatch",
                 account = account_id.expose_uuid().to_string(),
             )));
+        }
+        Err(e) => {
+            // The seed could not be checked against the account. Report it as what it is
+            // rather than as tampering, but still refuse to sign: `CorruptedData` is the
+            // only error that means "the recorded key disagrees", so every other error
+            // leaves the binding unverified.
+            tracing::error!(
+                "Could not validate the seed for account {}: {e}",
+                account_id.expose_uuid(),
+            );
+            return Err(LegacyCode::Database.with_message(e.to_string()));
         }
     }
 
--- a/zallet-core/src/components/json_rpc/payments.rs
+++ b/zallet-core/src/components/json_rpc/payments.rs
@@ -4,12 +4,13 @@ use abscissa_core::Application;
 use jsonrpsee::core::JsonValue;
 use jsonrpsee::{core::RpcResult, types::ErrorObjectOwned};
 use schemars::JsonSchema;
+use secrecy::ExposeSecret;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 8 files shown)


schell:
I'm working on this one, @pacu. I'm adding a fixup commit for a few issues and depending on the result I may have more after that.

schell:
@pacu Pushed two fixups.
1. err-account-seed-mismatch was unreachable on the real attack path (e1b0afc)
Upstream `validate_seed` returns `Err(CorruptedData)` when the seed fingerprint matches but the derived viewing key does not match the recorded one, the real attack (same seed, swapped accounts.ufvk). The previous map_err to LegacyCode::Database swallowed that into a generic error, so the tamper message was never shown. Now maps both Ok(false) and Err(_) to err-account-seed-mismatch.
2. Test for the tamper path (5ed9aba)
`spending_key_for_account_rejects_tampered_uivk` tampers account A's ufvk column to account B's (same seed fingerprint, different viewing key) and asserts the helper returns the tamper message, not a generic database error. Also corrected the comment on the existing test, which mischaracterized the different-seed case as "the substitution the check exists to catch."

zallet-core/src/components/json_rpc/payments.rs
```diff
@@ -154,6 +158,79 @@ pub(super) fn confirmations_policy_for_minconf(
...
+        .map_err(|e| match e.kind() {
+            // TODO: Improve internal error types.
+            //       https://github.com/zcash/zallet/issues/256
+            crate::error::ErrorKind::Generic if e.to_string() == "Wallet is locked" => {
+                LegacyCode::WalletUnlockNeeded.with_message(e.to_string())
+            }
+            _ => LegacyCode::Database.with_message(e.to_string()),
+        })?;
+
+    match wallet.validate_seed(account_id, &seed) {
```
nullcopy: I don't think `validate_seed` is strict enough to be the binding check. At the pinned librustzcash rev (`1f6bb207`), it goes to `seed_matches_derived_account` (`zcash_client_sqlite/src/wallet.rs:381-437`), and that treats the recorded UFVK as matching if **any one** component matches:

```rust
orchard_match || sapling_match || p2pkh_match
```

So someone who can write to `wallet.db` can rewrite `accounts.ufvk` as a UFVK that keeps the real transparent (or Sapling) component and puts in their own Orchard FVK, leaving `hd_seed_fingerprint` alone. Then:

- The fingerprint matches and one component matches, so `validate_seed` returns `Ok(true)` and this helper returns the USK.
- Input selection and change addresses use the substituted Orchard component, so shielded change goes to the attacker's key.
- `verify_and_broadcast_transactions` only checks transparent outputs against the seed-derived UFVK, so nothing catches it later.

Both new tests replace the whole UFVK with one from an unrelated seed, which is the only case upstream rejects. That's why they pass.

Suggested fix: derive the UFVK from the seed (`usk.to_unified_full_viewing_key()`) and require that the recorded UFVK is component-for-component equal to it: every component present in both must match, with none missing or added. A test that swaps only the Orchard component would cover this.

nullcopy (commented):
Thanks for this, the plumbing looks right. All three spend paths (`z_sendmany`, `z_sendfromaccount`, `z_shieldcoinbase`) go through the new helper, a locked wallet still fails at seed decryption, and the `CorruptedData` arm matches what `DbConnection::validate_seed` returns.

There is one blocking concern, noted inline: `validate_seed` accepts a partially substituted UFVK, so the binding can be bypassed.

Minor, non-blocking: upstream `validate_seed` returns `Err(UnknownZip32Derivation)` for imported spending accounts that have a derivation, and this helper would turn that into a refusal to spend. Zallet doesn't create such accounts today (`import_key` passes `derivation: None`), but it may be worth a comment.

nullcopy (changes requested):
NACK 834765c103421de530ad212d0194d71c67b3e726

PR #736, opened 2026-08-05
cli: Lock the data directory during zcashd wallet migration

pacu:
## Summary

`migrate-zcashd-wallet` was the only command that writes to the wallet database without holding the data directory lock.

Every other mutating command takes it — `start` (`start.rs:28`), `generate-mnemonic`, `import-mnemonic`, `export-mnemonic`, `init-wallet-encryption` — but the migration, which writes accounts, transparent key material and exposure information into that same database, did not. Running it alongside a live `zallet start` interleaved writes into a database neither process had exclusive use of.

## Change

One `config.lock_datadir()?` at the top of `MigrateZcashdWalletCmd::run_with`, bound for the whole function so the lock is held for the entire migration rather than just the opening.

## Notes

- **Behaviour change:** running the migration while another Zallet instance holds the datadir now fails with the existing `err-init-zallet-already-running` message instead of proceeding. That is the intent — it is the same guard every other wallet-writing command already applies.
- No double-lock risk: `run_with` is reached only via `AsyncRunnable::run` → `chain_runtime().run_migrate_zcashd_wallet()`, and nothing on that path takes the lock. It is the only `lock_datadir` call in the file.
- `lock_datadir` also restricts the datadir to `0700`, so a migration into a fresh datadir now gets the same permissions as one created by any other command.

## Testing

`cargo test -p zallet-core --lib --all-features` — 331 passed, 0 failed. `cargo fmt --check` and `cargo clippy` clean.

https://claude.ai/code/session_01L2qGvpzxEM1Vk1XNEz7Xg1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -261,6 +261,12 @@ be considered breaking changes.
   `generate-mnemonic`) rather than failing initialization. `@`-prefixed
   recipient entries are rejected, as indirection through external sources
   cannot be part of the wallet's stored recipient set.
+- `migrate-zcashd-wallet` now acquires the data directory lock for the duration
+  of the migration. It writes accounts and transparent key material into the
+  same wallet database `zallet start` uses, so running the two concurrently
+  could interleave writes; the command now reports that Zallet is already
+  running instead.
+
 - The queue of asynchronous RPC operations (`z_sendmany`, `z_shieldcoinbase`)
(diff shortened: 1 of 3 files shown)


schell (approved):
utACK [fa03a9e](https://github.com/zcash/zallet/pull/736/commits/fa03a9e81fab4d5d250cf39609a55c468391f473)

PR #742, opened 2026-08-05
migrate-zcashd-wallet: fail when the import leaves wallet material behind

### LINKED ISSUE #740: migrate-zcashd-wallet: should fail loudly when the import skips accounts or drops key material
## Problem

`zcash_client_sqlite::zewif::import_wallet` reports — but does not error on — skipped
accounts and unregistered keys (`ZewifImportReport::skipped_accounts`,
`skipped_transparent_keys`). Zallet logs these as warnings (`log_import_report`) and exits 0.

For a pre-Sapling seedless wallet today, the result is a migration that "succeeds" while
importing nothing: the only account is skipped, every spending key is dropped as
`NoOwningAccount`, and the wallet DB is empty — yet the keystore already holds the secrets
(they are persisted before the DB import), leaving orphaned key material. A user who trusts
the exit code could reasonably believe their funds migrated.

This is independent of the seedless fix: any future skip path has the same failure shape.

## What to implement

After `import_wallet`, treat the report as part of the command's success criteria:

* If no account was imported but the document contained accounts, exit nonzero with a clear
  "nothing was migrated" message.
* If accounts were skipped or transparent keys could not be registered, either exit nonzero
  or require an explicit acknowledgment flag (e.g. `--allow-partial-import`), enumerating
  exactly what was left behind and that it exists only in the original `wallet.dat`.
* Keep the keystore/DB consistency story in view: if the DB import registers nothing, the
  secrets persisted to the keystore beforehand should be called out (or rolled back).

nullcopy:
Closes #740.

`import_wallet` reports — but does not error on — skipped accounts and unregistered
keys, and zallet logged those as warnings and exited 0. For a pre-Sapling seedless
wallet that meant a migration that "succeeded" while importing nothing.

The import report is now part of the command's success criteria, evaluated by a pure
function (`check_import_report`) right after the report is logged:

- Hard failure, no override: the document contained accounts but none imported —
  "nothing was migrated; your funds remain only in wallet.dat".
- Failure unless the new `--allow-partial-import` flag is passed: skipped accounts or
  unregistered transparent spending keys — both are spendable material silently
  stranded in wallet.dat. The error enumerates each item.
- Remain warnings: unrepresentable watch-only redeem scripts and transactions without
  raw data — representability limits, not loss of spendable key material.

Error wording is matched to what actually happens on failure: secrets are persisted to
the keystore before the DB import, so the nothing-imported message notes the orphaned
keystore material is harmless (all keystore inserts are idempotent) and a re-run after
fixing the cause is safe. The partial-import message does not claim that — the DB
import commits accounts before the report can be evaluated, so a re-run trips the
duplicate-import check; it says so, and that accepting a partial migration means
redoing it against a fresh wallet with the flag. Rolling back the committed import on
partial failure is future work that would make the flag a simple retry knob.

Unit tests cover all report shapes without a database; CLI docs and CHANGELOG updated.

https://claude.ai/code/session_01S7s9MQzQ3Bmp42uabiSDbU

[COR-1780](https://linear.app/zodl/issue/COR-1780)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ be considered breaking changes.
   (default 100).
 - `getwalletstatus` now reports a `locked` field indicating whether the wallet
   is currently usable.
+- `migrate-zcashd-wallet --allow-partial-import` flag, permitting a migration
+  to complete even when some accounts or transparent spending keys in the
+  `zcashd` wallet could not be imported. The skipped items are reported as
+  warnings; they remain accessible only via the original `wallet.dat` file.
 
 ### Fixed
 
@@ -59,6 +63,15 @@ be considered breaking changes.
   rejected while the queue is full of unfinished operations. Previously the
   queue was unbounded, allowing an authenticated caller to exhaust wallet
   memory by starting operations without collecting their results.
+- `migrate-zcashd-wallet` now fails instead of exiting successfully when the
+  import leaves wallet material behind. A migration that imports no accounts at
+  all from a wallet that contains some is always an error; a migration that
+  skips some accounts or transparent spending keys is an error unless the new
+  `--allow-partial-import` flag is passed. The partial-import error enumerates
+  what was left behind; both errors note that the skipped material exists only
+  in the original `wallet.dat` file. Previously these outcomes were logged as
+  warnings and the command exited 0, so a wallet that migrated nothing looked
+  like a success.
 
 ## [0.1.0-beta.2] - 2026-07-28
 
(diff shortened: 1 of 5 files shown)


nullcopy (commented):
Self-ack e4b6059317d417c58ef037216178d3399fe90ebf

schell (approved):
utACK [452224b](https://github.com/zcash/zallet/pull/742/commits/452224bb0931d9db3f46f49b9286931efc956d35)

Issue #587, opened 2026-07-11
zebra backend: wallet never detects a shrinking node tip after invalidateblock (regtest)


zcash-harry:
## Summary

The zebra-backend wallet never detects that the node's chain tip shrank after `invalidateblock`, and `getwalletstatus().wallet_tip` stays pinned at the pre-invalidate height indefinitely.

## Reproduction

Test: [`qa/rpc-tests/wallet_ironwood_reorg.py`](https://github.com/zcash/integration-tests/blob/0121a30f09d393ee5351499bcef0c2290a183dbd/qa/rpc-tests/wallet_ironwood_reorg.py) in zcash/integration-tests (currently in `DISABLED_SCRIPTS`, PR #159). Single zebrad + zallet-zebra regtest node:

1. Shield coinbase into an Ironwood note, bury it 5 blocks, confirm it's spendable.
2. `node.invalidateblock(confirming_block)` — the node's tip drops below the block that created the note.
3. `wait_for_wallet_sync(node, w)` — polls `getwalletstatus()` until `wallet_tip.height == node.getblockcount()`. **Times out (60s default).**

## What was observed

Built zallet from `main` @ `d168efe1b3cd0537a061ff16c2f430d137d98e60` (past #560, #563, and #576), `zallet-zebra` backend, against a locally built `zebrad` (`ZcashFoundation/zebra` @ `054da05600ff9ea46fcd55829fa41e18369a259b`). Ran the test standalone (not through the shared CI cache) — `wait_for_wallet_sync` never converges; the wallet's reported tip never moves after the invalidate.

## What's been ruled out

- **#560** ("Retry on BlockConflict instead of crashing the whole wallet") fixes a *different* failure mode: a **replacing** block whose hash conflicts with what's already stored at a height (`SqliteClientError::BlockConflict`). A bare `invalidateblock` with nothing yet mined over the shortened tip may never produce a conflicting write, so `steady_state` may simply have nothing that tells it the node's view changed at all.

## Ask

Root-cause why the wallet doesn't notice/follow a shrinking node tip in this scenario (as opposed to a reorg where a new block already replaces the old one), and whether `steady_state` needs an explicit tip-height/hash poll (not just a reactive check on the next write) to catch this case. Test is ready to re-enable once this is fixed — see `qa/rpc-tests/wallet_ironwood_reorg.py`'s header comment for the exact repro.

PR #744, opened 2026-08-07
Stop an unreadable note commitment tree from silently corrupting the wallet

### LINKED ISSUE #743: zebra backend: empty Ironwood frontier substituted for an unreadable tree corrupts the wallet's shardtree, then crash-loops
## Summary

A user running **zallet v0.1.0-beta.2** with the **zebra read-state backend** on **mainnet** hit a<br>fatal sync error and now crash-loops on every restart. `zallet --datadir … repair truncate-wallet 3435000` clears it.

```
2026-08-06T11:35:57.460205Z  INFO steady_state: zallet_core::components::sync::steps:
    Scanning block 3438009 (00000000006260fabc6360bf1de15462e5e67f77dd78838df5f1fba22c80819d)
2026-08-06T11:35:58.015368Z  INFO zallet_core::commands::start: Wallet steady-state sync task exited
    wallet_sync_result=Err(Error(Context { kind: Sync, …
      source: Some(Other(PutBlocksCommitmentTree {
          pool: Ironwood,
          block_range: BlockHeight(3438008)..BlockHeight(3438010),
          error: Insert(Conflict(Address { level: Level(3), index: 3660 })) })) }))
2026-08-06T11:35:58.015399Z  INFO zallet_core::commands::start: Exiting Zallet because an ongoing task exited
```

This is **not** zcash/zallet#126 (see below) and **not** zcash/zallet#587.

## Decoding the error

| Fact | Source |
| -- | -- |
| `block_range` = `from_state.block_height() .. last_scanned_height + 1` | `zcash_client_backend` `data_api/ll/wallet.rs`, `PutBlocksError::ShardTreeForBlockRange` construction |
| ⇒ frontier taken **as of 3438008**; exactly **one** block applied: **3438009** | decode |
| `Address(level 3, index 3660)` = Ironwood note positions **29280..29287** (`index << level`) | `incrementalmerkletree` `Address::position_range_start/end` |
| That address is exactly the **level-3 ommer** of a frontier whose leaf is at position 29288..29295 | `shardtree` `from_frontier_parts` |
| ⇒ the failing insert is `insert_frontier`, not `insert_tree` | `update_tree` in `ll/wallet.rs` |

Both blocks are canonical mainnet (3438008 = `00000000001b2e7a…0df0`, 3438009 =<br>`000000000062 60fa…819d`), so this is not a fork at the failure point.

**Conclusion: the wallet's stored Ironwood tree was already wrong before block 3438009 arrived.**<br>The crash is where the damage becomes visible, not where it happened. NU6.3 activated on mainnet at<br>3,428,143, so this is ~9.9k blocks post-activation, and an Ironwood tree of ~29.3k commitments is<br>consistent with that.

## Root cause: nothing validates `from_state`

`put_blocks` looks like it validates the chain state it is handed — it checks<br>`from_state.tree_size() + commitments.len() == block.final_tree_size()`. **In Zallet's usage that****<br>****check cannot fail**, because both sides derive from the same value:

* Zallet builds `prior_block_metadata` **from** `from_state` —<br>`zallet-core/src/components/sync/steps.rs` passes `from_state.final_ironwood_tree().tree_size()`.
* `zcash_client_backend::scanning::full::scan_block` derives the scanned block's<br>`ironwood_final_tree_size` **from that same metadata** (`tree_sizes_around(…, prior_block_metadata.and_then(|m| m.ironwood_tree_size()), …)`).

So a wrong `from_state` frontier of *any* size — **including an empty one** — passes the<br>`NonSequentialBlocks` gate and is committed to the wallet's shardtree. It only detonates later, as a<br>`Conflict`, when a correct frontier disagrees with it. By then the wallet needs a manual rewind.

Every guarantee about `from_state` therefore has to come from the chain backend — and the zebra<br>backend does not provide one:

```rust
// backends/zebra/src/chain.rs, tree_state_as_of
let pinned_finalized = height <= self.finalized_floor;
…
let final_ironwood_tree = match self.reader.ironwood_tree_bytes(hash).await? {
    Some(bytes) => read_commitment_tree::<…>(&bytes[..])…,
    None if pinned_finalized => CommitmentTree::empty(),   // <-- no activation guard
    None => return Err(ChainError::unavailable("pinned ironwood treestate reorged away")),
}
```

`finalized_floor = tip.height() - MAX_REORG_DEPTH`, and the failing block was ~7h behind the tip, so<br>`pinned_finalized` was **true**. If Zebra returns `None` for a post-activation height, Zallet<br>substitutes an **empty** Ironwood frontier and the wallet writes commitments at wrong positions.

### Why Zebra doesn't stop this

`zebra-state` 12.0.1 *does* guard the height path — `ironwood_tree_by_height` panics rather than<br>masking a missing post-NU6.3 tree:

> "From NU6.3 activation onward a missing tree is a real invariant violation (corruption or an<br>interrupted backfill), so fail loudly instead of masking it."

But Zallet reads by **hash** (`ReadRequest::IronwoodTree(HashOrHeight::Hash(..))`), and:

```rust
pub fn ironwood_tree_by_hash_or_height(&self, hash_or_height: HashOrHeight) -> Option<…> {
    let height = hash_or_height.height_or_else(|hash| self.height(hash))?;  // <-- short-circuits
    self.ironwood_tree_by_height(&height)                                    // <-- guard lives here
}
```

The `?` returns `None` **before** the guard ever runs. So any hash the finalized state cannot resolve<br>yields a silent `None` — which Zallet then turns into an empty frontier. Zallet is masking exactly<br>what Zebra treats as fatal.

The same unguarded arm exists for Sapling and Orchard. Ironwood surfacing first is consistent with<br>`update_tree` being called per pool with Sapling and Orchard succeeding.

There is precedent: this failure class was already fixed once in this backend (`f8dea0e`, "backend:<br>read the Ironwood note commitment tree in the zebra backend"), and the beta.1 CHANGELOG entry for it<br>names this exact error string. A partial version survived in the `None if pinned_finalized` arm.

## Why it loops forever, rather than recovering

1. `put_blocks` is transactional and rolls back, so a restart reproduces byte-identically.
2. `steady_state` recovers from `SqliteClientError::BlockConflict` (fork-point search + truncate),<br>but `PutBlocksCommitmentTree` falls through to `return Err(error)`.
3. `commands/start.rs` hardcodes `exit_when_task_finishes = true`, so any sync-task exit ends the<br>process.

Note that naively routing `PutBlocksCommitmentTree` into the existing `BlockConflict` arm would make<br>things *worse*, not better: for tree corruption on the canonical chain, `build_block_locator` puts<br>the wallet's real tip hash at the head of the locator, `find_fork_point` matches immediately,<br>`locate_fork_point` returns `prev_tip` unchanged, and `truncate_to_height` at that height is a no-op<br>that returns `Ok(h)` rather than erroring — converting a crash loop into a tight spin.

## A second, independent defect

`WalletWrite::truncate_to_height` returns the height it **actually** truncated to — its own docs say<br>*"Implementations of this method may choose a lower block height … this method returns the height to**<br>**which the data store was actually truncated."* `select_truncation_height` snaps the request down to<br>the highest height that is a checkpoint in Sapling **and** Orchard **and** Ironwood, and checkpoints<br>prune to `PRUNING_DEPTH = 100` plus anchors every 288 blocks.

Zallet **discards that return value** at both call sites in `sync.rs`, then sets<br>`prev_tip = fork_point` and resumes streaming from `fork_point + 1`. For any rollback deeper than<br>~100 blocks the wallet can land far below the fork point, leaving a gap and applying blocks onto a<br>stale frontier — an independent route to the same divergence. (The block stream is also constructed<br>*before* the truncate, so honoring the real height requires reordering.)

## Why this is not zcash/zallet#126

zcash/zallet#126 was closed on 2026-08-04 — two days before this report — citing [#335](<https://github.com/zcash/zallet/issues/335>), [#512](<https://github.com/zcash/zallet/issues/512>), zcash/zallet#560, zcash/zallet#561, zcash/zallet#576.<br>**All five are ancestors of** `v0.1.0-beta.2` (`git merge-base --is-ancestor`), so those fixes were<br>already in the build that failed. This is either a regression or a distinct variant that the testnet<br>reproduction in that thread did not cover (that run used a fresh wallet; this one had existing state).

Related open issues: zcash/zallet#559 (zebra backend reports a stale tip), zcash/zallet#587, zcash/zallet#588, zcash/zallet#491.

## What we need from the reporter

The workaround destroys the evidence, so please capture these **before** running `repair truncate-wallet` again:

1. **A copy of** `wallet.db` **from before the truncate**, if one still exists. This settles it outright.
2. 
   ```sql
   SELECT height, ironwood_commitment_tree_size FROM blocks
   WHERE height BETWEEN 3437900 AND 3438010;
   ```

   A `0` or `NULL` at a height above 3,428,143 confirms the empty-frontier hypothesis directly.
3. **zebrad version and on-disk state format version**, and whether the zebrad log shows a<br>**v27→v28** `add_ironwood_tree` **format upgrade** (it runs on a background thread via<br>`spawn_format_change`) at any point before the failure.
4. Anything in the zebrad log about Ironwood tree reads, or a reorg, around 2026-08-06 04:43 UTC.
5. Whether the wallet was syncing from a birthday below NU6.3 activation, or had been running<br>continuously across it.
6. A re-run with `RUST_LOG=shardtree=warn`: `shardtree`'s `insert_subtree` emits<br>`warn!(cur_root, new_root, "Insertion conflict")`, which tells us *which side* is wrong.
7. The second block hash from the report, re-pasted — as given<br>(`000000000052a1bc584c034ef2ae91f634cbe77c049e5a1b927aca637b1c1ba11`) it is **65 hex characters**,<br>one too many, and neither 64-char variant resolves on a block explorer. If it is a genuine orphan<br>that would point at a reorg instead.

## Proposed fixes

1. **Guard the empty-tree fallback on pool activation** (zebra backend; same for Sapling/Orchard, and<br>the Zaino backend has the identical arm). Below activation an absent tree legitimately means the<br>empty tree; at or above it, return `ChainError::Unavailable` — which `is_retryable` already treats<br>as transient — so the wallet stalls loudly instead of corrupting itself. *This is the only change**<br>**that prevents new corruption; everything else is remediation.*
2. **Honor** `truncate_to_height`**'s returned height** at both `sync.rs` call sites, and build the block<br>stream after the rewind rather than before.
3. **Bounded escalation for tree divergence**: classify `PutBlocksCommitmentTree` /<br>`CommitmentTree` / `TruncateCommitmentTree`, roll back to progressively older checkpoints<br>(floored at the wallet birthday, capped in attempts, with a high-water mark so a<br>rollback/rescan cycle cannot spin), and on exhaustion fail with a message naming<br>`zallet repair truncate-wallet <height>`. `recover_history` needs bounded retry too — it<br>currently has no error handling at all.
4. **Narrow** `initialize`**'s best-effort tip scan**, which swallows *all* errors and then returns the<br>chain tip as `steady_state`'s starting `prev_tip` regardless — letting `steady_state` start from a<br>block the wallet never stored, which the next tip change misreads as a reorg.

`check_witnesses` is *not* a usable detector here: it only builds witnesses for notes the wallet<br>believes spendable, and positions 29280..29287 are almost certainly chain data the wallet holds no<br>note in — it would report the wallet healthy while it crash-loops.

pacu:
Closes #743

A user running v0.1.0-beta.2 with the zebra backend on mainnet hit
`PutBlocksCommitmentTree { pool: Ironwood, block_range: 3438008..3438010, error:
Insert(Conflict(Address { level: Level(3), index: 3660 })) }` and crash-looped on every
restart until they rewound the wallet by hand. #743 has the full decode; the short version is
that the address decodes to Ironwood note positions 29280..29287 and to the level-3 ommer of
the frontier at 3438008, so the wallet's tree was already wrong *before* the block it died on.

The core finding is that **nothing validates `from_state`**. `put_blocks` looks like it does —
it checks `from_state.tree_size() + commitments.len() == block.final_tree_size()` — but in
Zallet's usage both sides derive from the same value: `sync::steps` builds `BlockMetadata` from
`from_state`, and `scanning::full::scan_block` derives `*_final_tree_size` from that metadata.
The check is circular, so a wrong frontier of any size is committed to the wallet's shardtree
and only detonates later, when a correct frontier contradicts it.

## The commits

Each is independently reviewable and independently revertable.

**`5e0738e` — zebra: reject an empty frontier for an unreadable post-activation tree**

`tree_state_as_of` substituted `CommitmentTree::empty()` whenever the node reported no
treestate for a finalized block, for every pool. Below a pool's activation height that is
correct; at or above it, it silently corrupts the wallet.

zebra-state does guard this — `ironwood_tree_by_height` treats a missing post-NU6.3 tree as an
invariant violation rather than masking it — but only on the by-height path. We read by hash,
and `ironwood_tree_by_hash_or_height` resolves hash → height first and returns `None` if that
fails, short-circuiting before the check runs.

Adds `TreePool` / `empty_tree_is_legitimate` to zallet-core so both backends share one
definition, and documents the invariant on `ChainView::tree_state_as_of`. Failed reads at or
above activation now return `ChainError::Unavailable`, which `is_retryable` already treats as
transient, so sync stalls loudly instead of corrupting.

**`73fb61c` — zaino: the same guard.** Not implicated in any report, but it had the same
fallback in a wider form (no finalized/non-finalized distinction at all).

**`d09efc3` — sync: resume from the height the wallet actually rewound to**

`truncate_to_height` documents that it "may choose a lower block height … returns the height to
which the data store was actually truncated". Both call sites discarded that and set `prev_tip`
to the requested fork point, so a rollback deeper than the per-block checkpoint window could
leave a gap and apply later blocks onto a stale frontier — an independent route to the same
divergence. In `steady_state_iteration` the block stream was also built *before* the rewind, so
this reorders it.

**`e497c9a` — sync: recover from tree divergence instead of exiting**

Routing these into the existing `BlockConflict` arm would be *worse* than the crash loop: on the
canonical chain the locator head still matches, `find_fork_point` returns the wallet's own tip,
and truncating there is a no-op returning `Ok`, giving a tight retry loop instead of a clean
exit. So recovery gets its own driver: a fixed ladder of rollback depths, floored at the wallet
birthday and capped in attempts, failing with a `SyncError::WalletTreeDiverged` that names a
`repair truncate-wallet` height range.

The subtle part is `TreeRecovery::high_water`. Attempts reset only once the wallet climbs back
*above* where the divergence was first seen — resetting on any successful scan would let "roll
back one, rescan one, conflict again" repeat forever, since each rescan looks like progress and
returns the ladder to its first rung. There's a test pinning exactly that.

Also: `recover_history` gets bounded retry (it can raise the identical error and had no error
handling at all; it deliberately does *not* roll back itself, since it holds a separate DB
connection and would race `steady_state`), and `initialize`'s best-effort tip scan no longer
swallows non-retryable errors — it was handing `steady_state` a tip the wallet never stored,
which the next tip change misreads as a reorg.

**`86912e2` — repair: add `check-witnesses`**

Exposes `WalletDb::check_witnesses` as an operator tool. Deliberately **not** wired into the
recovery path: it only builds witnesses for notes the wallet holds, so a tree that disagrees
with the chain's where the wallet has no note of its own passes it cleanly — it would have
called this wallet healthy while it crash-looped. The CLI help and book page both say so.

## Testing

`zallet-core` 332 tests, zebra 8 unit + 9 acceptance, zaino 2 + 2. `cargo clippy --all-targets`
and `cargo fmt --check` clean in all three workspaces. New coverage: per-pool activation
boundaries (including Ironwood at the reported mainnet height 3,438,008), the zebra chain
view's pre/post-activation and non-finalized paths, `resume_point`, and the recovery ladder and
its anti-spin invariant.

The zebra chain-view tests use an all-`None` mock reader, and regtest activation heights carry
backward, so the guard trips on Sapling first there — the per-pool coverage lives in
zallet-core alongside the helper. That's called out in a comment rather than dressed up as
something more targeted than it is.

## What still needs checking

**This has not been run against mainnet.** The guard turns a silent-corruption path into a retry
stall, so if any backend legitimately returns no Ironwood tree for a post-activation *finalized*
block, those users would see sync stall rather than proceed. I don't expect it — zebra-state
treats that as an invariant violation itself — but it wants confirming against a synced mainnet
node past height 3,428,143 before this ships.

**The root cause is the best-supported hypothesis, not proven.** Confirming it needs the
reporter's pre-truncate `wallet.db` or their `ironwood_commitment_tree_size` values; #743 asks
for both. The main alternative is an undetected reorg leaving stale Ironwood leaves. These
changes are correct under either, which is why they're proposed without waiting for the answer.

Note this does **not** repair a wallet that has already diverged — those still need
`zallet repair truncate-wallet`. It prevents new corruption and stops the crash loop from being
permanent.

Unrelated to #587: that's the wallet failing to notice a *shrinking* tip after `invalidateblock`,
and nothing here unblocks `wallet_ironwood_reorg.py`.

Fixes #743

https://claude.ai/code/session_011NTkERgJYrU1oA7zKGDtPY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,26 @@ be considered breaking changes.
 
 ### Fixed
 
+- Fixed a wallet corruption that made sync fail permanently with a note
+  commitment tree conflict, reported as
+  `PutBlocksCommitmentTree { .. Insert(Conflict(..)) }`. Once a wallet reached
+  this state it exited on every start and could only be recovered by rewinding
+  it manually with `zallet repair truncate-wallet`.
+
+  The cause was a chain backend reporting no note commitment tree for a block at
+  or after a shielded pool's activation height, which Zallet accepted as an
+  empty tree and stored. Nothing detected the bad data at the time it was
+  written, so the wallet's tree silently diverged from the chain's and only
+  failed later, when a correct tree state disagreed with what had been stored —
+  potentially thousands of blocks after the damage was done. Backends now report
+  such reads as a temporary failure and sync retries, rather than storing a
+  placeholder. This was most likely to affect `zallet-zebra` users whose wallets
+  crossed the NU6.3 activation height, where the Ironwood pool's tree is new.
+
+  Wallets already in this state still need `zallet repair truncate-wallet` to
+  recover; this release prevents the corruption from occurring, and does not
+  repair a wallet that has already diverged.
+
 - `init-wallet-encryption` now canonicalizes the age recipients derived from
   the identity file before storing them, and validates that the resulting set
   is non-empty and parseable. Previously every line of the internally generated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,11 @@ be considered breaking changes.
   (default 100).
 - `getwalletstatus` now reports a `locked` field indicating whether the wallet
   is currently usable.
+- The `zallet repair check-witnesses` command, which reports the block ranges
+  that must be rescanned to rebuild missing witness data for the wallet's
+  spendable notes, and can queue them for rescanning with `--queue-rescan`.
+  Note it only examines notes the wallet holds, so a clean result does not prove
+  the wallet's note commitment trees agree with the chain's.
 - `migrate-zcashd-wallet --allow-partial-import` flag, permitting a migration
   to complete even when some accounts or transparent spending keys in the
   `zcashd` wallet could not be imported. The skipped items are reported as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,17 @@ be considered breaking changes.
 
 ### Fixed
 
+- A note commitment tree conflict during sync no longer shuts the wallet down
+  permanently. Zallet now rolls the wallet back to a progressively older point
+  and rescans, up to a bounded number of attempts and never below the wallet's
+  birthday. If that does not resolve it, Zallet still exits, but now says so
+  explicitly and suggests the `zallet repair truncate-wallet` height range to
+  try — rather than repeating the same conflict on every start with no
+  indication of what to do about it.
+
+  While recovering, the wallet is in its recovering state and balance and spend
+  methods are unavailable, as during a reorganization.
+
 - After a chain reorganization, the wallet now resumes scanning from the height
   it actually rewound to, rather than from the fork point it asked for. Rewinding
   the wallet can only stop at a height that is a checkpoint in every shielded
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,18 @@ be considered breaking changes.
 
 ### Fixed
 
+- After a chain reorganization, the wallet now resumes scanning from the height
+  it actually rewound to, rather than from the fork point it asked for. Rewinding
+  the wallet can only stop at a height that is a checkpoint in every shielded
+  pool's note commitment tree, and older checkpoints are pruned, so a rollback
+  deeper than a hundred or so blocks could land well below the fork point.
+  Zallet then resumed just above the fork point regardless, skipping the blocks
+  in between and applying the rest onto a stale tree state — silently diverging
+  the wallet's note commitment tree from the chain's, which surfaced later as an
+  unrecoverable conflict. Deep reorganizations will now rescan somewhat more
+  than before, and the wallet stays in its recovering state (during which
+  balance and spend methods are unavailable) for correspondingly longer.
+
 - Fixed a wallet corruption that made sync fail permanently with a note
   commitment tree conflict, reported as
   `PutBlocksCommitmentTree { .. Insert(Conflict(..)) }`. Once a wallet reached
--- a/backends/zaino/CHANGELOG.md
+++ b/backends/zaino/CHANGELOG.md
@@ -20,6 +20,20 @@ should be considered breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- The chain view no longer substitutes an empty note commitment tree when the
+  validator reports no treestate for a pool at or after that pool's activation
+  height. Previously any such read — for Sapling, Orchard, or Ironwood — was
+  treated as "this pool is not active yet, so its tree is empty", which stored a
+  placeholder frontier that nothing validates and that only surfaced later, as
+  an unrecoverable `shardtree` conflict, once a correct frontier disagreed with
+  it. Such reads are now reported as a transient chain error so sync retries.
+  Reads below a pool's activation height are unaffected.
+
+  This backend was not implicated in the reports that prompted the fix, which
+  were all on `zallet-zebra`; the same unguarded fallback existed here.
+
 ## [0.1.0-beta.2] - 2026-07-28
 
 ### Changed
--- a/backends/zaino/src/chain.rs
+++ b/backends/zaino/src/chain.rs
@@ -55,7 +55,7 @@ use crate::read_state::{AbortOnDrop, init_read_state_service, network_to_zebra};
 use zallet_core::components::chain::SpendStatus;
 use zallet_core::components::chain::{
     BlockLocator, Chain, ChainBlock, ChainError, ChainFactory, ChainTx, ChainView, ReportedUpgrade,
-    UpgradeStatus,
+    TreePool, UpgradeStatus, empty_tree_is_legitimate,
 };
 
 /// Classifies a block-fetch error, distinguishing transient reorg-window failures from
@@ -477,8 +477,26 @@ impl ChainView for ZainoChainView {
                 .await
                 .map_err(ChainError::backend)?;
 
+            // The validator reports a pool's treestate only from that pool's activation
+            // height onward; before then it is `None` and the final tree is empty. At or
+            // after activation, `None` means the validator could not serve it. Substituting
+            // an empty frontier there silently corrupts the wallet's note commitment tree
+            // (see `ChainView::tree_state_as_of`): nothing validates the frontier, so the
+            // bad data is stored and only surfaces later, potentially thousands of blocks
+            // on, as an unrecoverable `shardtree` conflict. Report it as unavailable so the
+            // sync engine retries instead.
+            let missing_tree = |pool: TreePool| -> ChainError {
+                ChainError::unavailable(format!(
+                    "the validator reported no {pool} treestate at height {height}, at or \
+                     after that pool's activation; refusing to substitute an empty frontier"
+                ))
+            };
+            let absent_tree_is_empty =
+                |pool: TreePool| empty_tree_is_legitimate(&self.params, pool, height);
+
             let final_sapling_tree = match sapling_treestate {
-                None => CommitmentTree::empty(),
+                None if absent_tree_is_empty(TreePool::Sapling) => CommitmentTree::empty(),
+                None => return Err(missing_tree(TreePool::Sapling)),
                 Some(sapling_treestate) => read_commitment_tree::<
                     sapling::Node,
                     _,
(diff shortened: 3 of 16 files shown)


schell:
@pacu - I rebased on `main` for you, then removed two redundant `begin_recovery` calls. Both of them reorg recovery paths (`steady_state` `BlockConflict` arm and `steady_state_iteration` reorg arm). Each called `status.begin_recovery(fork_point.height()) `then immediately called `truncate_to_wallet_checkpoint` and `status.begin_recovery(resume_from.height())`. The first call was always overwritten before any reader could observe it (no .await between them), so it was dead.

Everything else looks correct.

schell (approved):
ACK [6c9d090](https://github.com/zcash/zallet/pull/744/commits/6c9d0909d9ecef9815612c69320f31d58034ebef)

PR #746, opened 2026-08-08
Bump serde_with to 3.21.0 in zaino backend lockfile

### LINKED ISSUE #745: Dependabot alert 36: serde_with < 3.21.0 KeyValueMap serialization panic (zaino backend)
[Dependabot alert #36](https://github.com/zcash/zallet/security/dependabot/36) flags `serde_with` 3.17.0 in `backends/zaino/Cargo.lock` for [GHSA-7gcf-g7xr-8hxj](https://github.com/advisories/GHSA-7gcf-g7xr-8hxj) (moderate, CVSS v4 5.1): serializing an empty inner sequence or map entry through `#[serde_as(as = "KeyValueMap<_>")]` panics, because the serializer preallocates `Vec::with_capacity(len - 1)` before validating that the entry has a first field to use as the map key. A process serializing attacker-influenced data through `KeyValueMap` can be crashed (denial of service).

In Zallet the crate arrives transitively via `zebra-chain` and `zebra-rpc` in the zaino backend workspace. The fix is a semver-compatible lockfile bump to the first patched version, 3.21.0. The zebra backend and root workspace lockfiles already resolve to non-vulnerable versions on `main`.

pacu:
Closes #745. Resolves [Dependabot alert #36](https://github.com/zcash/zallet/security/dependabot/36) ([GHSA-7gcf-g7xr-8hxj](https://github.com/advisories/GHSA-7gcf-g7xr-8hxj)).

## Motivation

`serde_with` < 3.21.0 panics when serializing an empty inner sequence or map entry through `#[serde_as(as = "KeyValueMap<_>")]` (it preallocates `Vec::with_capacity(len - 1)` before validating the entry), which an attacker able to influence serialized data could use for denial of service. `backends/zaino/Cargo.lock` pinned the vulnerable 3.17.0, pulled in transitively via `zebra-chain` and `zebra-rpc`.

## Solution

`cargo update serde_with --precise 3.21.0 --manifest-path backends/zaino/Cargo.toml` — a semver-compatible, lockfile-only bump (serde_with 3.17.0 → 3.21.0, serde_with_macros 3.17.0 → 3.21.0, and its build-time `darling` dependency 0.21.3 → 0.23.0). The zebra backend and root workspace lockfiles already resolve to non-vulnerable versions on `main`, so no changes there.

## Test evidence

- `./utils/check-lockstep.sh`: Lockstep OK, 20 crates + 4 package versions consistent across the 3 lockfiles.
- `cargo check --locked --manifest-path backends/zaino/Cargo.toml --all-targets` passes.

### DIFF
--- a/backends/zaino/supply-chain/config.toml
+++ b/backends/zaino/supply-chain/config.toml
@@ -449,15 +449,15 @@ version = "0.8.1"
 criteria = "safe-to-deploy"
 
 [[exemptions.darling]]
-version = "0.13.4"
+version = "0.23.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.darling_core]]
-version = "0.13.4"
+version = "0.23.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.darling_macro]]
-version = "0.20.11"
+version = "0.23.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.dashmap]]
@@ -1569,11 +1569,11 @@ version = "0.7.1"
 criteria = "safe-to-deploy"
 
 [[exemptions.serde_with]]
-version = "3.17.0"
+version = "3.21.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.serde_with_macros]]
-version = "3.17.0"
+version = "3.21.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.serdect]]
--- a/backends/zaino/Cargo.lock
+++ b/backends/zaino/Cargo.lock
@@ -1187,9 +1187,9 @@ dependencies = [
 
 [[package]]
 name = "darling"
-version = "0.21.3"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9cdf337090841a411e2a7f3deb9187445851f91b309c0c0a29e05f74a00a48c0"
+checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
 dependencies = [
  "darling_core",
  "darling_macro",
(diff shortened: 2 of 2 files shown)


backends/zaino/supply-chain/config.toml
```diff
@@ -449,15 +449,15 @@ version = "0.8.1"
 criteria = "safe-to-deploy"
 
 [[exemptions.darling]]
-version = "0.13.4"
+version = "0.23.0"
```
schell, line 452: Claude says:
> The darling exemption was bumped from 0.13.4 to 0.23.0, but the old lockfile had darling 0.21.3, not 0.13.4. This is a pre-existing mismatch in the vet config (the exemption version didn't track the lockfile version). The new exemption now matches. Worth confirming the old 0.13.4 entry was stale and not covering a different darling version elsewhere in the graph.

pacu: Confirmed, with one correction to the framing: the old `0.13.4` entry was not stale — it was the **base of the imported Mozilla delta-audit chain** (`0.13.4 → 0.14.2 → 0.14.3 → 0.20.1 → 0.20.10 → 0.21.3` in `imports.lock`), which is what certified the previous lockfile's single darling cohort at 0.21.3. Exact-version exemptions often anchor delta chains like this.

Checked both lockfiles: there is exactly one darling version in the graph (all three crates at 0.21.3 before, 0.23.0 after), so the entry covered nothing else.

The imported chain ends at 0.21.3 and no audit covers `0.21.3 → 0.23.0`, so regenerating exemptions had to exempt 0.23.0 outright. Net effect worth noting: darling moves from audited-via-delta-chain to raw exemption. If we want audit-grade coverage back we can self-audit the `0.21.3 → 0.23.0` delta or wait for the Mozilla imports to catch up. 🤖

schell (approved):

schell (approved):
utACK 67abf30336

PR #750, opened 2026-08-08
rpc: Require explicit opt-in to serve plaintext JSON-RPC beyond loopback

pacu:
Closes #749.

> **Stacked PR** on #748 (`feature/cor-1656`); only the last commit is new here. Merge #748 first, then retarget this to `main`.

## Motivation

The JSON-RPC interface is served over plaintext HTTP: HTTP Basic credentials and request bodies (including the passphrase supplied to `walletpassphrase`, and spending keys) are not encrypted in transit, and a captured Basic header is replayable. Zallet nevertheless accepted any `rpc.bind` address without a runtime warning or opt-in.

Fixes the finding tracked as:
- Least Authority ZCG Zallet Initial Audit Report (24 Jul 2026), Issue N (Medium)
- Linear: [COR-1661](https://linear.app/zodl/issue/COR-1661)

## Solution

Implements the audit's stronger remediation (reject by default + explicit unsafe opt-in + runtime warning + docs):

- `zallet start` now refuses to open the JSON-RPC endpoint on a non-loopback `rpc.bind` address, with an error explaining the risk and the secure alternative (SSH port forwarding or VPN to a loopback bind).
- New `rpc.allow_insecure_remote_bind` config option (default `false`) explicitly opts in to plaintext non-loopback serving; a prominent `SECURITY WARNING` is logged at startup while it is enabled.
- The decision is a pure `BindPolicy` function on the bind address + opt-in flag.
- Docs: expanded `rpc.bind` config rustdoc (feeds `example-config` and the book's config reference; golden regenerated) and a security note with the SSH-tunnel pattern in the book's setup guide.

## Test evidence

- New proptests: any loopback bind (v4/v6, any port) is always permitted; any non-loopback `SocketAddr` is refused without the opt-in and warned-and-allowed with it.
- `TRYCMD` golden `example_config.out/zallet.toml` regenerated and `cli_tests` passes.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #752, opened 2026-08-08
cli: Write to the default config path when --output is omitted

### LINKED ISSUE #751: example-config and migrate-zcash-conf panic (todo!) when --output is omitted
Tracking issue for Least Authority *ZCG Zallet Initial Audit Report* (24 Jul 2026), **Suggestion 9: Handle the Documented Default Output Path Without Panicking**.

The documentation for the `example-config` and `migrate-zcash-conf` commands states that omitting the `-o/--output` flag uses the default Zallet configuration path, but both `ExampleConfigCmd` and `MigrateZcashConfCmd` invoke `todo!()` in that branch, so invoking either command without `--output` panics.

Linear: COR-1671.

pacu:
Closes #751.

> **Stacked PR** on #750 (`feature/cor-1661`), which stacks on #748; only the last commit is new here. Merge the stack bottom-up and retarget.

## Motivation

The documentation for `example-config` and `migrate-zcash-conf` states that omitting `-o/--output` uses the default Zallet config file path, but both commands invoked `todo!()` in that branch, so they panicked instead.

Fixes the finding tracked as:
- Least Authority ZCG Zallet Initial Audit Report (24 Jul 2026), Suggestion 9
- Linear: [COR-1671](https://linear.app/zodl/issue/COR-1671)

## Solution

Add a `resolve_output_target` helper next to `resolve_config_path`, resolving the omitted flag to the same path `zallet` loads its configuration from at startup (`zallet.toml` in the datadir, honouring `-d/--datadir`), `-` to stdout, and any other value as given. Both commands use it; if the default path already exists, `File::create_new` returns a descriptive error unless `--force` is passed (unchanged behaviour for explicit paths).

## Test evidence

- New unit tests for `resolve_output_target` (default path, `-`, explicit relative/absolute paths).
- `cargo test -p zallet-core --features zcashd-import`, `cargo clippy --features zcashd-import --all-targets -- -D warnings`, and `cargo fmt --check` are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,13 @@ be considered breaking changes.
 - A regtest wallet's transactions mined under NU6.3 (Ironwood) no longer fail
   to import with "Consensus branch ID not known". Bumped `zewif-zcashd` to
   0.1.0-rc.5, which includes NU6.3 in the regtest activation schedule.
+- `example-config` and `migrate-zcash-conf` no longer panic when `-o/--output`
+  is omitted. As their documentation already stated, they now write to the
+  default Zallet config file path (`zallet.toml` in the datadir), matching the
+  path `zallet` loads its configuration from at startup. `--force` overwrites
+  an existing file only when the output is named explicitly with `-o`; the
+  inferred default path — the wallet's live configuration — is never
+  overwritten.
 - On Windows, the RPC authentication cookie is now created with an explicit
   owner-only access policy (a protected DACL granting access solely to the
   user Zallet runs as), instead of inheriting the ACL of the data directory
--- a/zallet-core/i18n/en-US/zallet_core.ftl
+++ b/zallet-core/i18n/en-US/zallet_core.ftl
@@ -213,6 +213,9 @@ warn-init-rpc-bind-insecure-remote =
     the network path. Prefer a loopback bind plus an authenticated, encrypted tunnel
     (such as SSH port forwarding or a VPN).
 err-config-file-not-found = Configuration file at {$path} does not exist.
+err-config-output-exists =
+    Refusing to overwrite the existing config file at {$path}. To replace it, name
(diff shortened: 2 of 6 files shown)


schell (changes requested):
So there's a silent overwrite risk on `--force`. THere's two cases:

1. If the user runs zallet example-config --force without -o, it writes to <datadir>/zallet.toml, overwriting their existing config. Without --force, File::create_new will refuse if the file exists, so the default case is safe. But --force plus no -o is a footgun: the user might expect --force to only matter when an explicit -o target already exists, but it also silently clobbers the running config.
2. Same issue but worse for migrate-zcash-conf: if a user already migrated and then re-runs migrate-zcash-conf --force without -o, it overwrites their tuned zallet.toml with a fresh migration from their zcash.conf. Any manual edits to the Zallet config are lost.

I think `--force` should only apply when `-o` is explicitly provided. When the output path is inferred (no `-o` flag), refuse to overwrite even if `--force` is set. This forces the user to be explicit about clobbering their own config.

pacu:
Implemented exactly as you describe: `--force` now consents to overwriting only a target named explicitly with `-o` (`overwrite_allowed(force, output.is_some())` in `commands.rs`, shared by both commands). When the path is inferred, an existing file is never overwritten — `--force` or not — and instead of a bare `File exists` I/O error the command now fails with:

> Refusing to overwrite the existing config file at {path}. To replace it, name the target explicitly: '-o {path} --force'.

Also updated both `--force` flag docs, the changelog entry, and added a unit test pinning the four force/explicit combinations. 🤖

schell (dismissed):
utACK [edf85e5](https://github.com/zcash/zallet/pull/752/commits/edf85e5af0dfb87440565e1748135beaba202e1d)

schell (dismissed):
utACK d475c3a155

schell (approved):

PR #754, opened 2026-08-08
rpc: Enforce a maximum size when buffering JSON-RPC request bodies

### LINKED ISSUE #753: RPC compatibility middleware buffers the entire request body with no size limit
Tracking issue for a finding reported in both:

- GHSA-m539-482m-q66j ("RPC middleware buffers the entire request body with no size limit", Low)
- Least Authority *ZCG Zallet Initial Audit Report* (24 Jul 2026), **Suggestion 8: Enforce a Maximum Size When Buffering JSON-RPC Requests**

`request_to_json_rpc_2` (`zallet-core/src/components/json_rpc/server/http_request_compatibility.rs`) buffers the complete HTTP request body into memory with `body.collect()` and no cap. jsonrpsee's own `max_request_body_size` only applies after this middleware has already buffered, so an authenticated caller can exhaust wallet memory with an oversized body. (Unauthenticated requests are rejected by the outer `AuthorizationLayer` before buffering.)

Linear: COR-1670.

pacu:
Closes #753.

> **Stacked PR** on #752 (`feature/cor-1671`) → #750 → #748; only the last commit is new here. Merge the stack bottom-up and retarget.

## Motivation

The HTTP compatibility middleware buffered the complete request body into memory (`body.collect()`) before jsonrpsee's own `max_request_body_size` could apply, so an authenticated caller could exhaust wallet memory with a single oversized request.

Fixes the finding tracked as:
- GHSA-m539-482m-q66j ("RPC middleware buffers the entire request body with no size limit")
- Least Authority ZCG Zallet Initial Audit Report (24 Jul 2026), Suggestion 8
- Linear: [COR-1670](https://linear.app/zodl/issue/COR-1670)

## Solution

Wrap the body in `http_body_util::Limited` before collecting, capped at `jsonrpsee::core::TEN_MB_SIZE_BYTES` — the same server-default limit jsonrpsee applies to the requests it parses, reused rather than re-derived. Over-limit bodies get `413 Payload Too Large`; a body that cannot be read gets `400 Bad Request` (previously an `expect()` panicked the connection task).

## Test evidence

- New unit tests: a body one byte over the limit is rejected with 413 without being buffered; a body exactly at the limit is accepted; a small Bitcoin-flavoured request is still mapped to JSON-RPC 2.0.
- `cargo test -p zallet-core`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` clean.

### DIFF
--- a/zallet-core/src/components/json_rpc/server/http_request_compatibility.rs
+++ b/zallet-core/src/components/json_rpc/server/http_request_compatibility.rs
@@ -6,10 +6,10 @@ use std::future::Future;
 use std::pin::Pin;
 
 use futures::FutureExt;
-use http_body_util::BodyExt;
+use http_body_util::{BodyExt, LengthLimitError, Limited};
 use hyper::{StatusCode, header};
 use jsonrpsee::{
-    core::BoxError,
+    core::{BoxError, TEN_MB_SIZE_BYTES},
     server::{HttpBody, HttpRequest, HttpResponse},
     types::{ErrorCode, ErrorObject},
 };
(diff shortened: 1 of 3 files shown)


zallet-core/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -96,16 +96,42 @@ impl<S> HttpRequestMiddleware<S> {
...
 
+    /// The maximum HTTP request body size this middleware will buffer.
+    ///
+    /// This middleware buffers the complete body before `jsonrpsee` parses the
+    /// request, so `jsonrpsee`'s own limit cannot protect the buffering here; an
+    /// authenticated caller could otherwise stream an arbitrarily large body into
+    /// memory. Reuse the same limit `jsonrpsee` applies to the requests it parses
+    /// (its server default), so the middleware never buffers a request that the
+    /// server would reject anyway.
+    const MAX_REQUEST_BODY_SIZE: u32 = TEN_MB_SIZE_BYTES;
```
schell: Non-blocking: this `const` is compile-time, but the jsonrpsee server's `max_request_body_size` is runtime (defaulting to `TEN_MB_SIZE_BYTES`). They match today because `server.rs` uses `Server::builder()` without `.set_max_request_body_size(...)`. If someone later changes the server limit, the middleware won't track it. Consider threading the configured value in at construction time, or add a comment at the call site noting the coupling.

zallet-core/src/components/json_rpc/server/http_request_compatibility.rs
```diff
@@ -275,6 +304,59 @@ impl JsonRpcRequest {
...
+            .expect("valid request")
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn oversized_body_is_rejected_not_buffered() {
+        let body = vec![b'0'; Middleware::MAX_REQUEST_BODY_SIZE as usize + 1];
+        let response = Middleware::request_to_json_rpc_2(request_with_body(body))
+            .await
+            .expect_err("a body over the limit must be rejected");
+        assert_eq!(response.status(), StatusCode::PAYLOAD_TOO_LARGE);
```
schell, line 331: Non-blocking: there's a test for the 413 path and the success path, but no test for the 400 Bad Request branch (the `Err` arm where `e.is::<LengthLimitError>()` is false, a non-size read failure). Consider a test with a synthetic failing body type.

zallet-core/src/components/json_rpc/server/http_request_compatibility.rs
pacu: Good catch — verified, and fixed. Rather than threading the runtime value into the middleware, I inverted the coupling so there is one value: `MAX_REQUEST_BODY_SIZE` is now `pub(super)`, and `server.rs` configures the server from it:

```rust
let server_instance = Server::builder()
    .http_only()
    // Keep the server's parse limit equal to what the compatibility middleware
    // will buffer, so neither can drift into accepting a body the other rejects.
    .max_request_body_size(
        http_request_compatibility::HttpRequestMiddleware::<()>::MAX_REQUEST_BODY_SIZE,
    )
```

They can no longer disagree — changing the limit means changing one constant. The doc comment now states the invariant (the middleware's limit must not exceed the server's, or it buffers bodies the server then rejects, which is the exhaustion this exists to prevent) rather than relying on both happening to equal `TEN_MB_SIZE_BYTES`.

zallet-core/src/components/json_rpc/server/http_request_compatibility.rs
pacu: Added `unreadable_body_is_a_bad_request`: a `StreamBody` whose single item is an `Err`, so the read fails without ever reaching the length limit and takes the `else` arm. Asserts `400`, which distinguishes it from the `413` case — if it ever regressed into being treated as a size error the assert would catch it.

All four middleware tests pass.

schell (dismissed):
utACK 9e4b9a2

schell (dismissed):
utACK 2cec7b70b4

schell (approved):

PR #756, opened 2026-08-08
rpc: Restrict spend input selection to the source address's receiver pools

### LINKED ISSUE #755: z_sendmany selects inputs from all shielded pools regardless of fromaddress receivers
Tracking issue for Least Authority *ZCG Zallet Initial Audit Report* (24 Jul 2026), **Suggestion 1: Restrict Spend Input Selection to Receiver Pools for the Source Address**.

`z_sendmany` documents that a receiver-restricted Unified Address used as `fromaddress` selects inputs only from the value pools its receivers include, but `spend_policy_for` maps every non-transparent source to `SpendPolicy::default()`, which permits all shielded pools. An Orchard-only address can therefore spend the account's Sapling notes. Documented-contract/user-expectation mismatch (selection is account-scoped and signed with the caller's own key, so no third-party spend or fund loss).

Linear: COR-1663.

pacu:
Closes #755.

> **Stacked PR** on #754 (`feature/cor-1670`) → #752 → #750 → #748; only the last commit is new here. Merge the stack bottom-up and retarget.

## Motivation

`z_sendmany` documents that a receiver-restricted unified address used as `fromaddress` selects inputs only from the value pools its receivers include, but `spend_policy_for` mapped every non-transparent source to `SpendPolicy::default()` (all shielded pools), so an Orchard-only address could spend the account's Sapling notes.

Fixes the finding tracked as:
- Least Authority ZCG Zallet Initial Audit Report (24 Jul 2026), Suggestion 1
- Linear: [COR-1663](https://linear.app/zodl/issue/COR-1663)

## Solution

Derive the `SpendPolicy` from the source address's receivers (the audit's first suggested option), for both `z_sendmany` and `pczt_create` which share `spend_policy_for`:

- Bare Sapling address / Sapling-only UA → Sapling pool only.
- Orchard receiver → Orchard **and Ironwood** pools: once Ironwood is active, payments to Orchard receivers are accounted to the Ironwood bundle, so an Orchard-receiver source must draw on both — restricting to Orchard alone would strand migrated funds.
- A UA's transparent receiver still permits no transparent spending (unchanged), and a bare transparent source is unchanged.
- A TEX source now permits nothing (it cannot correspond to a wallet account; account resolution rejects it first).

## Test evidence

- New proptests: a bare Sapling source permits exactly `{Sapling}`; receiver-restricted UAs (Orchard-only, Sapling-only, both — with and without a transparent receiver) permit exactly the pools their receivers name; existing properties (no transparent spending from shielded sources, transparent-source confinement, change policy) still pass.
- `cargo test -p zallet-core`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` clean.

## Integration tests

The stack-level cross-pool coverage lives in zcash/integration-tests#193, which has been merged; this PR's CI now exercises it from the integration-tests default branch.

### DIFF
--- a/zallet-core/src/components/json_rpc/payments.rs
+++ b/zallet-core/src/components/json_rpc/payments.rs
@@ -161,6 +161,15 @@ pub(super) fn confirmations_policy_for_minconf(
 /// account's other transparent receivers. Every other source stays shielded-only, so a
 /// shielded send can never silently reach into transparent funds.
 ///
+/// A shielded source selects inputs only from the value pools its receivers name, as the
+/// `z_sendmany` documentation promises: a bare Sapling address or a Sapling-only unified
+/// address does not reach into the account's Orchard notes, and an Orchard-only unified
+/// address does not reach into its Sapling notes. An Orchard receiver corresponds to both
+/// the Orchard pool and the Ironwood pool: once Ironwood is active, payments to Orchard
+/// receivers are accounted to the Ironwood bundle, so an Orchard-receiver source must be
+/// able to draw on both. A unified address's transparent receiver deliberately does *not*
+/// permit transparent spending (see above).
+///
 /// Coinbase UTXOs are excluded: `TransparentSpendPolicy` defaults to
 /// `CoinbasePolicy::NonCoinbase`, and consensus requires coinbase to be spent to a single
 /// shielded output, which is `z_shieldcoinbase`'s job.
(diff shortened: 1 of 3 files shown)


zallet-core/src/components/json_rpc/payments.rs
```diff
@@ -169,10 +178,31 @@ pub(super) fn confirmations_policy_for_minconf(
...
+                        .is_some()
+                        .then_some(ORCHARD_RECEIVER_POOLS)
+                        .into_iter()
+                        .flatten(),
+                ),
+        ),
+        // A TEX address (ZIP 320) names transparent funds held by a counterparty and
+        // cannot correspond to a wallet account, so no spend policy applies; account
+        // resolution rejects it before this is consulted.
+        _ => SpendPolicy::shielded_pools([]),
```
schell: Claude says:
> The catch-all `_ => SpendPolicy::shielded_pools([])` silently returns an empty policy for any future Address variant. If a new shielded address type is added, it will get no pools instead of the previous default-all-pools behavior. An explicit `Address::Tex` arm (or a comment) would make the intent clear and prevent a future variant from silently getting an empty policy.

pacu: Replaced the catch-all with an explicit `Address::Tex(_)` arm (with a comment on why it permits nothing), so a future `Address` variant is now a compile error here instead of silently receiving an empty policy. 🤖

schell (dismissed):
utACK cc7438b

schell:
Oh, @pacu we need to either update the integration test to use an Orchard source for Test 2 of `qa/rpc-tests/wallet_ironwood_crosspool.py` (since the intent is to combine Sapling + Ironwood, and an Orchard receiver permits both pools), or the test needs to be split so the cross-pool case uses a source whose receivers actually name both pools.

pacu:
> Oh, @pacu we need to either update the integration test to use an Orchard source for Test 2 of `qa/rpc-tests/wallet_ironwood_crosspool.py` (since the intent is to combine Sapling + Ironwood, and an Orchard receiver permits both pools), or the test needs to be split so the cross-pool case uses a source whose receivers actually name both pools.

Good catch! I'll fix that.

pacu:
Integration-test fix is up: zcash/integration-tests#193 — Test 2 of `wallet_ironwood_crosspool.py` now sends from a UA carrying both Sapling and Orchard receivers (so the receiver-derived policy permits the Sapling + Ironwood combination it exercises), and the file's header comment no longer describes the old every-pool policy. Test 1 needed no change (Orchard-only source spending Ironwood is still permitted). This PR's body now carries `ZIT-Revision: fix/ironwood-crosspool-receiver-pools` so CI here exercises that branch until both merge. 🤖

schell (dismissed):
utACK ae17cdc930

schell:
Ugh. The `test-rpc-venv-ubuntu-latest-py3.12` cache is poisoned - it restores but the `venv/bin/python3` binary is missing. This will keep failing on every re-run until the cache is invalidated or the workflow is fixed.

schell (approved):

PR #760, opened 2026-08-10
rpc: Refactor get_new_account and recover_accounts to use fetch_account_birthday helper

### LINKED ISSUE #759: zallet-core: Refactor get_new_account and recover_accounts to use fetch_account_birthday helper
`fetch_account_birthday` in `utils.rs` constructs an `AccountBirthday` by fetching the treestate at a given height. Two other call 
  sites inline the same logic and should be refactored to use the helper:                                                            
                                                                                                                                     
  - `get_new_account.rs` (lines 56–80)                                                                                               
  - `recover_accounts.rs` (lines 86–103)                                                                                             
                                                                                                                                     
  Neither is a drop-in substitution currently. To support both, the helper would need:                                             
  - An optional `recover_until` parameter (`recover_accounts` passes `Some(recover_until)` to `AccountBirthday::from_parts`, while   
  the helper currently hardcodes `None`)                                                                                           
  - Accept a `ChainView` instead of `Chain` to avoid creating redundant snapshots (both call sites already hold a snapshot)  
                                                                                                                                     
  Context: [#431 (comment)](https://github.com/zcash/zallet/pull/431#discussion_r3615484619)

mnm458:
Refactors `get_new_account` and `recover_accounts` to use the existing`fetch_account_birthday` helper in `utils.rs` instead of inlining treestate fetching and `AccountBirthday` construction. 

The helper's signature changes from `<C: Chain>` to `<V: ChainView>` so callers can pass their own snapshot (avoiding a redundant snapshot), and gains an optional `recover_until` parameter so `recover_accounts` can pass its scan-forward bound. Existing callers (`import_key`, `import_viewing_key`) now create a snapshot before calling the helper. 

As a side effect, `get_new_account`'s `.expect("always in range")` on a missing treestate becomes a proper RPC error instead of a panic.                                                                   
                                                                                                                                     
  Closes https://github.com/zcash/zallet/issues/759


nullcopy (approved):
utACK 786001409dc7ceaa2b93c32cdabd9249b6952106

PR #777, opened 2026-08-15
rpc: Add z_listreceivedbyaddress

### LINKED ISSUE #84: rpc: Implement `z_listreceivedbyaddress`

pacu:
Implements the `z_listreceivedbyaddress` JSON-RPC method. Closes #84.

Data access goes exclusively through a new typed librustzcash API, `WalletRead::get_account_received_outputs` (zcash/librustzcash#2972, implemented on branch `pacu/account-received-outputs`, which this PR pins via `[patch.crates-io]`); no raw SQL was added to Zallet. The filters and pagination are pushed down into a single indexed wallet-DB query, so large wallets only pay for the page they request.

Semantics (documented in `book/src/zcashd/json_rpc.md`):
- **Shielded/unified addresses match at the account level**: any address of a unified account returns the account's received outputs irrespective of diversifier, including internal change notes (reported with `change: true`). This is deliberately broader than `zcashd`, which reported only what the given address received. The reason it is broader: an address filter resolves to an `addresses` row, and internal shielded receivers are never stored as addresses, so a per-address filter could not report change at all. Note this is *not* the behaviour of `z_listunspent`, which matches shielded notes against the exact address it is given — reviewers should weigh whether the broader listing is the semantics we want here.
- **Transparent addresses match per-address**, because the account synthesized by `zcashd` `wallet.dat` migration aggregates many otherwise-unrelated imported taddrs.
- Bare receivers of a larger UA in the wallet are rejected as in `zcashd`; an address whose receivers exactly equal a wallet address's receivers (e.g. the Sapling address of a `z_importkey` account) is accepted.
- Spent outputs are listed (this is a received-history listing, not a UTXO listing); locked notes are included, consistent with `z_listunspent`.
- New optional `offset`/`limit` parameters, following `z_listtransactions`, page over a deterministic height-ascending ordering.

## Dependency pin

`[patch.crates-io]` points at `pacu/account-received-outputs` @ `8759c9e40f7333f89b184aa8f0563e8ce7f8c2bf`. That branch is based on `1f6bb2072e` — the librustzcash `main` rev this repository already pinned — so it retains the zero-expiry unmined-transaction fix (zcash/librustzcash#2983) that `migrate-zcashd-wallet` needs, and the crate versions are unchanged from `main` (`zcash_client_backend =0.24.0`, `zcash_client_sqlite =0.22.0`).

**This PR cannot merge until the librustzcash change lands in a release.** zcash/librustzcash#2972 is still an open issue; the API exists only on that branch. Review of the Zallet side can proceed independently.

## Testing

- Unit tests for the response rendering (memo/memoStr, unmined placeholder fields, watch-only `change` omission) and the minconf→height-bound push-down mapping.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across all three workspaces; `utils/check-lockstep.sh` passes.
- Previously verified end-to-end by the paired integration test in zcash/integration-tests against a locally built zebrad+zainod+zallet stack (all 13 sub-cases passing, including minconf/asOfHeight semantics, bare-receiver errors, and pagination). **That run predates this rebase onto `main` and onto the newer librustzcash base, so it should be re-run before merge.**

## Known issue, out of scope

A transaction whose only relevance to the wallet is a transparent output (detected from the mempool) does not get its mined height stamped by the block scan — nothing in the block trial-decrypts — and the asynchronous status-polling path did not converge within the test's timeout, so such receipts stay at 0 confirmations in wallet views for an extended period. This affects `z_listunspent` equally and is noted in the integration test with a `minconf=0` workaround. Tracked in #778, with root-cause analysis and fix options.

ZIT-Revision: wallet-z-listreceivedbyaddress

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,13 @@ be considered breaking changes.
   to complete even when some accounts or transparent spending keys in the
   `zcashd` wallet could not be imported. The skipped items are reported as
   warnings; they remain accessible only via the original `wallet.dat` file.
+- `z_listreceivedbyaddress` JSON-RPC method, returning amounts received by an
+  address belonging to the wallet. For a shielded or unified address that
+  corresponds to a unified account, the account's received outputs are
+  returned irrespective of the provided address's diversifier, matching
+  `z_listunspent`'s account-level semantics; a transparent address returns
+  only that address's outputs. New `offset` and `limit` parameters (as for
+  `z_listtransactions`) allow paginating large results.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,16 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Added
+
+- `z_listreceivedbyaddress` JSON-RPC method, returning amounts received by an
+  address belonging to the wallet. For a shielded or unified address that
+  corresponds to a unified account, the account's received outputs are
+  returned irrespective of the provided address's diversifier, including the
+  change received on its internal addresses; a transparent address returns
+  only that address's outputs. New `offset` and `limit` parameters (as for
+  `z_listtransactions`) allow paginating large results.
+
 ## [0.1.0-beta.3] - 2026-08-24
 
 ### Added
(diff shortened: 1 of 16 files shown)


pacu:
This is a low priority given 
https://docs.google.com/spreadsheets/d/1UJxH1cowexGqadU32Uei5Qak6jGhXjb18-T_QBPmDAA/edit?gid=0#gid=0

PR #782, opened 2026-08-17
Add `createmultisig` and `addmultisigaddress`, and fix the `listaddresses` omission they depend on

### LINKED ISSUE #48: rpc: Implement `addmultisigaddress`

### LINKED ISSUE #91: rpc: Implement `createmultisig`

sellout:
Closes #91. Closes #48.

Adds `createmultisig` and `addmultisigaddress`, and first fixes the
`listaddresses` defect that would otherwise make the second of them useless.

### `listaddresses` (first two commits)

`listaddresses` built its transparent lists from `list_addresses` alone, which
reports only addresses that have been exposed. The row recording a standalone
import — a public key or a redeem script — is written without an exposure
height, so such an address was not listed at all until funds arrived at it, and
was then listed among the account's derived addresses, which it is not one of.

The bucketing is extracted into a `bucket_transparent` predicate so the decision
is testable without standing up a chain, and then merged with
`get_transparent_receivers`, whose `include_standalone` flag exists for exactly
this. Only its underived entries are taken: it applies no exposure filter at
all, so it also holds every derived receiver `list_addresses` withholds — the
transparent receiver cached on each Unified Address, and the addresses generated
ahead of use to fill the gap limit — and reporting those would put addresses the
wallet has never handed out into `derived_transparent`.

An import is reported in the `transparent` object rather than
`derived_transparent`, even when the account holding it is derived. Reporting it
as derived would state that it can be re-derived from that mnemonic, when only a
full `wallet.db` backup restores imported material.

The first commit is deliberately red: two of its five tests fail, representing
the defect, and the second commit clears them.

### `createmultisig` and `addmultisigaddress`

Both build the same m-of-n redeem script through one helper, as `zcashd` did
from `_createmultisig_redeemScript`; `addmultisigaddress` is that plus recording
the script against an account. The script comes from `zcash_script`'s
`pattern::check_multisig` and is read back through the same `solver::standard`
that `decodescript` uses, so what Zallet builds and what the rest of the stack
reads cannot drift apart.

A key may be given as a hex-encoded public key or as a transparent address the
wallet holds the public key for. A hex key is placed in the script in the
encoding it was given, so an address `zcashd` created from uncompressed keys can
be reproduced exactly.

Two departures from `zcashd`'s `addmultisigaddress` signature, both forced by
Zallet's account model: the third argument is an account UUID where `zcashd`
took an account label (omitted, it falls back to the legacy `zcashd` pool, which
requires `features.legacy_pool_seed_fingerprint`), and there is no `rescan`
argument, as a newly created address has no history.

### Known limitation

Zallet cannot yet spend from a multisig address: it has no way to assemble a
P2SH `scriptSig`. Funds sent to one are tracked but cannot be moved. The book
and CHANGELOG say so rather than implying the funds are usable.

### Testing

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D
warnings` under both default and `zcashd-import,rpc-cli,tokio-console`, and the
full test suite for all three workspaces under the CI feature set, all clean.
Each commit passes clippy individually; the first is red only in the two tests
its message names.

There is a companion integration test — `qa/rpc-tests/wallet_multisig.py`,
covering both methods end to end against a regtest stack — which is not yet
pushed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,13 @@ be considered breaking changes.
   pre-Sapling wallet that has no HD seed. The minted seed never existed in
   zcashd, so backups of `wallet.dat` do not cover it; the migration prints a
   warning to that effect.
+- `createmultisig`, which builds an m-of-n multisignature redeem script and
+  reports its P2SH address. Each key may be given as a hex-encoded public key or
+  as a transparent address this wallet holds the public key for. A key given in
+  hex is placed in the script in the encoding it was given, so an address
+  created by `zcashd` from uncompressed keys can be reproduced exactly.
+  Note that Zallet cannot yet spend from a multisig address, as it has no way to
+  assemble a P2SH `scriptSig`; funds sent to one can be tracked but not moved.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,13 @@ be considered breaking changes.
   created by `zcashd` from uncompressed keys can be reproduced exactly.
   Note that Zallet cannot yet spend from a multisig address, as it has no way to
   assemble a P2SH `scriptSig`; funds sent to one can be tracked but not moved.
+- `addmultisigaddress`, which creates the same address and redeem script as
+  `createmultisig` and additionally records the script, so that the wallet
+  detects funds sent to the address. Its third argument is the UUID of the
+  account to track the address in, where `zcashd` took an account label; omitted,
+  it defaults to the legacy `zcashd` pool of funds, which requires
+  `features.legacy_pool_seed_fingerprint` to be set. Unlike `z_importaddress`
+  there is no `rescan` argument, as a newly created address has no history.
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


zallet-core/src/components/json_rpc/methods/list_addresses.rs
```diff
@@ -216,45 +272,84 @@ pub(crate) fn call(wallet: &DbConnection) -> Response {
...
+                    underived_addresses.append(&mut buckets.external);
+                    underived_change_addresses.append(&mut buckets.change);
                 }
             }
 
+            // A standalone import is reported here even when its account is derived.
+            // It is not reachable from the account's seed and ZIP 32 index, so listing
+            // it beside the addresses that are would misdescribe what restoring from
+            // that mnemonic recovers.
+            underived_addresses.append(&mut buckets.imported);
```
sellout, line 340: Somewhere around here is where ZIP 48 might affect things – `listaddresses` basically splits its response between derived addresses (recoverable from a mnemonic) and non-derived addresses (which aren’t). But ZIP 48 addresses _are_ derived, but can’t be recovered via a mnemonic because it also requires the XPubs from the other signers.

So, is the current structure (`derived_transparent` and `transparent` fields) fine for now, or should we work out something that ZIP 48 falls into nicely once it’s here?

CHANGELOG.md
```diff
@@ -296,6 +310,13 @@ be considered breaking changes.
   `walletpassphrase` unlocks a running server’s identities rather than this
   process’s, which cannot run concurrently with it in any case. The regtest
   account command is fixed the same way.
+- `listaddresses` now reports transparent addresses imported as a standalone
```
nullcopy, line 313: I have mixed feelings about this. Presumably `listaddresses` is used by a wallet to get an address to provide for a payment. If P2MS/ZIP-48 addresses are listed alongside other standalone P2SH addresses and HD derived addresses, then we're inviting problems, IMO.

I can't imagine a scenario where a wallet would ask for an address, and not care if its one of the HD addresses, or a P2MS address, and which P2MS if P2MS, etc

nullcopy (changes requested):
We decided in chat, that this API should not be implemented (for now) and instead provide an ZIP 48 impl for multisig users

PR #783, opened 2026-08-17
rpc: actually reserve the `__cookie__` JSON-RPC username

### LINKED ISSUE #664: "reserved" RPC username `__cookie__` is not actually reserved
RPC user can be set to __cookie__ and that should be reserved

pacu:
Closes #664.

## The bug

`__cookie__` is documented as the username belonging to the cookie credential Zallet
generates at startup, but nothing enforced it. A `[[rpc.auth]]` entry could claim the
name: Zallet logged a warning (`rpc-cookie-user-conflict`), **skipped cookie generation
entirely**, and then served `__cookie__` with the configured password.

Two consequences:

- A long-lived credential configured in plain text in `zallet.toml` presented itself
  under the username that clients treat as the short-lived cookie credential — the one
  that exists only in a `0600` file for the lifetime of the process and is removed on
  shutdown.
- With no cookie file written at all, `zallet rpc`'s default local authentication
  silently stopped working for anyone not using the configured credentials.

## The fix

Zallet now refuses to start if a `[[rpc.auth]]` user is named `__cookie__`, and always
generates its cookie. The check lives in two places:

- In `spawn()`, before the cookie file is written, so the operator gets a specific error
  naming the reserved username and what to do about it.
- In `AuthorizationLayer::new()`, so no construction path can produce a layer that
  serves the reserved name. This is where the unit tests bite.

`zallet add-rpc-user __cookie__` is rejected on the same grounds, rather than printing a
config block that Zallet would then refuse to start with.

## Commits

Per CONTRIBUTING, the bug is demonstrated by a failing unit test in a commit preceding
the fix:

1. `test:` — adds `configured_cookie_user_is_rejected` and
   `configured_cookie_user_is_rejected_alongside_valid_users`, which fail at that commit.
   Also retargets `pwhash_of_empty_password_is_rejected` at an ordinary username, so it
   keeps testing empty-password rejection rather than the new rule.
2. `rpc:` — the fix. The pre-existing test asserting the old behavior
   (`configured_cookie_user_works_without_generated_cookie`) is what commit 1 inverts.

## Compatibility

This is a breaking config change, and is called out as such in `CHANGELOG.md`: anyone
who configured `__cookie__` as an ordinary RPC user must rename that entry and update
the clients using it before upgrading. That seems right for an `I-SECURITY` issue — the
alternative is continuing to serve the reserved name — but say the word if you'd prefer
a deprecation cycle.

Note that `migrate-zcash-conf` can still translate an `rpcauth=__cookie__:...` line from
a `zcash.conf` into a config that Zallet refuses to start with. I left that alone: the
startup error names the offending entry and says what to change. Happy to add a
migration-time warning if you'd rather catch it there.

## Testing

- `cargo test -p zallet-core --lib` — 362 passed.
- `cargo test --manifest-path backends/zebra/Cargo.toml` — 9 passed, including the
  `example_config` golden (updated for the new `user` doc comment).
- `cargo clippy --all-targets -- -D warnings` and the same for `backends/zebra` — clean.
- `cargo fmt --all --check` — clean.

Docs updated: `book/src/cli/rpc.md`, the `RpcAuthSection::user` rustdoc, and the
generated example config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,19 @@ be considered breaking changes.
 
 ### Fixed
 
+- The `__cookie__` JSON-RPC username is now actually reserved. It is documented
+  as belonging to the cookie credential that Zallet generates at startup, but a
+  `[[rpc.auth]]` entry could claim it: Zallet warned, skipped cookie generation,
+  and served that name with the configured password, so a configured credential
+  gained access under the username clients treat as the short-lived,
+  filesystem-permission-protected cookie. Zallet now refuses to start if a
+  `[[rpc.auth]]` user is named `__cookie__`, and always generates its cookie.
+  `zallet add-rpc-user __cookie__` is rejected for the same reason, instead of
+  printing a config block that would not start.
+
+  Anyone who relied on configuring `__cookie__` as an ordinary RPC user must
+  rename that `[[rpc.auth]]` entry before upgrading, and update the clients
+  using it.
 - `migrate-zcashd-wallet` now passes a fallback network to the wallet parser
   for regtest wallets. A pre-Sapling wallet (zcashd < 5.0.0) has no
   `networkinfo` record; without a fallback, parsing failed with "wallet has no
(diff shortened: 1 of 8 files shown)


schell (approved):
utACK 4502908

PR #793, opened 2026-08-20
migrate-zcashd-wallet: expose imported standalone spending-key addresses

nullcopy:
Zallet's `listaddresses` only surfaces transparent addresses that have been marked as exposed. The ZeWIF importer registers standalone transparent spending keys (zcashd `importprivkey`) with their owning accounts but only marks addresses exposed on document-recorded or transaction-implied evidence, so an imported-but-never-funded key was invisible after migration even though zcashd always listed it.

After the import commits, derive each registered key's P2PKH address from its compressed pubkey and mark it exposed at the same height used for watch-only pubkey registration. Keys the importer skipped are excluded (their addresses have no wallet rows to mark), and so are addresses that exist as derived receivers: zcashd stored seed-derived keys as standalone `key` records too, and those must keep the importer's gap-inferred exposure — force-exposing a derived address beyond the gap limit could hide funded addresses from a fresh-seed recovery.

Also chains the post-import steps so the backup reminder prints before their errors propagate, and scopes the import-report exposure count to document-recorded exposures.

Replaces the last outstanding piece of `nc/migration-pin-bump`.

https://claude.ai/code/session_01S7s9MQzQ3Bmp42uabiSDbU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,14 @@ be considered breaking changes.
 
 ### Fixed
 
+- `migrate-zcashd-wallet` now marks the addresses of imported standalone
+  transparent spending keys (from zcashd's `importprivkey`) as exposed, so
+  `listaddresses` surfaces them even if they were never funded, as `zcashd`
+  did. Addresses that are also seed-derived receivers of an imported account
+  are excluded — both here and in the watch-only pubkey registration — so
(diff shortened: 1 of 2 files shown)


schell (dismissed):
utACK 8689289

I think it needs a FIXED changelog entry, though.

schell (approved):
utACK 172d835

PR #796, opened 2026-08-22
deps: bump librustzcash to main for the zero-expiry read-back fix

### LINKED ISSUE #795: Bump librustzcash to current main to pick up the zero-expiry transaction read-back fix
## Summary

Bump the wallet-critical librustzcash crates to current `main`, to pick up the fix for [zcash/librustzcash#2975](https://github.com/zcash/librustzcash/issues/2975) and to move the pinned version requirements off release candidates now that `zcash_client_sqlite` 0.22.0 and `zcash_client_backend` 0.24.0 are published.

## Why now

`migrate-zcashd-wallet` still aborts with `Consensus branch ID not known` when the `zcashd` wallet holds an **unmined** transaction with `nExpiryHeight = 0`, such as an imported coinbase transaction. The importer stores such a transaction with neither a mined height nor a usable expiry height, and the read-back it performs to tally the import then has no epoch from which to select a consensus branch ID, so `parse_tx` fails and the whole migration aborts.

zcash/zallet#727 fixed the common case at the source by backfilling resolved main-chain mined heights, but a genuinely unmined zero-expiry transaction still reproduces it. That residual case is fixed upstream by [zcash/librustzcash#2983](https://github.com/zcash/librustzcash/pull/2983), merged as `1f6bb2072e7fcb142b0d90ff7b267a8699a84818`:

- `parse_tx` gained a `fallback_height` parameter, supplied by `chain_tip_height(conn)` from `get_transaction` and `get_txs_spending_transparent_outputs_of`. The consensus branch ID (which does not affect parsing) is now selected from that height; the error survives only when the chain tip is also unknown.
- `zewif::import_wallet` now establishes the chain view from the document when account import does not, so a pre-Sapling-birthday `zcashd` wallet is not left tip-less — which would otherwise have defeated the fallback for exactly this migration path.

The fix is **not in any published release**: `zcash_client_sqlite` 0.22.0 was cut on 2026-08-18, two days before the merge, so it sits in the crate's `[Unreleased]` section. Picking it up therefore requires moving the existing `[patch.crates-io]` git rev forward, not dropping the patch.

## Secondary cleanup

The `[patch.crates-io]` block in all three workspace manifests is annotated:

> TEMPORARY: pin the wallet-critical librustzcash crates to the branch adding `WalletWrite::import_standalone_transparent_address` (zcash/librustzcash#2941) […] Repoint at crates.io once the change lands in a release.

That stated reason is now stale. #2941 landed on `main` via zcash/librustzcash#2944, *before* the 0.22.0 release, so `import_standalone_transparent_address` is in the published crates and no longer justifies a git pin on its own. The comment should be rewritten to name the reason that does still apply (the unreleased #2983 fix), so the next person to read it knows what has to land before the pin can go.

Relatedly, the version requirements are still pinned to release candidates — `zcash_client_backend = "=0.24.0-rc.7"` and `zcash_client_sqlite = "=0.22.0-rc.7"` — and should move to the published `=0.24.0` / `=0.22.0`.

## Scope

- Repoint the `[patch.crates-io]` librustzcash rev from `375fc8e5` to `1f6bb207` in `Cargo.toml`, `backends/zebra/Cargo.toml`, and `backends/zaino/Cargo.toml`.
- Bump `zcash_client_backend` to `=0.24.0` and `zcash_client_sqlite` to `=0.22.0` in the same three manifests.
- Rewrite the stale `TEMPORARY` comment in all three.
- Regenerate the three lockfiles, keeping `utils/check-lockstep.sh` green.
- Add a `CHANGELOG.md` entry under `Fixed`, since the migration failure is user-facing.

## Follow-up

Once a `zcash_client_sqlite` release contains #2983, the `[patch.crates-io]` block can be dropped entirely and the crates consumed from crates.io — worth a separate issue at that point.

---

🤖 Investigated and written up with Claude (Claude Code) by @pacu

pacu:
Closes #795

Bumps the wallet-critical librustzcash crates from `375fc8e5` to current `main` (`1f6bb2072e7fcb142b0d90ff7b267a8699a84818`), to pick up the fix for [zcash/librustzcash#2975](https://github.com/zcash/librustzcash/issues/2975).

## What this fixes

`migrate-zcashd-wallet` aborts with `Consensus branch ID not known` when the `zcashd` wallet holds an **unmined** transaction with `nExpiryHeight = 0`, such as an imported coinbase transaction. The importer stores such a transaction with neither a mined height nor a usable expiry height, and the read-back it performs to tally the import then has no epoch from which to select a consensus branch ID, so `parse_tx` fails and the migration aborts.

#727 fixed the common case at the source by backfilling resolved main-chain mined heights; a genuinely unmined zero-expiry transaction still reproduces it. [zcash/librustzcash#2983](https://github.com/zcash/librustzcash/pull/2983) fixes that residual case upstream:

- `parse_tx` gained a `fallback_height` parameter, supplied by `chain_tip_height(conn)` from `get_transaction` and `get_txs_spending_transparent_outputs_of`. The consensus branch ID does not affect parsing (absent from the pre-v5 serialized form; v5 onward commit to their own), so it is now selected from that height. The error survives only when the chain tip is also unknown.
- `zewif::import_wallet` now establishes the chain view from the document when account import does not, so a pre-Sapling-birthday `zcashd` wallet is not left tip-less — which would otherwise have defeated the fallback for exactly this migration path.

## Why the patch block stays

The fix postdates the `zcash_client_sqlite` 0.22.0 release (cut 2026-08-18, two days before the merge) and sits in the crate's `[Unreleased]` section, so it is reached by moving the `[patch.crates-io]` rev forward rather than by dropping the patch.

The block's `TEMPORARY` comment is rewritten, because its stated reason had expired: `WalletWrite::import_standalone_transparent_address` (zcash/librustzcash#2941) landed via zcash/librustzcash#2944 *before* the 0.22.0 release, so it is in the published crates and no longer justifies a git pin on its own. The comment now names the reason that does still apply, so the next reader knows what has to land before the pin can go. Dropping the patch entirely is noted as follow-up in #795.

## Also

Version requirements move off release candidates now that both crates are published: `zcash_client_backend` `=0.24.0-rc.7` → `=0.24.0`, `zcash_client_sqlite` `=0.22.0-rc.7` → `=0.22.0`.

## Testing

- `utils/check-lockstep.sh` passes — 22 crates + 4 package versions consistent across all three lockfiles.
- `cargo check --workspace --all-targets` is clean in all three workspaces (root, `backends/zebra`, `backends/zaino`); none of the 63 upstream commits in the range broke an API Zallet uses.

https://claude.ai/code/session_01HejoFsphMwaw8EdqVa6U8J

### DIFF
--- a/backends/zaino/Cargo.toml
+++ b/backends/zaino/Cargo.toml
@@ -34,7 +34,7 @@ transparent = { package = "zcash_transparent", version = "0.10.0" }
 zaino-common = "0.4"
 zaino-fetch = "0.4"
 zaino-state = "0.5"
-zcash_client_backend = "=0.24.0-rc.7"
+zcash_client_backend = "=0.24.0"
 zcash_keys = { version = "0.16.0", features = ["unstable"] }
 zcash_primitives = "0.30.0"
 zcash_protocol = "0.10.0"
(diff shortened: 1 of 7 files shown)


backends/zaino/Cargo.toml
```diff
@@ -76,25 +76,27 @@ tempfile = "3"
...
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_pool_migration = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_transparent = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zip321 = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
```
nuttycom, line 99: Why is this section still present at all?

backends/zebra/Cargo.toml
```diff
@@ -79,25 +79,27 @@ tokio-console = ["zallet-core/tokio-console"]
...
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_pool_migration = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zcash_transparent = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
+zip321 = { git = "https://github.com/zcash/librustzcash.git", rev = "1f6bb2072e7fcb142b0d90ff7b267a8699a84818" }
```
nuttycom, line 102: It should be possible to delete or comment these out.

nuttycom (changes requested):
We should no longer need the patch.crates-io sections at all.

nuttycom (approved):
oops, I hadn't read the justification carefully enough.

PR #816, opened 2026-08-31
migrate-zcashd-wallet: Fix standalone P2SH support

### LINKED ISSUE #572: Fix standalone P2SH support in zcashd migration
See: https://github.com/zcash/zallet/pull/571/changes#r3554335918

sellout:
Closes #572.

## Motivation

#571 removed the `cscripts()` walk from `migrate-zcashd-wallet` on the basis that the
ZeWIF importer covers standalone P2SH, and [noted](https://github.com/zcash/zallet/pull/571#discussion_r3554335918)
that it "was removed and not replaced". The importer covers it two-thirds of the way:

- `zewif-zcashd` emits one legacy-account address entry per `cscript`, keyed by the P2SH
  address, carrying the redeem script and no spend authority or pubkey.
- `zcash_client_sqlite::zewif` registers those via `import_standalone_transparent_script`,
  but accepts only multisig within the 520-byte P2SH limit. Everything else is counted in
  `redeem_scripts_not_representable` and **dropped**, leaving the migrated wallet blind to
  funds `zcashd` could see.
- Its exposure pass marks an address only where the document accounts for one, and its
  child-index inference cannot reach an address with no derivation — so even the *accepted*
  P2SH addresses ended up with no exposure height, and `listaddresses` (which reports only
  exposed addresses) did not surface them where `zcashd` did.

## Solution

`finish_import` gains two steps after the existing ones:

- `register_watch_addresses` registers, as bare watch-only addresses, the transparent
  addresses `zcashd` watched but for which no key material reached the wallet — both
  `importaddress <address>` and the redeem scripts the wallet cannot represent. Membership
  is decided by reading back what the wallet tracks rather than by re-deriving the
  representability rules, so the two cannot drift apart.
- `expose_registered_script_addresses` marks the P2SH addresses of the registered redeem
  scripts exposed. A redeem script reaches a `zcashd` wallet only because its address was
  already in use outside it.

The preceding commits are the refactors that make those two expressible, each a stated
no-op. The last four address review findings on the branch: an unparsable watch pubkey is
now reported separately from an uncompressed one; `register_watch_pubkeys` no longer
imports a pubkey whose address the wallet already tracks (which reached
`StandaloneImportConflict` and failed a migration *after* it had committed, when migrating
a second `zcashd` wallet that shared an `importpubkey` entry); the watched-address
registration and its exposures now share one database transaction; and the dormant
document-exposure-height branch is documented as dormant.

## Test evidence

`cargo fmt --all -- --check`, `cargo clippy --all-targets --features
zcashd-import,rpc-cli,tokio-console -- -D warnings`, and `cargo test --features
zcashd-import,rpc-cli,tokio-console` all pass locally — 426 tests.

Unit tests cover each selection: which addresses are registered and which are left to the
importer, exposure-height precedence, per-address deduplication and undecodable strings,
which P2SH addresses need exposing, and which watch pubkeys survive the tracked-receiver
filter. Each was checked to fail against the behaviour it pins.

Not covered: the database-touching wrappers themselves have no integration test, as the
crate has no `migrate-zcashd-wallet` harness. The ordering and conflict-avoidance claims
are argued in doc comments rather than asserted by a test.

### Unrelated flaky test

`components::sync::tests::wallet_sync_error_shuts_down_the_spawned_batch_decryptor` fails
intermittently. It reproduces on this branch *without* these commits (1 failure in 4
full-suite runs at the pre-existing tip), so it is not introduced here. Its assertion
admits three outcomes and handles two: after a failed `WalletSync::spawn`, `reload_keys()`
returning `Some` whose future resolves `Ok` — the engine serving the reload before it shuts
down — trips the assert. Filed as #818.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,13 @@ be considered breaking changes.
   see. As in `zcashd`, and as for any address imported with `zallet
   import-address`, funds they receive are reported but cannot be spent without
   importing the corresponding key material.
+- `migrate-zcashd-wallet` no longer aborts when a `zcashd` wallet's watch-only
+  pubkeys (from `importpubkey`) include one whose address the Zallet wallet already
+  tracks. Every such pubkey was registered regardless, and the wallet rejects an
+  import of key material another account holds, so migrating a second `zcashd`
+  wallet that shared a watched key failed with a database error *after* the import
+  had committed. Such an address now keeps the registration and the exposure it
+  already had, as the watch-only address registration does for its own candidates.
 - `migrate-zcashd-wallet` prints the backup reminder before reporting an
   error from the post-import registration and verification steps.
 - The `__cookie__` JSON-RPC username is now actually reserved. It is documented
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,15 @@ be considered breaking changes.
   did. Addresses that are also seed-derived receivers of an imported account
   are excluded — both here and in the watch-only pubkey registration — so
   they keep the gap-limit-inferred exposure that seed recovery depends on.
+- `migrate-zcashd-wallet` now records the P2SH addresses of the redeem scripts it
+  imports (from zcashd's `importaddress <redeemscript>` and
+  `addmultisigaddress`) as exposed. Such an address was in use outside the wallet
+  before it was imported, but it has no derivation, so it was reached by neither
+  the exposures recorded in the migrated wallet nor the gap-limit inference that
+  covers derived receivers; one that had never appeared in a wallet transaction
+  was left with no exposure height at all. One consequence is that
+  `listaddresses`, which reports only exposed addresses, did not surface it where
+  `zcashd` did.
 - `migrate-zcashd-wallet` prints the backup reminder before reporting an
   error from the post-import registration and verification steps.
 - The `__cookie__` JSON-RPC username is now actually reserved. It is documented
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,15 @@ be considered breaking changes.
   was left with no exposure height at all. One consequence is that
   `listaddresses`, which reports only exposed addresses, did not surface it where
   `zcashd` did.
+- `migrate-zcashd-wallet` now migrates the transparent addresses that `zcashd`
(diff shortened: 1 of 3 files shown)


sellout:
Two failing checks, neither of which seems related to this work.

nullcopy (commented):
https://github.com/zcash/librustzcash/pull/3028 makes account spend authority pure (no more watch-only addresses in an account with spend authority). I suspect this PR needs to enforce that too, but I haven't confirmed.

sellout:
Confirmed, and thanks — it needed more than enforcement.

`zewif-zcashd` synthesises the zcashd legacy account with `AccountPurpose::Spending`
(`migrate/accounts.rs:123`), and `register_watch_addresses` targeted exactly the accounts
the ZeWIF import created. So for any seed-derived zcashd wallet — the common case — it was
registering watch-only material into a spending account, which is the shape #3028 forbids.

Gating it on `AccountPurpose` would have made it inert for that same common case, and the
material has nowhere else to go while `accounts.uivk` is `NOT NULL` (zcash/librustzcash#2582).
And since #3028 imports those records itself for view-only accounts and reports the rest in
`ZewifImportReport::skipped_watch_only`, the step had no remaining niche either way. So I've
dropped it rather than gated it.

What's left here is the half #3028 doesn't touch: the exposure marking. A P2SH address
registered from a redeem script has no derivation, so the importer's child-index inference
can't reach it and it ends up with no exposure height, which keeps `listaddresses` from
surfacing it. Nine commits now, and this is "Part of #572" rather than closing it — it closes
jointly with #3028 and the follow-up.

Follow-ups filed:

- zcash/zallet#868 — consume `skipped_watch_only` / `watch_only_receivers_imported` instead
  of re-walking the document, and re-import into a view-only account once #2582 allows one.
  Consuming #3028 is a `[patch.crates-io]` rev bump rather than a release wait, since the
  cohort is already pinned to a git rev.
- zcash/zallet#869 — `register_watch_pubkeys` has the same problem, and so does `zallet
  import-address` with no `--account`, since `legacy_pool_account` only accepts a ZIP
  32-derived account. Both predate this branch, so separated rather than widened into it.

Issue #817, opened 2026-08-31
Transparent balance can stall part-way on an idle chain (z_gettotalbalance / z_listunspent under-report coinbase outputs)


zcash-harry:
## Summary

On regtest under CI load, `z_gettotalbalance`'s `transparent` field reports fewer coinbase outputs than the wallet owns, while `getwalletstatus` reports the wallet fully synced. It does not recover for as long as the chain stays idle.

Observed against zallet `main` @ f9040406 (v0.1.0-beta.3).

**Important caveat:** this does not reproduce on a normal workstation. Unpatched `main` passes the test below 3/3 locally with all 101 transactions enhanced. It reproduces consistently on GitHub-hosted CI runners, so it appears load- or timing-dependent. Everything below is from CI logs.

## Evidence

From a `zcash/integration-tests` CI run of `qa/rpc-tests/wallet.py` ([job log](https://github.com/zcash/integration-tests/actions/runs/33422328436/job/99590865264)). The test mines 101 regtest blocks to a wallet address and reads the balance.

| zallet time | event |
|---|---|
| 18:08:44 | zallet starts |
| 18:08:47 | `Scanning block 1` |
| **18:08:54** | **`Scanning block 101` — scan complete, ~10s in** |
| 18:08:54 → 18:10:5x | balance polled once a second for **120s** |
| — | never moves off **500.00000000 ZEC** = **80 coinbases** |

At that point:

- `getwalletstatus` reports `wallet_tip.height == 101`
- zallet's log shows it scanned through 101 and moved on to streaming mempool
- the node confirms all 101 coinbases at the address (`getaddressbalance` == `101 * 625000000`, asserted just above and passing)
- `z_gettotalbalance(1, true).transparent` returns 500 ZEC and holds there

The shortfall tracks the enhancement queue exactly: the last queue announcement was `90 transaction data requests to service`, **80** distinct transactions were enhanced, the last `data_requests` activity was at 18:08:54 — the instant scanning finished — and the balance corresponds to exactly those 80.

## Where it comes from

`z_gettotalbalance` → `get_wallet_summary` → `add_transparent_account_balances` sums `transparent_received_outputs`. Those rows are written by `decrypt_and_store_transaction`, which runs in the `data_requests` task in `zallet-core/src/components/sync.rs`, not by block scanning.

That task waits on the chain tip, services one snapshot of the queue, then goes back to waiting:

```rust
loop {
    tip_change_signal.notified().await;
[...]

Issue #766, opened 2026-08-12
Flaky test: wallet_sync_error_shuts_down_the_spawned_batch_decryptor


schell:
## Problem

The test `components::sync::tests::wallet_sync_error_shuts_down_the_spawned_batch_decryptor` in `zallet-core/src/components/sync.rs:1681` is flaky. It failed on macOS CI for PR #752 (which does not touch the sync module at all):

- Run: https://github.com/zcash/zallet/actions/runs/31545360975/job/93956568038
- 376 passed, 1 failed; the single failure was this test.

## Root cause

The test asserts that a spawned batch decryptor task is shut down when wallet sync startup fails. The assertion at line 1712-1716 has two acceptable outcomes:

```rust
if let Some(reload_finished) = decryptor_observer.reload_keys().await {
    assert!(
        reload_finished.await.is_err(),
        "failed wallet sync startup must not leave its batch decryptor running",
    );
}
```

- `None` from `reload_keys()` means the decryptor engine was already dropped (acceptable).
- `Some` with `Err` from `reload_finished.await` means the task was cancelled (acceptable).
- `Some` with `Ok` triggers the panic (the task completed normally instead of being cancelled).

The `Ok` path is a race: the batch decryptor task may complete its current work before the cancellation signal propagates, especially under CI runner load. The test is timing-sensitive in a way that produces false failures.

## Suggested fix

Introduce a deterministic signal that the shutdown path was reached, rather than relying on the task's completion outcome. Options:

1. **Use a barrier or oneshot channel** that the decryptor task signals when it observes the cancellation, and assert on that signal rather than on `reload_finished.await` outcome.
2. **Add a yield point or sleep** before the assertion to give the runtime time to propagate the cancellation. Less principled but minimal.
3. **Restructure the test** so the mock chain blocks the decryptor task at a known await point, ensuring it cannot complete before cancellation arrives.

Option 1 is the most robust. The current `Ok`/`Err` outcome is not a reliable proxy for "was the task shut down" because a task that finishes its current unit of work before observing the cancel signal will return `Ok`.

## Reproduction

Intermittent. Observed on macOS CI (`Test default on macOS` job, `Run tests (zallet-core)` step). Not reproducible on every run.

Issue #200, opened 2025-08-20
Consider making `builder.tx_expiry_delta` usable


str4d:
The `zcash_primitives` transaction builder has an `expiry_height` field internally, but it is always set to the default delta of 40 blocks after the target height. `zcash.conf` has a field allowing the wallet operator to configure this, and we currently migrate that config option into `zallet.toml`. We need to either make the necessary builder changes to allow setting this, or stop migrating the field and document its absence.

PR #819, opened 2026-08-31
sync: drain transaction data requests instead of one pass per tip change

zcash-harry:
Refs #817. Not a complete fix on its own — see Status below.

## The bug

`data_requests` waits on the chain tip, services **one snapshot** of the queue, then goes back to waiting:

```rust
loop {
    tip_change_signal.notified().await;   // wait for a tip change
    let requests = db_data.transaction_data_requests()?;
    for request in requests { /* service */ }
}                                          // loop back, block again
```

Anything still queued when a pass ends waits for the *next* tip change: requests added while the pass was running, entries deliberately kept for retry (`SpentSpenderUnknown`), and ones that hit a transient chain error.

While the chain is idle there is no next tip change, so that remainder sits unserviced. Because `decrypt_and_store_transaction` writes `transparent_received_outputs` — the table `get_wallet_summary` sums — and runs only in this task, the wallet can be fully scanned while RPCs like `z_gettotalbalance` and `z_listunspent` report a subset of the transparent outputs it owns.

## The change

Check the queue first and only wait on the tip when it is empty, so a pass that leaves work behind loops straight round and drains it.

The second wait is the important half: several arms deliberately leave entries queued, so an unconditional "loop until empty" would spin against them. Waiting whenever a pass fails to shrink the queue preserves the existing backoff for those cases while letting genuine progress continue without needing a new block.

No behaviour change when the queue empties in one pass, which is the common case.

## Evidence

From the integration-tests interop run this PR triggers ([run](https://github.com/zcash/integration-tests/actions/runs/33426089753)), against the same suite on the same hardware:

| | queue behaviour | balance observed |
|---|---|---|
| `main` | stalls at **80 of 90** queued | 80 of 101 coinbases |
| this PR | drains **110 → 10** | 95 of 101 coinbases |

The ~10 remaining are the deliberately-requeued kind, which is the intended steady state rather than a stall.

## Status

That run still fails, on a different cause: it pairs this branch with integration-tests `main`, whose `wallet.py` reads the balance **once**, two seconds after `wallet_tip`. It sampled 95 while enhancement was still in flight and needed ≥96 — it read too early rather than the wallet being stuck. zcash/integration-tests#200 turns that read into a bounded wait.

The two are complementary and **have not yet been tested together**: this run pairs a patched wallet with unpatched tests, and #200's CI pairs patched tests with an unpatched wallet. I'd like a combined run before calling this verified, which is why it says `Refs` rather than `Fixes`.

## Testing

- `cargo check -p zallet-core` passes; CI is green (the single NU7 Linux failure was the known flake #766 and passed on rerun).
- Locally the failure does not reproduce: unpatched `main` passes `wallet.py` 3/3 with all 101 transactions enhanced. So the local runs show this change is not harmful, but cannot show that it helps; the CI queue-drain behaviour above is the real signal.
- Worth a careful look at the no-progress guard in particular, and ideally a regtest case asserting the balance settles while the chain is idle.

### DIFF
--- a/zallet-core/src/components/sync.rs
+++ b/zallet-core/src/components/sync.rs
@@ -1290,17 +1290,23 @@ async fn data_requests<C: Chain>(
     tip_change_signal: Arc<Notify>,
 ) -> Result<(), SyncError> {
     loop {
-        // Wait for the chain tip to advance
-        tip_change_signal.notified().await;
+        let requests = db_data.transaction_data_requests()?;
+        if requests.is_empty() {
+            // Nothing to do; wait for the chain tip to advance and bring more work.
+            debug!("No transaction data requests, sleeping until the chain tip changes.");
+            tip_change_signal.notified().await;
+            continue;
+        }
 
         let chain_view = chain.snapshot().await.map_err(SyncError::Chain)?;
 
-        let requests = db_data.transaction_data_requests()?;
-        if requests.is_empty() {
-            // Wait for new requests.
-            debug!("No transaction data requests, sleeping until the chain tip changes.");
-            continue;
-        }
+        // Servicing a pass can leave work behind: requests queued while the pass was
+        // running, entries deliberately kept for retry (`SpentSpenderUnknown`), and ones
+        // that hit a transient chain error. Previously the loop went straight back to
+        // waiting on the chain tip, so on an idle chain that remainder was stranded and
+        // the wallet's transparent balance stayed frozen part-way. Re-check the queue
+        // after each pass instead, and only wait on the tip when a pass resolves nothing.
+        let outstanding_before = requests.len();
 
         let view_tip = chain_view.tip().await.map_err(SyncError::Chain)?.height();
         info!("{} transaction data requests to service", requests.len());
@@ -1398,6 +1404,17 @@ async fn data_requests<C: Chain>(
                     }
                 }
             }
+        }
+
+        // If the queue did not shrink, this pass resolved nothing that it can resolve on
+        // its own, so go back to waiting on the chain rather than spinning against the
+        // same entries. Otherwise loop straight round and drain what is left.
+        if db_data.transaction_data_requests()?.len() >= outstanding_before {
+            debug!(
+                "Transaction data requests made no progress this pass; \
+                 sleeping until the chain tip changes."
+            );
+            tip_change_signal.notified().await;
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,18 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- Wallet balances no longer stall part-way while the chain is idle. Transaction
+  data requests were only serviced when the chain tip advanced, so any request
+  still queued at the end of a pass — including ones added while that pass was
+  running — waited for the next block. On a chain that was not producing blocks
+  the remainder was never serviced, leaving RPCs such as `z_gettotalbalance` and
+  `z_listunspent` reporting a subset of the wallet's transparent outputs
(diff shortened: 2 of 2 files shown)


zcash-harry:
Some verification data, including a correction to what this PR claims.

## The queue now drains

The integration-tests interop run this PR triggered ([run](https://github.com/zcash/integration-tests/actions/runs/33426089753)) shows the intended change in behaviour:

| | queue behaviour | balance observed |
|---|---|---|
| `main` | stalls at **80 of 90** queued | 80 of 101 coinbases |
| this PR | drains **110 → 10** | 95 of 101 coinbases |

The ~10 that remain are the deliberately-requeued kind (unresolvable `GetStatus`), which is the intended steady state rather than a stall.

## It is not sufficient on its own

That run still fails, but on a different cause: it pairs this PR with integration-tests `main`, whose `wallet.py` reads the balance **once**, two seconds after `wallet_tip`. It sampled 95 while enhancement was still in flight and needed ≥96. It read too early rather than the wallet being stuck.

zcash/integration-tests#200 changes that read into a bounded wait. The two are complementary, and **neither has yet been tested against the other** — this run pairs a patched wallet with unpatched tests, and #200's own CI pairs patched tests with an unpatched wallet. I'd like to get a combined run before this is treated as verified.

## Correction to #817

I originally described the stranding as deterministic. That overstates it. On a normal workstation, unpatched `main` passes `wallet.py` 3/3 with all 101 transactions enhanced — the failure does not reproduce off CI hardware, so it is load-dependent. The mechanism described in #817 (a pass ending with work still queued, and nothing to re-trigger it while the chain is idle) still holds and is what this PR addresses; the "frozen indefinitely" framing was extrapolated from a single 120s CI window and was not established. I'll correct the issue text.

## Testing status

- `cargo check -p zallet-core` passes; zallet CI is green (the one NU7 Linux failure was the known flake #766, and it passed on rerun).
- Ran `wallet.py` locally against this branch 3× (2 pass, 1 fail) and against `main` 3× (3 pass). Locally the bug does not reproduce, so those runs show this change is not harmful but cannot demonstrate that it helps.
- The meaningful signal is the CI queue-drain behaviour above.

Happy to hold this until a combined run exists if you'd prefer that ordering.

zallet-core/src/components/sync.rs
```diff
@@ -1399,6 +1405,17 @@ async fn data_requests<C: Chain>(
                 }
             }
         }
+
+        // If the queue did not shrink, this pass resolved nothing that it can resolve on
```
nullcopy, line 1409: I don't think this assertion is true. What if the initial was 1 requests, and that request is resolved at the same time a new request arrives?

zallet-core/src/components/sync.rs
```diff
@@ -1399,6 +1405,17 @@ async fn data_requests<C: Chain>(
...
+        // its own, so go back to waiting on the chain rather than spinning against the
+        // same entries. Otherwise loop straight round and drain what is left.
+        if db_data.transaction_data_requests()?.len() >= outstanding_before {
+            debug!(
+                "Transaction data requests made no progress this pass; \
+                 sleeping until the chain tip changes."
+            );
+            tip_change_signal.notified().await;
```
nullcopy, line 1417: Why do you wait on the `tip_change_signal` here, instead of looping back to the beginning to wait on it there?

nullcopy (changes requested):
Unless I'm misunderstanding something, I don't think it's a good idea to:
1) measure the outstanding requests in this way
2) use the before/after delta in the # of requests as a signal

zallet-core/src/components/sync.rs
zcash-harry: I actually am closing this PR, will open another one. This one had some flaws and make the progress worse

PR #821, opened 2026-09-01
sync: signal data requests after storing chain data

### LINKED ISSUE #817: Transparent balance can stall part-way on an idle chain (z_gettotalbalance / z_listunspent under-report coinbase outputs)

zcash-harry:
Fixes #817.

## Problem

The data-request task waited for a tip-change signal. The signal fired before new blocks were stored. The task could find no requests and wait for the next block.

## Change

- Rename the signal to describe its purpose.
- Signal after new blocks are stored.
- Signal after history blocks are stored.
- Signal after each mempool transaction is stored.
- Keep the startup signal for requests from an earlier run.

## Commits

1. Revert the change from #820.
2. Apply the focused fix for #817.

There is no separate refactor commit.

## Testing

- cargo fmt --all -- --check
- cargo test -p zallet-core
- cargo test -p zallet-core --features spend-index

The end-to-end test is in zcash/integration-tests#200.

ZIT-Revision: harry/wallet-py-sync-barrier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,18 +25,6 @@ be considered breaking changes.
 
 ## [Unreleased]
 
-### Fixed
-
-- Wallet balances no longer stall part-way while the chain is idle. The
-  data-requests task was woken as soon as a new chain tip was observed, which
-  is before those blocks are scanned and before the transaction data requests
-  they generate exist, so it could wake, find nothing to do, and sleep again —
-  stranding that work until the next tip change. On a chain that was not
-  producing blocks it was never picked up, leaving RPCs such as
-  `z_gettotalbalance` and `z_listunspent` reporting a subset of the wallet's
-  transparent outputs. The task is now also woken once the blocks have been
-  applied and their requests queued.
-
 ## [0.1.0-beta.3] - 2026-08-24
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed a race that could leave wallet data requests unprocessed. The sync task
+  now signals data-request processing after new blocks are stored. It also
+  signals after each mempool transaction is stored.
+
 ## [0.1.0-beta.3] - 2026-08-24
 
 ### Added
--- a/zallet-core/src/components/sync.rs
+++ b/zallet-core/src/components/sync.rs
@@ -1028,15 +1028,6 @@ async fn steady_state_iteration<C: Chain>(
     // any) has been rescanned.
     status.set_fully_synced(db_data.block_fully_scanned()?.map(|m| m.block_height()));
     status.mark_tip_reached();
-
-    // Wake the data-requests task now that those blocks are applied. The notification
-    // above fires as soon as a new tip is *observed*, which is before the blocks have
-    // been scanned and before the requests they generate exist; a task woken by it can
-    // find nothing to do and go back to sleep, stranding that work until the next tip
-    // change. While the chain is idle there is no next tip change, so the wallet's
-    // transparent balance can sit part-way through the outputs it owns. Signalling
-    // again here means there is always a wakeup after the work is actually queued.
-    tip_change_signal.notify_one();
     // TODO(zcash/zallet#195): when this actually clears an in-progress recovery (i.e. on
     // the `Recovering` → synced edge, not on every tip-reached), trigger an online backup
     // so the recovered state is durably captured before any subsequent failure.
--- a/zallet-core/src/components/sync.rs
+++ b/zallet-core/src/components/sync.rs
@@ -172,11 +172,9 @@ impl WalletSync {
         // an atomic.
         let current_boundary = Arc::new(AtomicU32::new(starting_boundary.into()));
 
-        // TODO: Zaino should provide us an API that allows us to be notified when the chain tip
-        // changes; here, we produce our own signal via the "mempool stream closing" side effect
-        // that occurs in the light client API when the chain tip changes.
-        let tip_change_signal_source = Arc::new(Notify::new());
-        let req_tip_change_signal_receiver = tip_change_signal_source.clone();
+        // Wake the data-requests task after sync steps that can add or unblock requests.
+        let data_request_signal = Arc::new(Notify::new());
+        let data_request_signal_receiver = data_request_signal.clone();
 
         // Spawn the ongoing sync tasks.
         let steady_state_task = {
@@ -184,6 +182,7 @@ impl WalletSync {
             let lower_boundary = current_boundary.clone();
             let decryptor = decryptor.clone();
             let status = status.clone();
+            let data_request_signal = data_request_signal.clone();
             crate::spawn!("Steady state sync", async move {
                 steady_state(
                     chain,
@@ -191,7 +190,7 @@ impl WalletSync {
                     db_data.as_mut(),
                     starting_tip,
                     lower_boundary,
-                    tip_change_signal_source,
+                    data_request_signal,
                     decryptor,
                     shutdown_height,
                     status,
@@ -206,12 +205,14 @@ impl WalletSync {
             let chain = chain.clone();
             let mut db_data = db.handle().await?;
             let upper_boundary = current_boundary.clone();
+            let data_request_signal = data_request_signal.clone();
             crate::spawn!("Recover history", async move {
                 recover_history(
                     chain,
                     &params,
                     db_data.as_mut(),
                     upper_boundary,
+                    data_request_signal,
                     decryptor,
                     recover_batch_size,
                     shutdown_height,
@@ -229,7 +230,7 @@ impl WalletSync {
                 chain,
                 &params,
                 db_data.as_mut(),
-                req_tip_change_signal_receiver,
+                data_request_signal_receiver,
             )
             .await?;
             Ok(())
@@ -789,16 +790,15 @@ async fn steady_state<C: Chain>(
     db_data: &mut DbConnection,
     mut prev_tip: ChainBlock,
     lower_boundary: Arc<AtomicU32>,
-    tip_change_signal: Arc<Notify>,
+    data_request_signal: Arc<Notify>,
     decryptor: WalletDecryptorHandle,
     shutdown_height: Option<BlockHeight>,
     status: SyncStatusWriter,
 ) -> Result<(), SyncError> {
     info!("Steady-state sync task started");
 
-    // Wake up any tasks waiting on the tip-change signal, so they can service work that
-    // accumulated while the wallet was offline.
-    tip_change_signal.notify_one();
+    // Process data requests that accumulated while the wallet was offline.
+    data_request_signal.notify_one();
 
     let mut tree_recovery = TreeRecovery::default();
 
@@ -809,7 +809,7 @@ async fn steady_state<C: Chain>(
             db_data,
             &mut prev_tip,
             &lower_boundary,
-            &tip_change_signal,
+            &data_request_signal,
             &decryptor,
             shutdown_height,
             &status,
@@ -922,7 +922,7 @@ async fn steady_state_iteration<C: Chain>(
     db_data: &mut DbConnection,
     prev_tip: &mut ChainBlock,
     lower_boundary: &AtomicU32,
-    tip_change_signal: &Notify,
+    data_request_signal: &Notify,
     decryptor: &WalletDecryptorHandle,
     shutdown_height: Option<BlockHeight>,
     status: &SyncStatusWriter,
@@ -948,7 +948,6 @@ async fn steady_state_iteration<C: Chain>(
                 )
             })
             .expect("closure always returns Some");
-        tip_change_signal.notify_one();
         status.set_tip(current_tip.height());
 
         // Find where the wallet's history rejoins the backend's best chain.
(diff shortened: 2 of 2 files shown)


zallet-core/src/components/sync.rs
```diff
@@ -948,7 +941,10 @@ async fn steady_state_iteration<C: Chain>(
                 )
             })
             .expect("closure always returns Some");
-        tip_change_signal.notify_one();
+        // A new chain view may make previously unresolved requests serviceable. This
+        // notification can arrive before scanning creates new requests; a second wakeup
+        // is therefore sent after the blocks have been committed below.
+        data_request_signal.notify_one();
```
nullcopy: This should this just be dropped in favor of the notification at the end of this step (the notification after the blocks have actually been processed)

zallet-core/src/components/sync.rs
```diff
@@ -1048,6 +1049,8 @@ async fn steady_state_iteration<C: Chain>(
                 // decryptor (`Handle::queue_tx`) once a single-tx store path exists.
                 // See zcash/zallet#477.
                 decrypt_and_store_transaction(params, db_data, &tx, None)?;
+                // Storing a transaction can expose additional data the wallet needs.
+                data_request_signal.notify_one();
```
nullcopy, line 1055: This should be moved out of the loop. If any mempool data was added, just send one notification after we've ingested each mempool tx

zallet-core/src/components/sync.rs
```diff
@@ -172,26 +172,27 @@ impl WalletSync {
...
-        // TODO: Zaino should provide us an API that allows us to be notified when the chain tip
-        // changes; here, we produce our own signal via the "mempool stream closing" side effect
-        // that occurs in the light client API when the chain tip changes.
-        let tip_change_signal_source = Arc::new(Notify::new());
-        let req_tip_change_signal_receiver = tip_change_signal_source.clone();
+        // The data-requests task needs waking both when existing requests may have become
+        // serviceable (for example after a tip change) and when scanning may have queued new
+        // requests. Keep that meaning explicit instead of representing either event as the
+        // other.
+        let data_request_signal = Arc::new(Notify::new());
```
nullcopy, line 176: Is there a reason not to clone this here anymore (I see it's now cloned at time of use everywhere below)? If not, let's leave it how it was

zallet-core/src/components/sync.rs
```diff
@@ -1280,28 +1288,40 @@ async fn service_address_request<V: ChainView>(
     Ok(())
 }
 
+/// Waits until the data-requests task should make another pass, except that its first pass
+/// runs immediately to service durable work left by an earlier process.
+async fn wait_for_data_request_pass(signal: &Notify, run_immediately: &mut bool) {
```
nullcopy: I think this is a misstep, from a prior AI debug thread. Please remove this and revert the changes to `data_requests` (aside from the rename)

CHANGELOG.md
```diff
@@ -25,6 +25,19 @@ be considered breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- Wallet balances no longer stall part-way while the chain is idle. The
```
nullcopy: I would revise this, now that we understand the failure mode better.

Essentially, there was a race condition: the tip-change signal was being used as a prompt to go process wallet data requests, but that signal fired as soon as the tip change was observed, while new blocks were still being ingested. Now, the signal has been renamed to reflect it's intent, and the signal doesn't fire until the new data has been downloaded and is available for the requests.

Additionally, added a request signal when mempool data becomes available, instead of just on new blocks.

nullcopy (changes requested):
Thanks for running this down. I understand the issue now and what you're fixing. I still have a few requests, but we're on the right track.

NACK 80d88a2034c3e7982dd96363590f1e63325aaf24

zallet-core/src/components/sync.rs
zcash-harry: Agreed. I removed the early notification. The worker now receives the block notification after scanning commits.

zallet-core/src/components/sync.rs
zcash-harry: I kept this notification inside the loop. The ChainView contract says that the mempool stream ends only when the tip changes. Moving the notification after the loop would delay new requests until the next block. Tokio Notify coalesces pending notifications, so these calls do not create an unbounded queue. I added a short code comment to explain this.

zallet-core/src/components/sync.rs
zcash-harry: Agreed. I restored an explicit data_request_signal_receiver clone next to signal creation and pass that clone to the consumer.

zallet-core/src/components/sync.rs
zcash-harry: Agreed. I removed the immediate-pass helper and restored data_requests aside from the signal rename. The original startup notification remains and now uses the data-request signal.

CHANGELOG.md
zcash-harry: Revised. The changelog now gives the race and solution in three short sentences. I also reduced the PR description and commit message.

zallet-core/src/components/sync.rs
zcash-harry: This is AI's reply ^. I am still trying to think whether it makes sense or not, thoughts? 

1. Transaction A arrives
2. Store A
3. Wait an unknown amount of time
4. Transaction B arrives
5. Store B
6. Wait an unknown amount of time

...
There is no “all available mempool transactions are finished” event exposed to Zallet. After transaction A, the next transaction might arrive:
- One millisecond later.
- Five minutes later.
- Never.
Therefore, the smallest known completed batch is one stored mempool transaction:

zallet-core/src/components/sync.rs
```diff
@@ -172,26 +172,25 @@ impl WalletSync {
...
+        let data_request_signal = Arc::new(Notify::new());
+        let data_request_signal_receiver = data_request_signal.clone();
 
         // Spawn the ongoing sync tasks.
         let steady_state_task = {
             let chain = chain.clone();
             let lower_boundary = current_boundary.clone();
             let decryptor = decryptor.clone();
             let status = status.clone();
+            let data_request_signal = data_request_signal.clone();
```
zcash-harry, line 185: This one is cloned again here for `steady_state_task`  as each async move needs it's own clone.

zallet-core/src/components/sync.rs
```diff
@@ -206,12 +205,14 @@ impl WalletSync {
             let chain = chain.clone();
             let mut db_data = db.handle().await?;
             let upper_boundary = current_boundary.clone();
+            let data_request_signal = data_request_signal.clone();
```
zcash-harry, line 208: Same for recovery like steady state, need a clone for this async move

zallet-core/src/components/sync.rs
```diff
@@ -948,7 +948,6 @@ async fn steady_state_iteration<C: Chain>(
                 )
             })
             .expect("closure always returns Some");
-        tip_change_signal.notify_one();
```
zcash-harry, line 951: This is the revert.

zallet-core/src/components/sync.rs
zcash-harry: Steady-state clone sends the post-block and mempool notifications.

zallet-core/src/components/sync.rs
zcash-harry: History-recovery clone sends post-history-scan notifications.

zallet-core/src/components/sync.rs
```diff
@@ -1011,6 +1010,9 @@ async fn steady_state_iteration<C: Chain>(
             // Now that we're done applying the block, update our chain pointer.
             *prev_tip = current_block;
         }
+
+        // Block scanning can add data requests. Notify after the database writes commit.
```
zcash-harry, line 1014: This is the actual fix, we now position the notification at the end of block scanning loop, so it isn't consumed early

zallet-core/src/components/sync.rs
```diff
@@ -1287,18 +1297,16 @@ async fn data_requests<C: Chain>(
     chain: C,
     params: &Network,
     db_data: &mut DbConnection,
-    tip_change_signal: Arc<Notify>,
+    data_request_signal: Arc<Notify>,
 ) -> Result<(), SyncError> {
     loop {
-        // Wait for the chain tip to advance
-        tip_change_signal.notified().await;
+        data_request_signal.notified().await;
```
zcash-harry, line 1303: This is what is waiting for the notifications from either 1015 or line 1180

zallet-core/src/components/sync.rs
```diff
@@ -1169,6 +1175,10 @@ async fn recover_history<C: Chain>(
                     Err(error) => return Err(error),
                 }
             };
+            // `scan_blocks` may commit the prefix below a consensus-divergence boundary
+            // before returning `Break`, so signal before inspecting the outcome.
+            data_request_signal.notify_one();
+
```
zcash-harry, line 1181: This is the other instance of the solution which sends notification during `recover_history`

zallet-core/src/components/sync.rs
nullcopy: Ah, ok, that makes sense

zallet-core/src/components/sync.rs
nullcopy: This is not the revert. The revert commit is before this commit.

nullcopy (commented):
Looks good, except this mysterious deletion I commented on

1d5a012931675caeed01c29aef30dcea829788ae

zallet-core/src/components/sync.rs
zcash-harry: Ah my bad, this isn't the revert. This is the old code before the PR which is not in the correct order of operation 

Its position is immediately after Zallet noticed a new chain tip

zcash-harry: Now it is at line 1015 and 1180, giving two paths to have the notifications

nullcopy (approved):
utACK 1d5a012931675caeed01c29aef30dcea829788ae

PR #823, opened 2026-09-03
migrate-zcashd-wallet: count un-mined transactions toward the birthday

### LINKED ISSUE #573: Include un-mined transactions in zcashd migration in birthday estimation
See: https://github.com/zcash/zallet/pull/571/changes#r3553438989

nullcopy:
Closes #573.
Closes [COR-1533](https://linear.app/zodl/issue/COR-1533).

The birthday estimate previously ignored un-mined transactions: in chain mode
only block hashes that resolved to main-chain heights counted, and an expired
or conflicted transaction contributed nothing. Such a transaction is still
evidence that the wallet's addresses were exposed around a given height, so
history near it could be skipped by the post-import scan.

Both modes now derive the birthday from a shared `earliest_activity_estimate`:
the minimum over recorded mined heights and, for transactions without one,
expiry height less a 1000-block margin (expiry is typically
`creation_height + 40`, zcashd's default `TX_EXPIRY_DELTA` — the same heuristic
no-scan mode already used). In chain mode the estimate is additionally capped
at the chain tip, which continues to stand in for it when the wallet records no
transactions.

Unit tests cover the estimate over mined, un-mined, non-expiring (expiry 0),
and near-genesis (saturating) transactions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ be considered breaking changes.
 
 ### Fixed
 
+- `migrate-zcashd-wallet` now includes un-mined transactions when estimating
+  the wallet birthday. An expired or conflicted transaction still shows that
+  the wallet's addresses were in use around its expiry height, so history near
+  such a transaction is no longer at risk of being skipped by the post-import
+  scan.
 - Fixed a race that could leave wallet data requests unprocessed. The sync task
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK; seems reasonable

PR #824, opened 2026-09-03
 migrate-zcashd-wallet: stop defaulting the birthday to Sapling activation

### LINKED ISSUE #574: Wallet creation height should not default to sapling_activation height for migrated zcashd wallets
See https://github.com/zcash/zallet/pull/571#discussion_r3553403197

nullcopy:
Stacked on https://github.com/zcash/zallet/pull/823 

Closes #574.
Closes [COR-1534](https://linear.app/zodl/issue/COR-1534).

Built on the `nc/birthday-unmined-txs` PR (#573's fix), which
reworks the same expressions; only the last commit is new here.

The migration clamped every birthday up to Sapling activation and used it as
the fallback wherever no evidence existed (birthday, export height, exposure
height). A zcashd wallet may predate Sapling, so the clamp could raise the
birthday above genuine pre-Sapling activity and cut that history out of the
post-import scan.

Evidence of activity now sets the birthday directly, and with no evidence at
all the fallback is the genesis height (`consensus::H0`). Nothing downstream
requires a post-Sapling birthday: `ChainView::tree_state_as_of` legitimately
returns empty frontiers below a pool's activation (both backends test this),
and the ZeWIF importer seeds its chain view from the export height precisely
for birthdays at or below Sapling activation.

Trade-off: a wallet whose evidence (or lack of it) puts the birthday below
Sapling activation will scan the pre-Sapling range too. That range holds no
shielded notes but can hold the wallet's transparent history, which is the
point of the issue. Wallets with post-Sapling evidence are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@ be considered breaking changes.
   the wallet's addresses were in use around its expiry height, so history near
   such a transaction is no longer at risk of being skipped by the post-import
   scan.
+- `migrate-zcashd-wallet` no longer raises the wallet birthday to the Sapling
+  activation height. A zcashd wallet may predate Sapling, so evidence of
+  earlier activity now sets the birthday directly, and a wallet with no usable
+  evidence at all falls back to the genesis height instead of Sapling
+  activation.
 - Fixed a race that could leave wallet data requests unprocessed. The sync task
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK. There are a number of places in librustzcash where we use the same bound; we should check all those.

